PR #150, opened 2014-12-15
add support for purge and refresh with symfony HttpCache

### LINKED ISSUE #146: Provide subscriber for PURGE
Provide a subscriber for cache purging.

The code is probably along the lines of the symfony doc:
http://symfony.com/doc/current/book/http_cache.html#cache-invalidation

But we should have a base class to configure access control. I suggest we use a request matcher for this, that defines whether the request is allowed to purge.

dbu:
fix #146

provide a purge and a refresh handler

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
     "require": {
         "php": ">=5.3.3",
         "guzzle/http": "3.*",
-        "symfony/event-dispatcher": "~2.3"
+        "symfony/event-dispatcher": "~2.3",
+        "symfony/options-resolver": "~2.3"
     },
     "require-dev": {
         "guzzle/plugin-mock": "*",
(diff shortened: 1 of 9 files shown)


borNfreee:
> should i add the invalidation client in this PR too?

Seems like this is what I was looking for. Do you have WIP PR with invalidation client for SymfonyCache?
@dbu 
It's clear that I need to use PurgeSubscriber but I can't understand how to manually *invalidate* by URL using SymfonyCache.

dbu:
for symfony full stack, we are hoping for somebody to pick up https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/183

the basic support would be there in the library: http://foshttpcache.readthedocs.org/en/latest/symfony-cache-configuration.html . its just a question of configuration and updating the documentation.

dbu:
started in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/209

Issue #144, opened 2014-12-03
More general documentation


hpatoio:
The documentation at the moment often refers directly and only to `Varnish` implementation. 

 - [x] http://foshttpcache.readthedocs.org/en/latest/testing-your-application.html
 - [x] http://foshttpcache.readthedocs.org/en/latest/contributing.html
 - etc.

Would be nice to have it more general with specific sections for each reverse proxy .

dbu:
testing doc is now updated. anything else missing or shall we close this issue? we can always further improve when we stumble over other things.

hpatoio:
It's ok to close this. As you said we can always open new issue.

PR #156, opened 2015-01-11
[WIP] Second block of changes | See #144

hpatoio:
Still working on this. Just wanted to share to get feedbacks.

### DIFF
--- a/doc/testing-your-application.rst
+++ b/doc/testing-your-application.rst
@@ -59,17 +59,16 @@ Configuration
 By default, the ``VarnishTestCase`` starts and stops a Varnish server for you.
 All you have to do is to set your Varnish configuration (VCL) file. 
 
-======================= ========================= ============================== ==========================================
-Constant                Getter                    Default                        Description
-======================= ========================= ============================== ==========================================
-``VARNISH_FILE``        ``getConfigFile()``                                      your Varnish configuration (VCL) file
-``VARNISH_BINARY``      ``getBinary()``           ``varnishd``                   your Varnish binary
-``VARNISH_PORT``        ``getCachingProxyPort()`` ``6181``                       port Varnish listens on
-``VARNISH_MGMT_PORT``   ``getVarnishMgmtPort()``  ``6182``                       Varnish management port
-``VARNISH_CACHE_DIR``   ``getCacheDir()``         ``/tmp/foshttpcache-varnish``  Varnish cache dir
-``WEB_SERVER_HOSTNAME`` ``getHostName()``                                        hostname your application can be reached at
-======================= ========================= ============================== ==========================================
-
+======================= ========================= ================================================= ============================================
+Constant                Getter                    Default                                           Description
+======================= ========================= ================================================= ============================================
+``VARNISH_FILE``        ``getConfigFile()``                                                         your Varnish configuration (VCL) file
+``VARNISH_BINARY``      ``getBinary()``           ``varnishd``                                      your Varnish binary
+``VARNISH_PORT``        ``getCachingProxyPort()`` ``6181``                                          port Varnish listens on
+``VARNISH_MGMT_PORT``   ``getVarnishMgmtPort()``  ``6182``                                          Varnish management port
+``VARNISH_CACHE_DIR``   ``getCacheDir()``         ``sys_get_temp_dir()`` + ``foshttpcache-varnish`` directory to use for cache
+``WEB_SERVER_HOSTNAME`` ``getHostName()``                                                           hostname your application can be reached at
+======================= ========================= ================================================= ============================================
 Enable Assertions
 '''''''''''''''''
 
@@ -88,19 +87,17 @@ all cached contents. You have to set your NGINX configuration file.
 
 These are all the configurations you can change
 
-======================= ========================= ============================== ==========================================
+======================= ========================= =============================================== ==========================================
 Constant                Getter                    Default                        Description
-======================= ========================= ============================== ==========================================
-``NGINX_FILE``          ``getConfigFile()``                                      your NGINX configuration file
-``NGINX_BINARY``        ``getBinary()``           ``nginx``                      your NGINX binary
-``NGINX_PORT``          ``getCachingProxyPort()`` ``8088``                       port NGINX listens to
-``NGINX_CACHE_PATH``    ``getCacheDir()``         ``/tmp/foshttpcache-nginx``    NGINX hostname your application can be reached at
-                                                                                 in PHP temporary directory file (sys_get_temp_dir())
-                                                                                 Must match `proxy_cache_path` directive in 
-                                                                                 your configuration file.
-``WEB_SERVER_HOSTNAME`` ``getHostName()``                                        hostname your application can be reached at
-======================= ========================= ============================== ==========================================
-
+======================= ========================= =============================================== ==========================================
+``NGINX_FILE``          ``getConfigFile()``                                                       your NGINX configuration file
+``NGINX_BINARY``        ``getBinary()``           ``nginx``                                       your NGINX binary
+``NGINX_PORT``          ``getCachingProxyPort()`` ``8088``                                        port NGINX listens on
+``NGINX_CACHE_PATH``    ``getCacheDir()``         ``sys_get_temp_dir()`` + ``foshttpcache-nginx`` directory to use for cache
+                                                                                                  Must match `proxy_cache_path` directive in 
+                                                                                                  your configuration file.
+``WEB_SERVER_HOSTNAME`` ``getHostName()``                                                         hostname your application can be reached at
+======================= ========================= =============================================== ==========================================
 Enable Assertions
 '''''''''''''''''
 
--- a/doc/testing-your-application.rst
+++ b/doc/testing-your-application.rst
@@ -51,10 +51,10 @@ You can override getters in your test class in the following way::
     }
 
 VarnishTestCase
--------------
+------------------
 
 Configuration
-'''''''''''''
+''''''''''''''''''
 
 By default, the ``VarnishTestCase`` starts and stops a Varnish server for you.
 All you have to do is to set your Varnish configuration (VCL) file. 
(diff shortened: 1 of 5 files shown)


doc/testing-your-application.rst
```diff
@@ -3,43 +3,26 @@
...
 
-By having your test classes extend ``VarnishTestCase``, you get:
+By having your test classes extend one of the testCase classes, you get:
 
-* independent tests: all previously cached content is removed by restarting
-  Varnish in the test setUp method
-* ``$this->varnish`` referring to an instance of this library’s Varnish client
-  that is configured to talk to your Varnish server
+* independent tests: all previously cached content is removed in the tests 
+  ``setUp`` method. The way this is done depend on which reverse proxy you use.
```
dbu: s/depend/depends/

doc/testing-your-application.rst
```diff
@@ -3,43 +3,26 @@
...
+* an instance of this library’s client that is configured to talk to your 
+  reverse proxy server. See reverse proxy specific sections for details.
 * convenience methods for executing HTTP requests to your application:
-  ``$this->getHttpClient()`` and ``$this->getResponse()``
+  ``$this->getHttpClient()`` and ``$this->getResponse()``.
```
dbu: symfony core doc uses `;` for items in bullet points, except the last item which has `.`. i suggest we do the same.

doc/testing-your-application.rst
```diff
@@ -3,43 +3,26 @@
...
 * custom assertions ``assertHit`` and ``assertMiss`` for validating a cache hit/miss.
 
-Configuration
--------------
-
-By default, the ``VarnishTestCase`` starts and stops a Varnish server for you.
-All you have to do, is to refer the class to your Varnish configuration (VCL)
-file. The recommended way to configure the test case is by setting constants
+The recommended way to configure the test case is by setting constants
 in your `phpunit.xml`. Alternatively, you can override the getter methods:
```
dbu: you need to replace `:` by `.`

doc/testing-your-application.rst
```diff
@@ -67,15 +50,61 @@ You can override getters in your test class in the following way::
...
     }
 
+VarnishTestCase
+------------------
+
+Configuration
+''''''''''''''''''
+
+By default, the ``VarnishTestCase`` starts and stops a Varnish server for you.
+All you have to do is to set your Varnish configuration (VCL) file.
```
dbu: you also need to set the hostname, no? or is the table incomplete and there is a default for the hostname?
/cc @ddeboer

doc/testing-your-application.rst
```diff
@@ -67,15 +50,61 @@ You can override getters in your test class in the following way::
         }
     }
 
+VarnishTestCase
+------------------
```
dbu: please align title underline with title length (also below)

doc/testing-your-application.rst
```diff
@@ -67,15 +50,61 @@ You can override getters in your test class in the following way::
...
 :ref:`custom X-Debug header <varnish_debugging>` to responses served
 by your Varnish.
 
+NginxTestCase
+------------------
+
+Configuration
+''''''''''''''''''
+
+By default, the ``NginxTestCase`` starts and stops NGINX server for you and delete
```
dbu: ... starts and stops the NGINX server for you and deletes ...

doc/testing-your-application.rst
```diff
@@ -67,15 +50,61 @@ You can override getters in your test class in the following way::
...
+
+These are all the configurations you can change
+
+======================= ========================= ============ ==========================================
+Constant                Getter                    Default      Description
+======================= ========================= ============ ==========================================
+``NGINX_FILE``          ``getConfigFile()``                    your NGINX configuration file
+``NGINX_BINARY``        ``getBinary()``           ``nginx``    your NGINX binary
+``NGINX_PORT``          ``getCachingProxyPort()`` ``8088``     port NGINX listens to
+``NGINX_CACHE_PATH``    ``getCacheDir()``                      NGINX cache dir path
```
dbu: is there a default for this? is this specified in the configuration file? (we should make it clear why you have to specify this, and if this must match what is configured in the config file)

phpunit.xml.dist
```diff
@@ -22,6 +22,8 @@
         <listener class="\Mockery\Adapter\Phpunit\TestListener" />
         <listener class="\FOS\HttpCache\Tests\Functional\WebServerListener" />
     </listeners>
+    
+    <!-- You probably need to change this settings when you write your own functional test -->
```
dbu: this is the phpunit file for the library. do we need this comment here?

src/Test/NginxTestCase.php
```diff
@@ -17,32 +17,37 @@
...
+ * WEB_SERVER_HOSTNAME  hostname where your application can be reached (required)
+ * NGINX_BINARY         Executable for NGINX. This can also be the full path
  *                      to the file if the binary is not automatically found
  *                      (default nginx)
- * NGINX_PORT         Test NGINX port to use (default 8088)
- * NGINX_FILE         NGINX configuration file (required if not passed to setUp)
- * NGINX_CACHE_PATH   NGINX configuration file (required if not passed to setUp)
+ * NGINX_PORT           Port NGINX is listening to (default 8088)
+ * NGINX_CACHE_PATH     directory to use for cache
+ *                      (default /tmp/foshttpcache-nginx)
```
dbu: looks like the table in the doc is out of sync. there is a default.

src/Test/Proxy/NginxProxy.php
```diff
@@ -15,7 +15,7 @@ class NginxProxy extends Abstractproxy
 {
     protected $binary = 'nginx';
     protected $configFile;
-    protected $port = 8080;
+    protected $port = 8088;
```
dbu: hm, changing this default is a BC break. we should rather change the doc if it is incorrect.

src/Test/VarnishTestCase.php
```diff
@@ -28,15 +27,15 @@
...
  * VARNISH_BINARY       executable for varnish. this can also be the full path
  *                      to the file if the binary is not automatically found
  *                      (default varnishd)
  * VARNISH_PORT         test varnish port to use (default 6181)
  * VARNISH_MGMT_PORT    test varnish mgmt port (default 6182)
- * VARNISH_FILE         varnish configuration file (required if not passed to setUp)
  * VARNISH_CACHE_DIR    directory to use for cache
- *                      (default /tmp/foshttpcache-test)
- * WEB_SERVER_HOSTNAME  name of the webserver varnish has to talk to (required)
+ *                      (default /tmp/foshttpcache-varnish)
```
dbu: could you please also add the VARNISH_CACHE_DIR to the table in the doc?

dbu:
coming good, thanks!

src/Test/NginxTestCase.php
hpatoio: I've updated the doc. BTW: The default is not `/tmp/foshttpcache-nginx` but `sys_get_temp_dir()` + `foshttpcache-nginx` (this is valid also for Vanish)

I've added the explanation in the "Description" column but dunno if change also del "Default" Column.

src/Test/Proxy/NginxProxy.php
hpatoio: Hmm. Actually the correct value is `8088`. My tests are not green with `8080` I'm digging on this.

src/Test/VarnishTestCase.php
hpatoio: Done

src/Test/NginxTestCase.php
dbu: lets fix it here too, better be correct in the doc ;-)

src/Test/Proxy/NginxProxy.php
dbu: maybe our phpunit.xml changes the value through a constant? when running nginx and varnish on the same system, they will need to run on separate ports maybe.

src/Test/VarnishTestCase.php
dbu: is this the same as with nginx, sys_get_temp_dir()?

doc/testing-your-application.rst
```diff
@@ -3,45 +3,28 @@
...
-``VARNISH_PORT``        ``getCachingProxyPort()`` ``6181``     port Varnish listens on
-``VARNISH_MGMT_PORT``   ``getVarnishMgmtPort()``  ``6182``     Varnish management port
-``WEB_SERVER_HOSTNAME`` ``getHostName()``                      hostname your application can be reached at
-======================= ========================= ============ ==========================================
+The recommended way to configure the test case is by setting constants
+in your `phpunit.xml`. Alternatively, you can override the getter methods.
 
 Setting Constants
-~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~
```
stof: this is wrong. It should have the same length than the title (Sphinx requires at least the length, but the Symfony standards are to use the same length)

src/Test/Proxy/NginxProxy.php
hpatoio: Well. Actually that value is not used. The port is set here https://github.com/hpatoio/FOSHttpCache/blob/master/src/Test/NginxTestCase.php#L81 and as you can see is always overwritten.

I'll open another PR with some ideas about this.

dbu: yeah, lets solve that separately. but lets not change this value for now then. maybe somebody is using this class in a different context, and then things would break for him.

dbu:
so i'd say: lets wrap up the open styling issues and undo the change about the default port, then squash and merge things.

src/Test/VarnishTestCase.php
hpatoio: Yes, is the same. I've fixed it here too.

src/Test/Proxy/NginxProxy.php
hpatoio: "reverted" wait for another PR.

hpatoio:
I think we have arrived to an acceptable point. It can be improved of course. But I think we can merge it.

doc/testing-your-application.rst
```diff
@@ -67,13 +50,60 @@ You can override getters in your test class in the following way::
...
+``NGINX_CACHE_PATH``    ``getCacheDir()``         ``sys_get_temp_dir()`` + ``foshttpcache-nginx`` directory to use for cache
+                                                                                                  Must match `proxy_cache_path` directive in 
+                                                                                                  your configuration file.
+``WEB_SERVER_HOSTNAME`` ``getHostName()``                                                         hostname your application can be reached at
+======================= ========================= =============================================== ==========================================
+Enable Assertions
+'''''''''''''''''
+
+For the `assertHit` and `assertMiss` assertions to work, you should add the HTTP 
+header ``X-Cache`` to your responses
```
dbu: @ddeboer just cleaned this up. its X-Cache-Debug if i am not really mistaken. i will squash your commits and adjust this and merge.

dbu:
merged in #160 - thanks a lot!

PR #158, opened 2015-01-17
Clean up debug docs

ddeboer:
Our Varnish 3 and 4 VCLs differed in that the first set the `X-Cache` header conditionally and the latter did so always.


dbu:
indeed. thanks!

PR #161, opened 2015-01-17
document x-cache header for nginx

dbu:
adding doc how to activate the X-Cache header with nginx.

@hpatoio if you happen to know how we could set the X-Cache header only if the backend sends an X-Cache-Debug header, like we do for varnish, then please tell and we can add that.

### DIFF
--- a/doc/nginx-configuration.rst
+++ b/doc/nginx-configuration.rst
@@ -7,10 +7,10 @@ Below you will find detailed NGINX configuration recommendations for the
 features provided by this library. The examples are tested with NGINX version
 1.4.6.
 
-NGINX cache is a set of key/value pairs. The key is built with elements taken from the requests 
+NGINX cache is a set of key/value pairs. The key is built with elements taken from the requests
 (URI, cookies, http headers etc) as specified by `proxy_cache_key` directive.
 
-When we interact with the cache to purge/refresh entries we must send to NGINX a request which has 
+When we interact with the cache to purge/refresh entries we must send to NGINX a request which has
 the very same values, for the elements used for building the key, as the request that create the entry.
 In this way NGINX can build the correct key and apply the required operation to the entry.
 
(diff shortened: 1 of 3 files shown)


ddeboer:
I’m guessing:

```
if ($sent_http_x_cache_debug) {
}
```

Then again, at NGINX they seem to [dislike ifs](http://wiki.nginx.org/IfIsEvil), so perhaps we shouldn't recommend adding one. @hpatoio Wdyt?

dbu:
@hpatoio can you help us here? can i conditionally set the X-Cache header?

hpatoio:
Ciao, just see this. Checking ASAP.

hpatoio:
Is possible using `map`. `X-Cache-Debug` get `HIT` or `MISS` as value. Nothing else right ?  
Going to write the conf.

References:
 - http://nginx.org/en/docs/http/ngx_http_map_module.html 
 - http://serverfault.com/questions/515501/add-expires-header-conditionally-based-on-mime-type-in-nginx

dbu:
it seems that we can know whether we have hit or miss from `$upstream_cache_status`, see https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/tests/Functional/Fixtures/nginx/fos.conf#L20

i just googled and it seems to me that the header is available as `$sent_http_x_cache_debug`, see  http://wiki.nginx.org/HttpCoreModule#.24sent_http_HEADER

would it not be enough to do 

    if ($sent_http_x_cache_debug) {
        add_header X-Cache $upstream_cache_status;
    }

dbu:
ah, except that nginx can not do an if statement on the header. i suggest we just drop that then, not worth the complication imho.

i have to say i am kind of surprised how awkward nginx is as a proxy. varnish seems a lot more straightforward for this (ok, but then varnish also does nothing else than be a proxy).

PR #162, opened 2015-01-18
Separating the tag handler into its own class

dbu:
Following the discussions, i want to build a separate TagHandler. This will be even more useful for the bundle, when we add further tag logic: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/182

/cc @dantleech

* [x] Add tests for the TagHandler
* [x] Adjust documentation
* [x] Create issue for v2 to drop all deprecated stuff #163

### DIFF
--- a/doc/invalidation-handlers.rst
+++ b/doc/invalidation-handlers.rst
@@ -1,0 +1,76 @@
+Extra Invalidation Handlers
+===========================
+
+This library provides decorators that build on top of the ``CacheInvalidator``
+to simplify common operations.
+
+.. _tags:
+
+Tag Handler
+-----------
+
+The tag handler helps you to invalidate all cache entries that where marked
+with a specified tag. It works only with a ``CacheInvalidator`` that supports
+``CacheInvalidator::INVALIDATE``.
+
+Setup
+~~~~~
+
+.. note::
+
+    Make sure to :doc:`configure your proxy <proxy-configuration>` for tagging first.
+
+The tag handler is a decorator around the ``CacheInvalidator``. After
+:doc:`creating the invalidator <cache-invalidator>` with a proxy client
+that implements the ``BanInterface``, instantiate the ``TagHandler``::
+
+    use FOS\HttpCache\Handler\TagHandler;
+
+    // $cacheInvalidator already created as instance of FOS\HttpCache\CacheInvalidator
+    $tagHandler = new TagHandler($cacheInvalidator);
+
+Usage
+~~~~~
+
+With tags you can group related representations so it becomes easier to
+invalidate them. You will have to make sure your web application adds the
+correct tags on all responses by setting the ``X-Cache-Tags`` header. The
+FOSHttpCacheBundle_ does this for you when you’re using Symfony.
+
+Assume you sent four responses:
+
++------------+-------------------------+
+| Response:  | ``X-Cache-Tags`` header:|
++============+=========================+
+| ``/one``   | ``tag-one``             |
++------------+-------------------------+
+| ``/two``   | ``tag-two, group-a``    |
++------------+-------------------------+
+| ``/three`` | ``tag-three, group-a``  |
++------------+-------------------------+
+| ``/four``  | ``tag-four, group-b``   |
++------------+-------------------------+
+
+You can now invalidate some URLs using tags::
+
+    $tagHandler->invalidateTags(array('group-a', 'tag-four'))->flush();
+
+This will ban all requests having either the tag ``group-a`` /or/ ``tag-four``.
+In the above example, this will invalidate ``/two``, ``/three`` and ``/four``.
+Only ``/one`` will stay in the cache.
+
+.. _custom_tags_header:
+
+Custom Tags Header
+~~~~~~~~~~~~~~~~~~
+
+Tagging uses a custom HTTP header to identify tags. You can change the default
+header ``X-Cache-Tags`` in the constructor::
+
+    use FOS\HttpCache\Handler\TagHandler;
+
+    // $cacheInvalidator already created as instance of FOS\HttpCache\CacheInvalidator
+    $tagHandler = new TagHandler($cacheInvalidator, 'My-Cache-Header');
+
+Make sure to reflect this change in your
+:doc:`caching proxy configuration <proxy-configuration>`.
--- a/src/CacheInvalidator.php
+++ b/src/CacheInvalidator.php
@@ -20,6 +20,7 @@ use FOS\HttpCache\ProxyClient\ProxyClientInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\PurgeInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
+use FOS\HttpCache\Handler\TagHandler;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
@@ -28,6 +29,7 @@ use Symfony\Component\EventDispatcher\EventSubscriberInterface;
  * Manages HTTP cache invalidation.
  *
  * @author David de Boer <david@driebit.nl>
+ * @author David Buchmann <mail@davidbu.ch>
  */
 class CacheInvalidator
 {
@@ -57,6 +59,11 @@ class CacheInvalidator
     private $eventDispatcher;
 
     /**
+     * @var TagHandler
+     */
+    private $tagHandler;
+
+    /**
      * @var string
      */
     private $tagsHeader = 'X-Cache-Tags';
@@ -69,6 +76,9 @@ class CacheInvalidator
     public function __construct(ProxyClientInterface $cache)
     {
         $this->cache = $cache;
+        if ($cache instanceof BanInterface) {
+            $this->tagHandler = new TagHandler($this, $this->tagsHeader);
+        }
     }
 
     /**
@@ -145,10 +155,14 @@ class CacheInvalidator
      * @param string $tagsHeader
      *
      * @return $this
+     *
+     * @deprecated Use constructor argument to TagHandler instead.
      */
     public function setTagsHeader($tagsHeader)
     {
-        $this->tagsHeader = $tagsHeader;
+        if ($this->tagHandler && $this->tagsHeader !== $tagsHeader) {
+            $this->tagHandler = new TagHandler($this, $tagsHeader);
+        }
 
         return $this;
     }
@@ -157,6 +171,8 @@ class CacheInvalidator
      * Get the HTTP header name that will hold cache tags
      *
      * @return string
+     *
+     * @deprecated Use TagHandler::getTagsHeader instead.
      */
     public function getTagsHeader()
     {
@@ -269,22 +285,21 @@ class CacheInvalidator
      * Invalidate cache entries that contain any of the specified tags in their
      * tag header.
      *
-     * @see BanInterface::ban()
-     *
      * @param array $tags Cache tags
      *
      * @throws UnsupportedProxyOperationException If HTTP cache does not support BAN requests
      *
      * @return $this
+     *
+     * @deprecated Use TagHandler::invalidateTags instead.
      */
     public function invalidateTags(array $tags)
     {
-        if (!$this->cache instanceof BanInterface) {
+        if (!$this->tagHandler) {
             throw UnsupportedProxyOperationException::cacheDoesNotImplement('BAN');
         }
 
-        $headers = array($this->getTagsHeader() => '('.implode('|', array_map('preg_quote', $tags)).')(,.+)?$');
-        $this->cache->ban($headers);
+        $this->tagHandler->invalidateTags($tags);
 
         return $this;
     }
--- a/src/Handler/TagHandler.php
+++ b/src/Handler/TagHandler.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCache package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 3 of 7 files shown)


src/CacheInvalidator.php
```diff
@@ -47,28 +48,38 @@ class CacheInvalidator
...
-     * @param ProxyClientInterface $cache HTTP cache
+     * @param ProxyClientInterface $cache      HTTP cache
+     * @param TagHandler           $tagHandler A tag handler instance (optional)
      */
-    public function __construct(ProxyClientInterface $cache)
+    public function __construct(ProxyClientInterface $cache, TagHandler $tagHandler = null)
     {
         $this->cache = $cache;
+        if (null === $tagHandler && $cache instanceof BanInterface) {
+            $tagHandler = new TagHandler($cache);
```
dbu: this is needed for bc. not sure if its a good idea to keep it.

src/Tag/TagHandler.php
```diff
@@ -0,0 +1,93 @@
...
+use FOS\HttpCache\ProxyClient\ProxyClientInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
+
+/**
+ * Handler for cache tagging.
+ *
+ * @author David de Boer <david@driebit.nl>
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+class TagHandler
```
dbu: i added no flush method to this class. this is fine when using the same cache proxy client as for the CacheInvalidator, but would not work with separate client instances. but it should not be the job of a TagHandler to flush the cache client imho.

src/CacheInvalidator.php
```diff
@@ -47,28 +48,38 @@ class CacheInvalidator
...
 
     /**
      * Constructor
      *
-     * @param ProxyClientInterface $cache HTTP cache
+     * @param ProxyClientInterface $cache      HTTP cache
+     * @param TagHandler           $tagHandler A tag handler instance (optional)
      */
-    public function __construct(ProxyClientInterface $cache)
+    public function __construct(ProxyClientInterface $cache, TagHandler $tagHandler = null)
```
dantleech: So does this mean we would add the ObjectHandler in this way?

````php
public function __construct(ProxyClientInterface $cache, TagHandler $tagHandler = null., ObjectHandler $objectHandler = null)
````

File: src/CacheInvalidator.php [hunk 4]
```diff
@@ -20,6 +20,7 @@
 use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\PurgeInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
+use FOS\HttpCache\Tag\TagHandler;
```
ddeboer: Wouldn’t a `Handler` namespace be more generic?

src/Tag/TagHandler.php
ddeboer: Agreed.

src/CacheInvalidator.php
```diff
@@ -279,12 +302,11 @@ public function invalidateRegex($path, $contentType = null, $hosts = null)
...
     public function invalidateTags(array $tags)
     {
-        if (!$this->cache instanceof BanInterface) {
+        if (!$this->tagHandler) {
             throw UnsupportedProxyOperationException::cacheDoesNotImplement('BAN');
         }
 
-        $headers = array($this->getTagsHeader() => '('.implode('|', array_map('preg_quote', $tags)).')(,.+)?$');
-        $this->cache->ban($headers);
+        $this->tagHandler->invalidateTags($tags);
```
ddeboer: With only such a simple proxy the value of the `CacheInvalidator` facade becomes less clear, so we may want to remove `CacheInvalidator::invalidateTags()` later. Still, let’s keep the `invalidateTags()` method for BC.

src/CacheInvalidator.php
stof: @dantleech I would say no. I don't think the ObjectHandler should come in the CacheInvalidator (the TagHandler is injected here for BC only

src/Tag/TagHandler.php
```diff
@@ -0,0 +1,93 @@
...
+     * @param array $tags Cache tags
+     *
+     * @throws UnsupportedProxyOperationException If HTTP cache does not support BAN requests
+     *
+     * @return $this
+     */
+    public function invalidateTags(array $tags)
+    {
+        if (!$this->cache instanceof BanInterface) {
+            throw UnsupportedProxyOperationException::cacheDoesNotImplement('BAN');
```
stof: this is dead code because of the constructor typehint

src/CacheInvalidator.php
ddeboer: @stof There’s still some discussion around whether the CacheInvalidator should function as a complete facade including tagging or not. If so, it obviously requires access to the new TagHandler.

dantleech: But then by what criteria do we say that the ObjectHandler shouldn't be in this facade? or should it? (I would lean towards not)

src/CacheInvalidator.php
dantleech: +1

src/CacheInvalidator.php
dbu: sure, until we bump to 2.0 we have to keep this either way.

src/CacheInvalidator.php
```diff
@@ -94,6 +105,8 @@ public function supports($operation)
                 return $this->cache instanceof RefreshInterface;
             case self::INVALIDATE:
                 return $this->cache instanceof BanInterface;
+            case self::TAG:
+                return null !== $this->tagHandler;
```
dbu: if we deprecate invalidating tags through the CacheInvalidator we should not add this

src/CacheInvalidator.php
dbu: the more we discuss this, i think it makes more sense to not inject this here. we need to create the cache invalidator in here for BC but injecting it is not a feature.

dbu:
ok, but if we separate the TagHandler completely, how exactly will we manage BC? if we create the TagHandler in the constructor, but the bundle starts providing a separate service for it, there exist two instances of the TagHandler and setTagHeader on one but then using the other to tag will have unexpected results.

the other thing is how to handle flush when we have several parts that theoretically could have different cache clients injected. i guess we just make the bundle use the same for all and try to explain it in each handler phpdoc that you need to flush the proxy client at the end.

src/CacheInvalidator.php
dbu: i was pondering this. the reason to *not* use Handler was that i thought about the ObjectHandler we plan. that will likely use more than one class and even different strategies (tags vs banning subpaths). stuffing all of that into Handler feels weird. UserContext is also in its own namespace (though its not a "handler" so we can see that differently).

though i have to say, putting the ObjectHandler into Object sounds bizarre too. any good ideas for naming the two?

dantleech:
> ok, but if we separate the TagHandler completely, how exactly will we manage BC?

I am not against injecting it for BC. 

> how to handle flush when we have several parts that theoretically could have different cache clients injected

A handler registry/manager? All tagged handlers could be pulled into the registry. They could then be individually injected where needed and a single call to flush on the registry would flush them all (the ProxyClients would need to be responsible for keeping track of if they need to be flushed or not). But maybe thats overkill.

src/CacheInvalidator.php
dantleech:
````php
Handler/ObjectHandler.php
Handler/TagHandler.php
Handler/Object/ObjectManager.php (or whatever)
Handler/Object/SomeOtherObjectRelatedClass.php
````

?

**Comment by dbu** [comment]
did the requested cleanup. and just realized that the TagHandler actually is a wrapper around the CacheInvalidator - it does not need the ProxyClient directly. with this in mind, the whole flush topic becomes moot - i just inject the whole CacheInvalidator to the TagHandler. much simpler and more logic.

the tag oriented ObjectHandler in turn is a wrapper around the TagHandler. if we do path based logic, we can additionally inject the CacheInvalidator to the ObjectHandler for direct access to that.

File: src/CacheInvalidator.php (hunk 4 above)
Comment (dbu) [inline src/CacheInvalidator.php]: hm, that can work, yes. other ideas @ddeboer or shall i change to that?

**Comment by dbu** [comment]
@dantleech @ddeboer do you agree with this now? if so, i would add tests and adjust the documentation.

File: src/CacheInvalidator.php (hunk 4 above)
Comment (ddeboer) [inline src/CacheInvalidator.php]: `Handler\TagHandler` looks fine.

**Comment by ddeboer** [comment]
:+1:

File: src/Tag/TagHandler.php [hunk 8]
```diff
@@ -0,0 +1,89 @@
...
+     * Invalidate cache entries that contain any of the specified tags in their
+     * tag header.
+     *
+     * @param array $tags Cache tags
+     *
+     * @return $this
+     */
+    public function invalidateTags(array $tags)
+    {
+        $headers = array($this->getTagsHeader() => '('.implode('|', array_map('preg_quote', $tags)).')(,.+)?$');
```
dantleech: This would be less cryptic with `sprintf`

dantleech:
Good for me

dbu:
alright. now with tests and documentation. wdyt?

doc/invalidation-handlers.rst
```diff
@@ -0,0 +1,71 @@
...
+This library provides decorators that build on top of the CacheInvalidator to
+simplify common operations.
+
+.. _tags:
+
+Tag Handler
+-----------
+
+The tag handler helps you to invalidate all cache entries that where marked
+with a specified tag. It works only with a CacheInvalidator that supports
```
dantleech: Double backticks for CacheInvalidator (or does this repo use different standards?)

doc/invalidation-handlers.rst
```diff
@@ -0,0 +1,71 @@
...
+CacheInvalidator::INVALIDATE.
```
dantleech: Double ticks

doc/invalidation-handlers.rst
```diff
@@ -0,0 +1,71 @@
...
+Setup
+~~~~~
+
+.. note::
+
+    Make sure to :doc:`configure your proxy <proxy-configuration>` for tagging first.
+
+The tag handler is a decorator around the ``CacheInvalidator``. After
+:doc:`creating the invalidator <cache-invalidator.rst>` with a proxy client
+that implements the BanInterface, instantiate the ``TagHandler``::
```
dantleech: Dbl ticks

doc/invalidation-handlers.rst
```diff
@@ -0,0 +1,71 @@
...
+| ``/three`` | ``tag-three, group-a``  |
++------------+-------------------------+
+| ``/four``  | ``tag-four, group-b``   |
++------------+-------------------------+
+
+You can now invalidate some URLs using tags::
+
+    $tagHandler->invalidateTags(array('group-a', 'tag-four'))->flush();
+
+
```
dantleech: Extra blank line

doc/invalidation-handlers.rst
```diff
@@ -0,0 +1,71 @@
...
+In the above example, this will invalidate ``/two``, ``/three`` and ``/four``.
+Only ``/one`` will stay in the cache.
+
+.. _custom_tags_header:
+
+Custom Tags Header
+~~~~~~~~~~~~~~~~~~
+
+Tagging uses a custom HTTP header to identify tags. You can change the default
+header ``X-Cache-Tags`` by calling ``setTagsHeader()``. Make sure to reflect this
```
dantleech: Should `setTagsHeader` be in configuration and passed in the `_construct` ?

dbu: right you are. fixing

ddeboer:
Docs build is failing.

src/CacheInvalidator.php
```diff
@@ -52,23 +53,26 @@ class CacheInvalidator
...
+     * @var EventDispatcherInterface
      */
-    private $tagsHeader = 'X-Cache-Tags';
+    private $eventDispatcher;
 
     /**
      * Constructor
      *
-     * @param ProxyClientInterface $cache HTTP cache
+     * @param ProxyClientInterface $cache      HTTP cache
```
ddeboer: Just a space (instead if tab) is fine here.

src/Handler/TagHandler.php
```diff
@@ -0,0 +1,94 @@
...
+     * @return $this
+     *
+     * @deprecated Pass this setting in the header.
+     */
+    public function setTagsHeader($tagsHeader)
+    {
+        $this->tagsHeader = $tagsHeader;
+
+        return $this;
+    }
```
ddeboer: Now we have a separate tagging class to which we can pass this value in its constructor, I say we drop this setter and getter. After all, only funny things will happen if users change the tag header value at runtime.

dbu: i thought we need it to keep CacheInvalidator BC but realized we can do
without.

dbu:
there we go, updated

src/CacheInvalidator.php
dbu: added deprecation note now. but we need to keep this until 2.0

dbu:
hm, should we update the branch alias to 1.3 for this? its a new feature, not a bugfix. and would allow the bundle PR for Tagging to bump the version requirement of the library to 1.3 to get the right one... but i can also bump the requirement to 1.2.1 if you think that is good enough.

dbu:
hm, and i forgot a changelog entry. 1.3? but lets not tag just yet, so that we still can add stuff if needed.

ddeboer:
If we update the branch alias now, users on 1.2.x-dev will not receive any bugfixes to the 1.2.0 release, right? We decided earlier to keep our release schedule simple, so perhaps we should just tag this 1.3.0 ASAP and update the alias then. Yes, we could wait with a tag, but there’s nothing wrong with releasing early and often.

dantleech:
Should the tag handler not also take care of tagging the response?

dbu:
> Should the tag handler not also take care of tagging the response?

it will in the bundle. the library offers no such thing as there is no
Response class (we do not use HttpFoundation in the library)

we could offer a wrapper for `header()` that properly escapes the tags
though, just thinking now that you asked that question.

Issue #147, opened 2014-12-11
Support basic auth in user context subscriber


dbu:
The user context subscriber added in #145 should support basic authentication header instead or in addition to the session cookie.

The default behaviour should be to look at both authorization header and cookies, just as we propose with varnish. we could add an option to select one of the two exclusively and then ignore the other if the option is set.

hpatoio:
The library call the the `URI` specified in `user_hash_uri` and there you generate the response content. See http://foshttpcache.readthedocs.org/en/latest/user-context.html#returning-the-user-context-hash

Am I right @dbu ?

ddeboer:
That's exactly right, @hpatoio. See also http://foshttpcache.readthedocs.org/en/latest/symfony-cache-configuration.html.

sprain:
Oh, I see. Thanks. I will look into it.

ddeboer:
@sprain Great if you can add this functionality!

dbu:
fixed in #167

PR #167, opened 2015-02-04
Add authentication support to user context subscriber

sprain:
This is a draft for https://github.com/FriendsOfSymfony/FOSHttpCache/issues/147, feedback is welcome! Tests will be added once the approach is accepted (I am new to all this caching stuff).

---
*everything below is obsolete*

I added the functionality to determine whether a request is anonymous based on session, basic authentication or oAuth2.

In my Symfony2 app, to only use oAuth2, I would overwrite `getDefaultSubscribers` in my own cache class like this:

```php
class AppCache extends EventDispatchingHttpCache
{
    protected function getDefaultSubscribers()
    {
        $options = $this->getOptions();
        $subscribers = array();
        $defaultSubscribersOption = isset($options['fos_default_subscribers']) ? $options['fos_default_subscribers'] : self::SUBSCRIBER_ALL;
        if ($defaultSubscribersOption & self::SUBSCRIBER_USER_CONTEXT) {
            $subscribers[] = new UserContextSubscriber(array('authorization' => array('oauth2')));
        }

        return $subscribers;
    }
}
```

### DIFF
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -118,9 +118,10 @@ User Context
 ~~~~~~~~~~~~
 
 To support :doc:`user context hashing <user-context>` you need to register the
-``UserContextSubscriber``. If the default settings are right for you, you don't
-need to do anything more. You can customize a number of options through the
-constructor:
+``UserContextSubscriber``. The user context is then automatically recognized
+based on session cookies or authorization headers. If the default settings are
+right for you, you don't need to do anything more. You can customize a number of
+options through the constructor:
 
 * **anonymous_hash**: Hash used for anonymous user. This is a performance
   optimization to not do a backend request for users that are not logged in.
@@ -159,6 +160,16 @@ constructor:
     Wrong session name will lead to unexpected results such as having the same
     user context hash for every users, or not having it cached at all, which
     hurts performance.
+
+.. note::
+
+    To use authorization headers for user context, you might have to add some server
+    configuration to make these headers available to PHP.
+
+    With Apache, you can do this for example in a ``.htaccess`` file:
+
+        RewriteEngine On
+        RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
 
 Cleaning the Cookie Header
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/SymfonyCache/UserContextSubscriber.php
+++ b/src/SymfonyCache/UserContextSubscriber.php
@@ -130,7 +130,10 @@ class UserContextSubscriber implements EventSubscriberInterface
                 $hashLookupRequest->cookies->set($name, $value);
             }
         }
-        $hashLookupRequest->headers->set('Cookie', http_build_query($sessionIds, '', '; '));
+
+        if (count($sessionIds) > 0) {
+            $hashLookupRequest->headers->set('Cookie', http_build_query($sessionIds, '', '; '));
+        }
     }
 
     /**
@@ -180,6 +183,16 @@ class UserContextSubscriber implements EventSubscriberInterface
      */
     private function isAnonymous(Request $request)
     {
+        // You might have to enable rewriting of the Authorization header in your server config or .htaccess:
+        // RewriteEngine On
+        // RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+        if ($request->server->has('AUTHORIZATION') ||
+            $request->server->has('HTTP_AUTHORIZATION') ||
+            $request->server->has('PHP_AUTH_USER')
+        ) {
+            return false;
+        }
+
         foreach ($request->cookies as $name => $value) {
             if ($this->isSessionName($name)) {
                 return false;
--- a/tests/Unit/SymfonyCache/UserContextSubscriberTest.php
+++ b/tests/Unit/SymfonyCache/UserContextSubscriberTest.php
@@ -132,6 +132,9 @@ class UserContextSubscriberTest extends \PHPUnit_Framework_TestCase
         $cookieString = "PHPSESSID=$sessionId1; foo=bar; PHPSESSIDsdiuhsdf4535d4f=$sessionId2";
         $request = Request::create('/foo', 'GET', array(), $cookies, array(), array('Cookie' => $cookieString));
 
+        // Add a cookie which should not be available in the eventual hash request anymore
+        $request->cookies->set('foo', 'bar');
+
         $hashRequest = Request::create($options['user_hash_uri'], $options['user_hash_method'], array(), array(), array(), $request->server->all());
         $hashRequest->attributes->set('internalRequest', true);
         $hashRequest->headers->set('Accept', $options['user_hash_accept_header']);
@@ -160,6 +163,7 @@ class UserContextSubscriberTest extends \PHPUnit_Framework_TestCase
                     $request->getMethod();
                     $request->getPathInfo();
                     $that->assertEquals($hashRequest, $request);
+                    $that->assertCount(2, $request->cookies->all());
 
                     return true;
                 }),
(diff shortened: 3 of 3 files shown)


ddeboer:
Does it make sense to separate the different authentication mechanisms (OAuth2, HTTP) or should we just check generically on whether an `Authorization` header is present? The problem I see with differentiating between mechanisms is that we can keep adding them: now e.g. OAuth1 (people still use that) is missing.

dbu:
i would vote for a simpler approach in the sense of what varnish is doing: just check if a Authorization header exists and if so consider the request not anonymous. no configuration, no options. this is the expected behaviour to me, and if i need to tweak that i would extend the subscriber class to change the behaviour

dbu:
@sprain and thanks a lot for your contribution! i appreciate that.

sprain:
Simplifying sounds good ;) Will look at it.

sprain:
Ok, I have simplified this. Any feedbacks?

src/SymfonyCache/UserContextSubscriber.php
```diff
@@ -180,6 +180,15 @@ private function getUserHash(HttpKernelInterface $kernel, Request $request)
      */
     private function isAnonymous(Request $request)
     {
+        // You might have to enable rewriting of the Authorization header in your server config or .htaccess:
+        // RewriteEngine On
+        // RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+        foreach ($request->server as $name => $value) {
```
dbu: lets do `if ($request->server->has('HTTP_AUTHORIZATION') ...` rather than loop through all the headers.

looking at the ServerBag::getHeaders method, i have the impression that the safe thing to do would be to check for any of the headers AUTHORIZATION, HTTP_AUTHORIZATION or PHP_AUTH_USER and consider the request not anonymous if any of those headers is present.

dbu:
can you please add some tests like https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/tests/Unit/SymfonyCache/UserContextSubscriberTest.php#L120 but with AUTHORIZATION headers instead of session cookie? the anonymous case is already covered, but we should have a test where we do have one of these headers and end up being not anonymous.

sprain:
Updated, please have a look.

src/SymfonyCache/UserContextSubscriber.php
```diff
@@ -130,7 +130,10 @@ protected function cleanupHashLookupRequest(Request $hashLookupRequest, Request
                 $hashLookupRequest->cookies->set($name, $value);
             }
         }
-        $hashLookupRequest->headers->set('Cookie', http_build_query($sessionIds, '', '; '));
+
+        if (count($sessionIds) > 0) {
+            $hashLookupRequest->headers->set('Cookie', http_build_query($sessionIds, '', '; '));
+        }
```
dbu: i think otherwise we should remove the cookie completely, to prevent it from being used in the hash if we assume its not a session thing.

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
```diff
@@ -177,6 +177,58 @@ public function testUserHashUserWithSession($arg, $options)
...
+            ->with(
+                $this->callback(function (Request $request) use ($that, $hashRequest) {
+                    // we need to call some methods to get the internal fields initialized
+                    $request->getMethod();
+                    $request->getPathInfo();
+                    $that->assertEquals($hashRequest, $request);
+
+                    return true;
+                }),
+                $catch
```
dbu: what is this doing? and why not just pass true literally?

dbu:
coming good, thanks!
@ddeboer can you also have a look at the test and think whether this is exactly what should happen? and do you agree with removing the complete cookie header if there is no single session cookie?

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
sprain: To be honest: I have no idea. I just tried to adjust the very complicated existing test from the session approach.

src/SymfonyCache/UserContextSubscriber.php
```diff
@@ -180,6 +183,16 @@ private function getUserHash(HttpKernelInterface $kernel, Request $request)
      */
     private function isAnonymous(Request $request)
     {
+        // You might have to enable rewriting of the Authorization header in your server config or .htaccess:
+        // RewriteEngine On
+        // RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
+        if ($request->server->has('AUTHORIZATION') ||
+            $request->server->has('HTTP_AUTHORIZATION') ||
+            $request->server->has('PHP_AUTH_USER'))
+        {
```
dbu: CS: please take the last `)` on the same line as the `{`, so that we have `) {`

src/SymfonyCache/UserContextSubscriber.php
dbu: @ddeboer do you agree that we should do an 
    } else {
         $hashLookupRequest->headers->remove('Cookie');
    }

dbu: or $hashLookupRequest->cookies->replace(array())?

dbu: oh, sorry. forget about this. we simply do not copy the cookies if there are none to copy. this is all good.

src/SymfonyCache/UserContextSubscriber.php
```diff
@@ -130,7 +130,10 @@ protected function cleanupHashLookupRequest(Request $hashLookupRequest, Request
                 $hashLookupRequest->cookies->set($name, $value);
             }
         }
-        $hashLookupRequest->headers->set('Cookie', http_build_query($sessionIds, '', '; '));
+
+        if (count($sessionIds) > 0) {
+            $hashLookupRequest->headers->set('Cookie', http_build_query($sessionIds, '', '; '));
```
dbu: @lolautruche do you know why we need to set the header explicitly, in addition to manipulating the cookies bag? isn't this doing things twice?

lolautruche: Setting the `Cookie` header is mandatory to make the response vary on it. Setting up the cookie bag is just for convenience for further down manipulation.

dbu: so the Request class has no convenience method to sync the two? that seems a bit unfortunate. but not something we can solve here.

lolautruche: Indeed...

ddeboer:
@dbu Where is the cookie header removed completely?

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
```diff
@@ -177,6 +177,58 @@ public function testUserHashUserWithSession($arg, $options)
...
+
+        // Ensure request properties have been filled up.
+        $hashRequest->getPathInfo();
+        $hashRequest->getMethod();
+
+        $expectedContextHash = 'my_generated_hash';
+        // Just avoid the response to modify the request object, otherwise it's impossible to test objects equality.
+        /** @var \Symfony\Component\HttpFoundation\Response|\PHPUnit_Framework_MockObject_MockObject $hashResponse */
+        $hashResponse = $this->getMockBuilder('\Symfony\Component\HttpFoundation\Response')
+            ->setMethods(array('prepare'))
```
ddeboer: Can be removed.

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
```diff
@@ -177,6 +177,58 @@ public function testUserHashUserWithSession($arg, $options)
...
+            ->getMock();
+        $hashResponse->headers->set($options['user_hash_header'], $expectedContextHash );
```
ddeboer: You don't need a mock here: just construct a `Response` object directly with the context header.

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
ddeboer: Argument `$catch` can be removed.

ddeboer:
Added some comments on the test; otherwise it’s valid. 

@sprain Can you update the test and fix the coding standard violation mentioned by @dbu so we can merge this?

dbu:
@ddeboer unless i misread https://github.com/sprain/FOSHttpCache/blob/master/src/SymfonyCache/UserContextSubscriber.php#L226 severely, there never is a cookie header on the hash lookup request, unless we manually set it in that method where i commented (and then removed my comments because i thought it less confusing.)

sprain:
@ddeboer Updated.

dbu:
@sprain can you please add a cookie header in the test (with a cookie that is *not* a session cookie), and assert that its *not* present in the hash lookup? if the basic auth header is copied along, i fear the cookie header might actually be copied as well...

sprain:
@dbu @ddeboer Done!

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
```diff
@@ -160,6 +163,7 @@ public function testUserHashUserWithSession($arg, $options)
                     $request->getMethod();
                     $request->getPathInfo();
                     $that->assertEquals($hashRequest, $request);
+                    $that->assertSame(2, count($request->cookies->all()));
```
ddeboer: `assertCount` is shorter

ddeboer:
@dbu Looks good to me now. Let’s merge?

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
dbu: +1 for assertCount - its also more expressive when failing

dbu:
looks good! so the cookie is not propagated. good we have a test that proves it.

now the last thing that is missing is the doc. @sprain can you please update https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/doc/symfony-cache-configuration.rst#user-context accordingly? right now its rather implicit that there is only session with cookies support - we should make it explicit that it can be both basic auth or a session cookie. think you manage that one? otherwise i can have a go at that.

tests/Unit/SymfonyCache/UserContextSubscriberTest.php
sprain: @ddeboer done

sprain:
@dbu Updated the docs. Please have a look at contents and format. I am not skilled with rst as I much more often use markdown.

doc/symfony-cache-configuration.rst
```diff
@@ -160,6 +161,16 @@ constructor:
     user context hash for every users, or not having it cached at all, which
     hurts performance.
 
+.. note::
+
+    To use authorization headers for user context, you might have to add some server
+    configuration to make these headers available to php.
```
dbu: s/php/PHP/

dbu:
looks good to me. the spellchecker complains about `php` - not sure if `PHP` will be recognized - otherwise we just add PHP to the list of extra words it should know about.

sprain:
@dbu Updated to `PHP`.

dbu:
i did a tiny cleanup on the unit test in 437e05c8df59f7863103b2481109e4995f81a66f (after convincing myself that this really is all correct)

thanks a lot @sprain !

PR #168, opened 2015-02-07
cleanup hash lookup request

### LINKED ISSUE #165: Inconsistent default for context hash lookup url
The default URL for the user context hash lookup in varnish is [/user_context_hash.php](http://foshttpcache.readthedocs.org/en/latest/varnish-configuration.html#user-context) but for the symfony http cache subscriber it is [/_fos_user_context_hash](http://foshttpcache.readthedocs.org/en/latest/symfony-cache-configuration.html#user-context).

To make matters worse, the FOSHttpCacheBundle recommends [/user-context-hash](http://foshttpcachebundle.readthedocs.org/en/latest/reference/configuration/user-context.html#context-hash-route) as default route, being not very specific that the default varnish config is different. the symfony http cache instantiates the subscriber with default options, so its /_fos_user_context_hash.

I suggest that we change this to be the same everywhere. I like /user-context-hash best - the library backend hopefully does not consist of separate php files. there would be no BC break for the bundle configuration, and only a change in recommended configuration for the library when using varnish. it would be a BC break for the symfony httpcache subscriber however.

dbu:
fix #165 

/_fos_user_context_hash is the only default that involves no BC break. the Symfony cache is using this url by default.

### DIFF
--- a/doc/user-context.rst
+++ b/doc/user-context.rst
@@ -27,7 +27,7 @@ Caching on user context works as follows:
 2. The :term:`caching proxy`  receives the request. It sends a request
    (the *hash request*) with a special accept header
    (``application/vnd.fos.user-context-hash``) to a specific URL,
-   e.g., ``/user_context_hash.php``.
+   e.g., ``/_fos_user_context_hash``.
 3. The :term:`application` receives the hash request. The application knows the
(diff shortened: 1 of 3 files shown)


dbu:
if we can agree on this, i will also cleanup FOSHttpCacheBundle to be consistent

PR #169, opened 2015-02-07
cleanup usage of event dispatcher with CacheInvalidator

### LINKED ISSUE #149: Why does CacheInvalidator have event dispatcher methods?
It seems wrong that the CacheInvalidator implements the `getEventDispatcher`, `addSubscriber` and even `setEventDispatcher` methods.

 If `EventDispatcher` is a dependency then it should be injected in the constructor. Subscribers should be added to that dispatcher and the CacheInvalidator class should not be a proxy.

dbu:
fix #149 
/cc @dantleech

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+* **2015-02-11** It is no longer possible to change the event dispatcher of the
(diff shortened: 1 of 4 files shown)


dantleech:
+1

dbu:
added a note to the changelog. good to merge?

PR #171, opened 2015-02-07
addTags on TagHandler

dbu:
follow up of #162

This allows TagHandler to track tags. I tried adding a flush method that calls `header()` but feel there is little value to that. See the doc diff for how i propose to use this outside symfony.

### DIFF
--- a/doc/invalidation-handlers.rst
+++ b/doc/invalidation-handlers.rst
@@ -9,9 +9,15 @@ to simplify common operations.
 Tag Handler
 -----------
 
-The tag handler helps you to invalidate all cache entries that where marked
-with a specified tag. It works only with a ``CacheInvalidator`` that supports
-``CacheInvalidator::INVALIDATE``.
+.. versionadded:: 1.3
+
+    The tag handler was added in FOSHttpCache 1.3. If you are using an older
+    version of the library and can not update, you need to use
+    ``CacheInvalidator::invalidateTags``.
+
+The tag handler helps you to mark responses with tags that you can later use to
+invalidate all cache entries with that tag. Tag invalidation works only with a
+``CacheInvalidator`` that supports ``CacheInvalidator::INVALIDATE``.
 
 Setup
 ~~~~~
@@ -34,8 +40,22 @@ Usage
 
 With tags you can group related representations so it becomes easier to
 invalidate them. You will have to make sure your web application adds the
-correct tags on all responses by setting the ``X-Cache-Tags`` header. The
-FOSHttpCacheBundle_ does this for you when you’re using Symfony.
+correct tags on all responses. You can add tags to the handler using::
+
+    $tagHandler->addTags(array('tag-two', 'group-a'));
+
+Before any content is sent out, you need to send the tag header_::
+
+    header(sprintf('%s: %s),
+        $tagHandler->getTagsHeaderName(),
+        $tagHandler->getTagsHeaderValue()
+    );
+
+.. tip::
+
+    If you are using Symfony with the FOSHttpCacheBundle_, the tag header is
+    set automatically. You also have `additional methods of defining tags`_ with
+    annotations and on URL patterns.
 
 Assume you sent four responses:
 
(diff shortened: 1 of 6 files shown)


src/Handler/TagHandler.php
```diff
@@ -50,6 +55,29 @@ public function __construct(CacheInvalidator $invalidator, $tagsHeader = 'X-Cach
...
+        $this->tags = array_merge($this->tags, $tags);
+    }
+
+    /**
+     * Set tags on the response.
+     *
+     * This must be called before any response is sent to the client.
+     * Make sure this is called after all addTags calls and only once.
+     */
+    public function flushTags()
```
dbu: this is for the library only, in the bundle we will have something using Response.

src/Handler/TagHandler.php
```diff
@@ -59,10 +87,48 @@ public function __construct(CacheInvalidator $invalidator, $tagsHeader = 'X-Cach
...
+     * Make sure that the tags are valid.
+     *
+     * @param array $tags The tags to escape.
+     *
+     * @return array Sane tags.
+     */
+    protected function escapeTags(array $tags)
+    {
+        array_walk($tags, function (&$tag) {
+            $tag = str_replace(array(',', "\n"), array('_', '_'), $tag);
```
dbu: we missed doing this before, leading to `,` in a tag breaking the tagging. not sure if we need \n or not

tests/Unit/Handler/TagHandlerTest.php
```diff
@@ -62,4 +78,22 @@ public function testInvalidateUnsupported()
...
+            ->shouldReceive('supports')
+            ->with(CacheInvalidator::INVALIDATE)
+            ->once()
+            ->andReturn(true)
+            ->getMock();
+
+        $tagHandler = new TagHandler($cacheInvalidator);
+        $tagHandler->addTags(array('post-1', 'test,post'));
+        $tagHandler->flushTags();
+        // it seems on cli we can not investigate the headers_list
```
dbu: looks like we would need an actual functional test if we want to test this.

doc/invalidation-handlers.rst
```diff
@@ -9,9 +9,15 @@ to simplify common operations.
 Tag Handler
 -----------
 
-The tag handler helps you to invalidate all cache entries that where marked
-with a specified tag. It works only with a ``CacheInvalidator`` that supports
-``CacheInvalidator::INVALIDATE``.
+.. versionadded:: 1.3
```
dbu: this does not render as nicely as expected, because our sphinx setup does not load basic.css but only theme.css - @ddeboer would you know why we do not load default.css and what i could do to get the

    .versionmodified {
        font-style: italic;
    }

into our css, or load basic.css?

src/Handler/TagHandler.php
```diff
@@ -59,10 +98,26 @@ public function __construct(CacheInvalidator $invalidator, $tagsHeader = 'X-Cach
...
     }
+
+    /**
+     * Make sure that the tags are valid.
+     *
+     * @param array $tags The tags to escape.
+     *
+     * @return array Sane tags.
+     */
+    protected function escapeTags(array $tags)
```
dbu: i think protected makes sense here, as an extending class might want to use this for other operations. or a special setup might need additional escaping.

dbu:
i propose we bump the branch-alias to version 1.3 and tag 1.3.0-alpha1 after merging this. then i can work on the bundle and once that is good and we need no further tweaks on TagHandler, we can tag 1.3.0 stable.

doc/invalidation-handlers.rst
ddeboer: We use the Read the Docs theme (`theme.css`) on local builds too, so you can see what it will really look like. Unfortunately, [they don’t support the version directives](https://github.com/snide/sphinx_rtd_theme/issues/99). Pushed a commit that at adds some basic formatting for `versionadded/versionchanged`.

src/Handler/TagHandler.php
ddeboer: Ok.

ddeboer:
Merged and... whoops, tagged 1.3.0 stable already. ;) Let’s hope you don't need changes to `TagHandler`, otherwise jumping to 1.4.0 shouldn’t be a problem.

dbu:
cool, thanks!

Issue #163, opened 2015-01-18
For version 2: Drop all deprecated methods and cleanup doc


dbu:

ddeboer:
As far as I know, the `@deprecated` annotation only warns users whose IDE supports it or when they thoroughly look through our code (which they shouldn't have to do). Should we therefore also add `trigger_error()` at some stage? This, however, did cause major pain when Symfony added them.

dbu:
i would only do the annotation and not bother with trigger_error. people who did not notice the deprecation will see fatal error which works for debugging too :-/

ddeboer:
All right.

ddeboer:
- [ ] convert to short array syntax

PR #173, opened 2015-02-20
Decouple from HTTP client library

### LINKED ISSUE #53: Guzzle 4
We could upgrade our Guzzle implementation to their recently released version 4, which allegedly is [faster](http://mtdowling.com/blog/2014/03/15/guzzle-4-rc/). However, it requires PHP 5.4. Can we already drop our PHP 5.3 support? :)

ddeboer:
Fix #53.

Goals:
* Completely decouple our caching proxy clients from any HTTP client implementation.
* Remove dependency on Guzzle (or any other) HTTP client.
* Depend only on PSR-7 and the generic php-http/adapter instead.

Todo:
- [x] update docs to reflect changes
- [x] fix missing host test
- [x] Switch RTD default to stable, so users won't be confused by our updated (2.0) docs


doc/cache-invalidator.rst
```diff
@@ -195,6 +195,9 @@ implements ``\Psr\Log\LoggerInterface``. For instance, when using Monolog_::
 Then add the logger as a subscriber to the cache invalidator::
 
     use FOS\HttpCache\EventListener\LogSubscriber;
+    use FOS\HttpCache\CacheInvalidator;
+    
+    $invalidator = new CacheInvalidator(
```
dbu: incomplete

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -61,58 +50,24 @@
...
-        $this->servers = array();
-        foreach ($servers as $server) {
-            $this->servers[] = $this->filterUrl($server, array('scheme', 'host', 'port'));
+        if ($client instanceof ClientInterface) {
+            // Only for BC; Guzzle 3 for PHP 5.3 compatibility
+            $this->client  = new Guzzle3Adapter($client);
+        } elseif ($client instanceof HttpClientInterface) {
+            $this->client = $client;
+        } elseif (null === $client) {
+            $this->client = new Guzzle3Adapter();
```
dbu: should we determine which client is available?

ddeboer: You mean defaulting to Guzzle 5 (or Buzz) when we find that? Makes sense to much this logic into a small factory, then.

dbu:
interesting. kind of sounds like the client abstraction should be a PSR that builds on top of PSR-7... but i guess for now we have to live with that.

i am not so sure if this is still a 1.x thing or we should bump to 2.0 for this. looks like a couple of protected methods severly change behaviour.

ddeboer:
Yeah, I tried to limit BC issues by falling back to our current Guzzle 3 implementation, but it’s undoable to have no breaks at all. I’m okay with postponing this to 2.0, but would like to finish this PR up shortly to get it over and done with. But I’m against opening a 2.0 branch as we decided to limit parallel branches as much as possible. Do you agree with changing the branch alias to 2.0 in this PR? I think we have some things that we should finish in 1.0 first, so we should hold off merging this for a little while.

dbu:
i agree to bump the version to 2 in this and make 1.3 the last 1.x version of the bundle. we should withhold from tagging 2.0.0 for a bit after merging this however, to give it time to settle and give us some time for more bc breaks if needed.

about the whole http client abstraction layer: i wonder if it would be worth to move that to its own repository. then again, that will explode the complexity of the abstraction, as now we only need to provide the minimum we need to operate.

stof:
There is already several such libraries: https://packagist.org/search/?q=http-adapter (and probably others with different names)

And given that PSR-7 is close, I think the best solution is probably to expect client with a single method ``->request(PSR7Request $request): PSR7Response``.
I don't know whether there is already an HTTP client library based on PSR-7 (either on its own or bridging existing libraries)

dbu:
thanks @stof, indeed a lot of choice. the egolen / ivory one looks quite good at first glance. is that a good choice, or do you have any recommendations which one to pick?

and any idea how risky it is to implement PSR7 right now? if the interface changes before its finally accepted, we would become incompatible. and do you mean to write a wrapper around that wrapper? ivory seems to provide a sendRequest method that accepts a PSR request object...

dbu:
https://github.com/saxulum/saxulum-http-client-interface another suggestion that came through twitter

ddeboer:
Yes, let’s use one of the already existing HTTP adapters; it makes no sense to reinvent the wheel. I like ivory-http-adapter as it’s compatible with PSR-7 and has parallel requests, which we need.

We do require some extra functionality on top of pure HTTP messages:
* [calculating the message signature](https://github.com/FriendsOfSymfony/FOSHttpCache/pull/173/files#diff-fa8ecfb5cfebb3c2f45eb4f8e9137723R60) 
* replicating the invalidation requests over all proxy instances, which I do in a [RequestQueue](https://github.com/FriendsOfSymfony/FOSHttpCache/pull/173/files#diff-1e0c9dd970de0ebe32c99a9b23cc7617). 

I guess I could make InvalidationRequest extend PSR’s [RequestInterface](https://github.com/php-fig/fig-standards/blob/master/proposed/http-message.md#32-psrhttpmessagerequestinterface) to offer this functionality.

dbu:
that sounds like a good plan to me.

ddeboer:
Okay, switched to the egloen/http-adapter for HTTP library abstraction. Unfortunately, that lib relies on phly/http for its PSR-7 implementation, and phly/http does not allow PURGE as a HTTP method (which makes sense in a way, as it’s not proper HTTP, but still). I’m still looking for a way to work around this.

dbu:
looks good!

voting for PSR-7 went into another iteration (though about documentation rather than code as far as i understood). i think we best not release any 2.0 version before it is final and we can rely on the official interfaces. lets hope this will happen within the next month or two.

ddeboer:
Yes, let’s wait for PSR-7. While working on this PR, egeloen/http-adapter has changed its API multiple times. I hope that with PSR-7 the dust will settle and both HTTP adapters and message library (phly/http) become stable enough to use them here.

ddeboer:
Update: ivory-http-adapter has been split into multiple packages and moved to a separate organisation: [php-http](https://github.com/php-http).

composer.json
```diff
@@ -20,15 +20,24 @@
             "homepage": "https://github.com/friendsofsymfony/FOSHttpCache/contributors"
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/ddeboer/guzzle6-adapter"
```
ddeboer: Should be removed when the adapter has been [merged](https://github.com/php-http/guzzle6-adapter/pull/1).

composer.json
```diff
@@ -20,15 +20,24 @@
...
+    ],
     "require": {
-        "php": ">=5.3.3",
-        "guzzle/guzzle": "~3.8",
+        "php": ">=5.4.8",
         "symfony/event-dispatcher": "~2.3",
-        "symfony/options-resolver": "~2.3"
+        "symfony/options-resolver": "~2.3",
+        "php-http/adapter": "dev-internal_separation",
+        "guzzlehttp/psr7": "~1.0"
```
ddeboer: I use the Request implementation provided by guzzlehttp/psr7 to construct internal requests to the caching proxy servers.

composer.json
```diff
@@ -20,15 +20,24 @@
...
     },
     "require-dev": {
         "mockery/mockery": "~0.9.1",
         "monolog/monolog": "~1.0",
+        "php-http/adapter-core": "dev-internal_separation",
+        "php-http/guzzle6-adapter": "dev-implementation",
```
ddeboer: Should be changed after adapter has been merged.

src/ProxyClient/Request/RequestQueue.php
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        if (isset($parts['path'])) {
+            $this->basePath = $parts['path'];
+        }
+    }
+    
+    public function getAllowedSchemes()
+    {
+        return array('http');
```
ddeboer: As this is different for Nginx and Varnish, this should probably be moved back into the Varnish and Nginx client classes.

composer.json
```diff
@@ -45,5 +54,7 @@
         "branch-alias": {
             "dev-master": "1.3.x-dev"
         }
-    }
+    },
+    "minimum-stability": "dev",
+    "prefer-stable": true
```
ddeboer: Should be removed when we can depend on stable php-http packages.

composer.json
dbu: as we use guzzle anyways that sounds right to me. otherwise we could see if symfony will provide one as well.

doc/cache-invalidator.rst
```diff
@@ -195,7 +195,7 @@ implements ``\Psr\Log\LoggerInterface``. For instance, when using Monolog_::
 Then add the logger as a subscriber to the cache invalidator::
 
     use FOS\HttpCache\EventListener\LogSubscriber;
-
+
```
dbu: extra whitespace

src/Exception/ExceptionCollection.php
```diff
@@ -15,9 +15,16 @@
...
  */
-class ExceptionCollection extends \Exception implements \IteratorAggregate, \Countable,  HttpCacheExceptionInterface
+class ExceptionCollection extends \Exception implements \IteratorAggregate, \Countable, HttpCacheExceptionInterface
 {
     private $exceptions = array();
+    
+    public function __construct(array $exceptions = array())
+    {
+        foreach ($exceptions as $exception) {
+            $this->add($exception);
```
dbu: should the add method become private? kind of unrelated to this PR but when we start refactoring for 2.0 anyways...

src/Exception/ProxyResponseException.php
```diff
@@ -11,32 +11,27 @@
 
 namespace FOS\HttpCache\Exception;
 
+use Http\Adapter\Exception\HttpAdapterException;
```
dbu: is PSR7 not in a Psr namespace? or is this not from PSR7?

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -12,43 +12,34 @@
...
-use Guzzle\Http\ClientInterface;
-use Guzzle\Http\Exception\CurlException;
-use Guzzle\Common\Exception\ExceptionCollection as GuzzleExceptionCollection;
-use Guzzle\Http\Exception\RequestException;
-use Guzzle\Http\Message\RequestInterface;
+use FOS\HttpCache\ProxyClient\Request\InvalidationRequest;
+use FOS\HttpCache\ProxyClient\Request\RequestQueue;
+use Http\Adapter\Exception\MultiHttpAdapterException;
+use Http\Adapter\Guzzle6HttpAdapter;
+use Http\Adapter\PsrHttpAdapter;
```
dbu: is that from ivory? could we use an interface from psr instead?

src/ProxyClient/Request/InvalidationRequest.php
```diff
@@ -0,0 +1,70 @@
...
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * A request to the HTTP caching server
+ */
+class InvalidationRequest
+{
+    
+    
+    private $method;
```
dbu: missing doc and extra blank lines

src/ProxyClient/Request/RequestQueue.php
dbu: yeah lets make this method abstract here.

dbu: ah sorry. no, lets move it to the clients.

src/ProxyClient/Varnish.php
```diff
@@ -125,28 +125,36 @@ public function refresh($url, array $headers = array())
...
-        $request = parent::createRequest($method, $url, $headers);
-
-        // For purge and refresh, add a host header to the request if it hasn't
-        // been set
-        if (self::HTTP_METHOD_BAN !== $method
-            && '' == $request->getHeader('Host')
-        ) {
-            throw MissingHostException::missingHost($url);
+        if (self::HTTP_METHOD_BAN !== $method) {
+            // AND NO HOST IS SET
```
dbu: ?

src/ProxyClient/Varnish.php
```diff
@@ -125,28 +125,36 @@ public function refresh($url, array $headers = array())
...
+    
+
+//    /**
+//     * {@inheritdoc}
+//     *
+//     * @throws MissingHostException If a relative path is queued for purge/
+//     *                              refresh and no base URL is set
+//     *
+//     */
+//    protected function queueRequest($method, $url, array $headers = array())
```
dbu: this needs to be finished or removed if not needed anymore

src/Exception/ProxyResponseException.php
ddeboer: No, this is the HTTP client adapter package that we use: https://github.com/php-http. Remember that PSR7 is only about HTTP *messages*.

src/ProxyClient/AbstractProxyClient.php
ddeboer: There is no HTTP client interface in PSR7.

tests/Functional/NginxTest.php
```diff
@@ -74,7 +74,10 @@ public function testRefresh()
         $nginx->refresh('/cache.php')->flush();
         usleep(1000);
         $refreshed = $this->getResponse('/cache.php');
-        $this->assertGreaterThan((float) $response->getBody(true), (float) $refreshed->getBody(true));
+        $this->assertGreaterThan(
+            (float)(string) $response->getBody(),
```
dbu: does this make sense? why not just `(float)`?

src/ProxyClient/Varnish.php
ddeboer: This is somewhat problematic: I need to know here whether the request has a Host header. That header is determined from either the base URL or the endpoint of the request itself (in case of an absolute URL). I moved base URL handling into a `RequestQueue`, so I cannot check that here. I guess this logic needs to be moved back to the Varnish and Nginx client classes.

dbu:
wow, huge thing! should we start a 2.0 branch so that we can merge this soonish? would be great if we get the symfony cache client in before that though.

tests/Functional/NginxTest.php
ddeboer: We need `(string)` to get the response body from the [StreamInterface](https://github.com/php-fig/http-message/blob/master/src/StreamInterface.php#L28), then the original `(float)` for correct timestamp comparison.

dbu:
also, i think we need some coordinated effort for 2.0 to not have too big pull requests, so we can merge them quickly, and not do too many things in the 1.x once 2.0 has started, to avoid complicated merges from 1.x to 2 branch. maybe we can discuss that at DPC?

ddeboer:
Yeah, let’s open up a 2.0 branch. I guess that needs closing and re-creating this PR against the new branch?

tests/Functional/NginxTest.php
dbu: okay

src/Exception/ProxyResponseException.php
dbu: ah ok, so the thing formerly known as ivory?

ddeboer: Exactly.

dbu:
> I guess that needs closing and re-creating this PR against the new branch?

yes, for whatever reason github does not allow to change the target
branch once the PR exists. (gitlab for example allows that, and i see no
technical reason why that should not be allowed). at least you can keep
using the same branch.

ddeboer:
Should we alias master to 2.0.x-dev or create a new 2.0/2.x branch and leave master at 1.0?

dbu:
i would say that depends on how many pull requests we expect for 1.x . if we get to 2.0 within a month, switching master sounds good. if it takes us longer, we should have a branch, so that small contributions to 1.x stay simple.

looking at the list of things, it seems not that huge. so if we make master 2.0.x-dev once we merge this pull request, it would seem realistic to me. unless it turns out we want to refactor other important bits not related to the http client. wdyt?

ddeboer:
Not huge, no. I guess https://github.com/FriendsOfSymfony/FOSHttpCache/labels/v2 is about it. Whether we can make it in one month depends in part on how development goes at https://github.com/php-http: in my opinion, it should be stable before we release our 2.0.

dbu:
okay. so either we leave this branch open until then, or we create a 2.0 branch that we merge this to. i would not want to have a half-finished client in master if we keep adapting to php-http changes. any ETA for php-http to become stable?

the other things should indeed go quick so once that is ready we can just change master to 2.0.x-dev.

composer.json
sagikazarmark: It is now.

composer.json
stof: We won't

stof: FYI, the PSR-7 bridge relies on the ZF implementation of PSR-7 by default

sagikazarmark: It isn't now. You can use any PSR-7 implementation.

sagikazarmark:
> I guess I could make InvalidationRequest extend PSR’s RequestInterface to offer this functionality.

I have created a [message decorator](https://github.com/php-http/message-decorator) implementation which lets you adding custom logic to messages without extending a specific implementation or implementing the whole PSR-7 interface set.

@dbu we don't have an ETA for stable right now. I scheduled a 0.1 release for tomorrow evening. All WIP branches have been merged. I plan to release the adapter in parts: since we separated all our components into separate packages, we can release them as soon as they are ready. The adapter contract and the example implementations (guzzle 5 and 6) are top priority, so I don't think you have to wait for too long.

composer.json
dbu: is the request factory in php-http generic enough that we can not require a specific implementation? can it somehow autodiscover what implementation is available? or do we need one concrete implementation so that we can create requests if no request factory is injected to the client?

ddeboer: php-http will have a MessageFactoryGuesser for autodiscovery.

sagikazarmark: It is tricky. ;)

We have a message factory interface which allows you to pass a custom implementation. However, you don't have to. We have a MessageFactoryGuesser implementation which autodiscovers a few common implementations (currently guzzle and zend).

Also, the message factory is only needed for non PSR-7 adapters.

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -61,58 +52,16 @@
...
-        $this->client->getBaseUrl();
+     * @param PsrHttpAdapter $httpAdapter If no HTTP adapter is supplied, a
+     *                                 default one will be created.
+     */
+    public function __construct(
+        array $servers,
+        $baseUrl = null,
+        PsrHttpAdapter $httpAdapter = null
+    ) {
+        $this->httpAdapter = $httpAdapter ?: new Guzzle6HttpAdapter();
```
ddeboer: Perhaps replace `new Guzzle6HttpAdapter()` with an [HttpAdapterFactory](https://github.com/php-http/adapter/issues/25) if it becomes available.

composer.json
ddeboer: True, but we’d need it here to construct the `InvalidationRequest`.

src/ProxyClient/AbstractProxyClient.php
dbu: exactly. so either you pass the adapter you want, or let the factory
find one. if there is no valid adapter, we (or the factory) throws an
exception that explains you need some implementation and adapter in your
project.

ddeboer: Yes. This has the benefit that we don’t need to depend on a concrete adapter, such as guzzle6-adapter now (which is only required in dev, but that’s my mistake: it is clearly required here when no adapter is passed). As guzzle6-adapter depends on guzzle/guzzle, FOSHttpCache will still download Guzzle as a dependency. That’s one of the things that we try to avoid.

composer.json
sagikazarmark: You don't need to use it to create custom requests. Ideally, custom request should also be implementation agnostic (at least they will be in php-http, see my comment bellow, how). I will show you an example about custom messages once I finished the client.

src/ProxyClient/AbstractProxyClient.php
sagikazarmark: Actually, you can handle this with composer. Each adapter implementation provides a virtual package: `php-http/adapter-implementation`. You can require this virtual package instead of a concrete implementation. The factory should find all available clients and choose the best (which depends on personal preference, but lets say guzzle is the best and always use that if available)

ddeboer: Good point. So we need to depend on both `php-http/adapter-implementation` and `php-http/discovery`.

sagikazarmark: Exactly.

src/ProxyClient/Request/RequestQueue.php
ddeboer: Just wondering: does `getAllowedSchemes()` actually make sense? Nginx and Symfony already allow both `http` and `https`. What about Varnish? If you have a SSL terminator in front of Varnish (such as Nginx or Pound), couldn’t you also send invalidation requests to `https://your-app.com`?

src/Exception/ExceptionCollection.php
ddeboer: See #163.

src/ProxyClient/Request/RequestQueue.php
ddeboer: And related to this, should we perhaps remove the scheme guessing (from `getDefaultScheme()`) when setting a baseUrl as well? I think it may prevent some confusion.

ddeboer: On the other hand, when sending invalidation requests, it is not the baseUrl’s scheme that is used, but that of the configured `$servers`.

src/ProxyClient/Varnish.php
ddeboer: The `MissingHostException` makes sense: *usually* you don’t want to accidentally send purge/refresh requests to `127.0.0.1` if your app runs on `my-app.com` (because the hostnames is included in the hash, so the purged `127.0.0.1/purge-this` does not match `my-app.com/127.0.0.1`). 

But what if I decide to run my app on an IP address? In that case, I would want to be able to send purges/refreshes to `127.0.0.1/purge-this`. Should we always throw this exception?

ddeboer:
Most of this now works, but I’m getting random test failures. For instance:

```
There was 1 failure:

1) FOS\HttpCache\Tests\Functional\VarnishProxyClientTest::testPurgeHost
Failed asserting that the responses are identical.

/vagrant/FOSHttpCache/src/Test/ProxyTestCase.php:46
/vagrant/FOSHttpCache/tests/Functional/Varnish/VarnishProxyClientTest.php:112
```

Relevant PHP:

```
111        $varnish->purge('http://localhost:6181/cache.php')->flush();
112        $this->assertMiss($this->getResponse('/cache.php'));
```

Or: 

```
There was 1 failure:

1) FOS\HttpCache\Tests\Functional\VarnishProxyClientTest::testPurgeContentType
Failed asserting that the responses are identical.

/vagrant/FOSHttpCache/src/Test/ProxyTestCase.php:46
/vagrant/FOSHttpCache/tests/Functional/Varnish/VarnishProxyClientTest.php:104
```

These and other functional Varnish tests randomly succeed and randomly fail. The general error cause seems to be: the first request after a PURGE/BAN is a HIT, but it should be a MISS (that’s the failing assertion above). In other words: the response is not properly purged/banned.

I have several indications that this is something on the Varnish-side:

1. When checking the Varnish logs (see below), I see proper entries for the GET requests that check for hits and misses, but *there is no entry for the `PURGE` request*.
2. Switching between cURL and Stream handlers in Guzzle 6 didn’t make a difference.
3. Raising Varnish memory (param `-l 200M` instead of default 80M) makes the problem go away on my dev machine, but not completely on Travis with PHP 5.6.

The problem reminds me of the one that was introduced when we started testing on Varnish 4 and that [went away by enabling `CURLOPT_FORBID_REUSE`](https://github.com/FriendsOfSymfony/FOSHttpCache/pull/98#issuecomment-49447288).

Varnish log output (for `testPurgeHost`):

```log
*   << Session  >> 1
-   Begin          sess 0 HTTP/1
-   SessOpen       127.0.0.1 59991 127.0.0.1:6181 127.0.0.1 6181 1433681901.680505 12
-   SessClose      REM_CLOSE 0.000
-   End

*   << BeReq    >> 4
-   Begin          bereq 3 fetch
-   Timestamp      Start: 1433681901.680735 0.000000 0.000000
-   BereqMethod    GET
-   BereqURL       /cache.php
-   BereqProtocol  HTTP/1.1
-   BereqHeader    User-Agent: GuzzleHttp/6.0.1 curl/7.38.0 PHP/5.6.7-1
-   BereqHeader    Host: localhost:6181
-   BereqHeader    X-Forwarded-For: 127.0.0.1
-   BereqHeader    Accept-Encoding: gzip
-   BereqHeader    X-Varnish: 4
-   VCL_call       BACKEND_FETCH
-   VCL_return     fetch
-   BackendOpen    13 default(127.0.0.1,,8080) 127.0.0.1 47093
-   Backend        13 default default(127.0.0.1,,8080)
-   Timestamp      Bereq: 1433681901.687893 0.007158 0.007158
-   Timestamp      Beresp: 1433681901.687937 0.007202 0.000044
-   BerespProtocol HTTP/1.1
-   BerespStatus   200
-   BerespReason   OK
-   BerespHeader   Host: localhost:6181
-   BerespHeader   Connection: close
-   BerespHeader   X-Powered-By: PHP/5.6.7-1
-   BerespHeader   Cache-Control: max-age=3600
-   BerespHeader   Content-type: text/html;charset=UTF-8
-   BerespHeader   X-Cache-Debug: 1
-   BerespHeader   Date: Sun, 07 Jun 2015 12:58:21 GMT
-   TTL            RFC 3600 -1 -1 1433681902 1433681902 1433681901 0 3600
-   VCL_call       BACKEND_RESPONSE
-   TTL            VCL 3600 0 0 1433681902
-   VCL_Log        GRACE 0.000
-   BerespHeader   X-Url: /cache.php
-   BerespHeader   X-Host: localhost:6181
-   VCL_return     deliver
-   Storage        malloc s0
-   ObjProtocol    HTTP/1.1
-   ObjStatus      200
-   ObjReason      OK
-   ObjHeader      Host: localhost:6181
-   ObjHeader      X-Powered-By: PHP/5.6.7-1
-   ObjHeader      Cache-Control: max-age=3600
-   ObjHeader      Content-type: text/html;charset=UTF-8
-   ObjHeader      X-Cache-Debug: 1
-   ObjHeader      Date: Sun, 07 Jun 2015 12:58:21 GMT
-   ObjHeader      X-Url: /cache.php
-   ObjHeader      X-Host: localhost:6181
-   Fetch_Body     4 eof stream
-   Timestamp      BerespBody: 1433681901.688070 0.007335 0.000133
-   BackendClose   13 default(127.0.0.1,,8080)
-   Length         15
-   BereqAcct      168 0 168 173 0 173
-   End

*   << Request  >> 3
-   Begin          req 2 rxreq
-   Timestamp      Start: 1433681901.680656 0.000000 0.000000
-   Timestamp      Req: 1433681901.680656 0.000000 0.000000
-   ReqStart       127.0.0.1 59992
-   ReqMethod      GET
-   ReqURL         /cache.php
-   ReqProtocol    HTTP/1.1
-   ReqHeader      User-Agent: GuzzleHttp/6.0.1 curl/7.38.0 PHP/5.6.7-1
-   ReqHeader      Host: localhost:6181
-   ReqHeader      Connection: close
-   ReqHeader      X-Forwarded-For: 127.0.0.1
-   VCL_call       RECV
-   VCL_Log        RECV purge.vcl METHOD GET URL /cache.php
-   VCL_Log        RECV ban.vcl
-   VCL_return     hash
-   VCL_call       HASH
-   VCL_return     lookup
-   Debug          "XXXX MISS"
-   VCL_call       MISS
-   VCL_return     fetch
-   Link           bereq 4 fetch
-   Timestamp      Fetch: 1433681901.688147 0.007491 0.007491
-   RespProtocol   HTTP/1.1
-   RespStatus     200
-   RespReason     OK
-   RespHeader     Host: localhost:6181
-   RespHeader     X-Powered-By: PHP/5.6.7-1
-   RespHeader     Cache-Control: max-age=3600
-   RespHeader     Content-type: text/html;charset=UTF-8
-   RespHeader     X-Cache-Debug: 1
-   RespHeader     Date: Sun, 07 Jun 2015 12:58:21 GMT
-   RespHeader     X-Url: /cache.php
-   RespHeader     X-Host: localhost:6181
-   RespHeader     X-Varnish: 3
-   RespHeader     Age: 0
-   RespHeader     Via: 1.1 varnish-v4
-   VCL_call       DELIVER
-   RespHeader     X-Cache: MISS
-   VCL_return     deliver
-   Timestamp      Process: 1433681901.688187 0.007532 0.000041
-   RespHeader     Content-Length: 15
-   Debug          "RES_MODE 2"
-   RespHeader     Connection: close
-   RespHeader     Accept-Ranges: bytes
-   Timestamp      Resp: 1433681901.688656 0.008000 0.000469
-   Debug          "XXX REF 2"
-   ReqAcct        122 0 122 353 15 368
-   End

*   << Session  >> 2
-   Begin          sess 0 HTTP/1
-   SessOpen       127.0.0.1 59992 127.0.0.1:6181 127.0.0.1 6181 1433681901.680647 12
-   Link           req 3 rxreq
-   SessClose      RESP_CLOSE 0.008
-   End

*   << Request  >> 6
-   Begin          req 5 rxreq
-   Timestamp      Start: 1433681901.697708 0.000000 0.000000
-   Timestamp      Req: 1433681901.697768 0.000060 0.000060
-   ReqStart       127.0.0.1 59994
-   ReqMethod      GET
-   ReqURL         /cache.php
-   ReqProtocol    HTTP/1.1
-   ReqHeader      User-Agent: GuzzleHttp/6.0.1 curl/7.38.0 PHP/5.6.7-1
-   ReqHeader      Host: localhost:6181
-   ReqHeader      Connection: close
-   ReqHeader      X-Forwarded-For: 127.0.0.1
-   VCL_call       RECV
-   VCL_Log        RECV purge.vcl METHOD GET URL /cache.php
-   VCL_Log        RECV ban.vcl
-   VCL_return     hash
-   VCL_call       HASH
-   VCL_return     lookup
-   Hit            2147483652
-   VCL_call       HIT
-   VCL_return     deliver
-   RespProtocol   HTTP/1.1
-   RespStatus     200
-   RespReason     OK
-   RespHeader     Host: localhost:6181
-   RespHeader     X-Powered-By: PHP/5.6.7-1
-   RespHeader     Cache-Control: max-age=3600
-   RespHeader     Content-type: text/html;charset=UTF-8
-   RespHeader     X-Cache-Debug: 1
-   RespHeader     Date: Sun, 07 Jun 2015 12:58:21 GMT
-   RespHeader     X-Url: /cache.php
-   RespHeader     X-Host: localhost:6181
-   RespHeader     X-Varnish: 6 4
-   RespHeader     Age: 0
-   RespHeader     Via: 1.1 varnish-v4
-   VCL_call       DELIVER
-   RespHeader     X-Cache: HIT
-   VCL_return     deliver
-   Timestamp      Process: 1433681901.698035 0.000326 0.000267
-   RespHeader     Content-Length: 15
-   Debug          "RES_MODE 2"
-   RespHeader     Connection: close
-   RespHeader     Accept-Ranges: bytes
-   Timestamp      Resp: 1433681901.698066 0.000358 0.000032
-   Debug          "XXX REF 2"
-   ReqAcct        122 0 122 354 15 369
-   End
```

dbu:
maybe we have a race condition here? as varnish has gotten more decoupled, i could imagine that it takes a short while for something to properly get invalidated for new requests. also not sure if grace mode could have an influence. maybe guzzle 6 is somehow faster and thus more likely to hit the race condition?

did you try with a period of sleep? maybe lets see with a full second. if that makes no difference, its not a race condition.

src/ProxyClient/Varnish.php
dbu: can we validate that either the constructor gets a host name/ip or $url contains one? even when purging 127.0.0.1 we should require this to be explicit?

src/ProxyClient/Request/RequestQueue.php
dbu: yeah, lets just make protocol required and not guess neither validate it. most likely varnish invalidation should go directly to varnish and not via the ssl terminator, but if somebody really needs that i don't see why we should forbid it.

ddeboer:
Yes, I had been playing around a bit with `sleep()` and that helps, so probably a race condition: in the original:

```php
    public function testPurgeHost()
    {
        $varnish = new Varnish(array('http://127.0.0.1:' . $this->getProxy()->getPort()));
        $this->getResponse('/cache.php');
        $varnish->purge('http://localhost:6181/cache.php')->flush();
        $this->assertMiss($this->getResponse('/cache.php'));   // this assertion randomly fails
    }
```

this solves the issue:

```php
    public function testPurgeHost()
    {
        $varnish = new Varnish(array('http://127.0.0.1:' . $this->getProxy()->getPort()));
        $this->getResponse('/cache.php');
        sleep(1);  // solves the issue
        $varnish->purge('http://localhost:6181/cache.php')->flush();
        $this->assertMiss($this->getResponse('/cache.php'));
    }
```

but this does not:

```php
    public function testPurgeHost()
    {
        $varnish = new Varnish(array('http://127.0.0.1:' . $this->getProxy()->getPort()));
        $this->getResponse('/cache.php');
        $varnish->purge('http://localhost:6181/cache.php')->flush();
        sleep(1);  // doesn’t help
        $this->assertMiss($this->getResponse('/cache.php'));
    }
```

dbu:
sounds a lot like that, yes. hm, what to do? its of course a bit of an artificial situation we create but i can think of scenarios where this would happen: i fetch a page that might be cached to see if it is outdated and upon deciding it is outdated, i invalidate it. this would lead to the same immediatly subsequent operations - and our varnish is not even under load or anything.

i propose we add the sleep (lets see if 100ms is also reliably enough) and open an issue at the varnish project to ask them if that is expected behaviour.

.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: php
 
 php:
-  - 5.3
-  - 5.4
```
ddeboer: As we use Guzzle 6 as our testing client, which requires PHP 5.5, we can no longer test 5.4 (we’re dropping 5.3 support anyway). Is this reason enough to switch to Guzzle 5? It’s, of course, a temporary problem: PHP 5.4 will be [EOL in 3 months](http://php.net/supported-versions.php).

composer.json
```diff
@@ -21,14 +21,18 @@
...
-        "php": ">=5.3.3",
-        "guzzle/guzzle": "~3.8",
+        "php": ">=5.4.8",
         "symfony/event-dispatcher": "~2.3",
-        "symfony/options-resolver": "~2.3"
+        "symfony/options-resolver": "~2.3",
+        "psr/http-message-implementation": "~1.0",
+        "php-http/adapter-implementation": "^0.1.0" ,
+        "php-http/discovery": "^0.1@dev",
+        "php-http/message-decorator": "^0.1.0"
```
ddeboer: Got rid of the Guzzle dependency. We now only have generic and virtual dependencies.

composer.json
```diff
@@ -43,7 +47,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.4.x-dev"
+            "dev-master": "2.0.x-dev"
```
ddeboer: Updating the alias. Do we still want to create a 1.0 branch? I say let’s just stick to our old plan of not maintaining parallel branches (as long we can get away with doing so) and upping master to 2.0.

.travis.yml
sagikazarmark: Being EOL doesn't mean you have to drop 5.4. Since in the HTTP interfaces nothing indicates that you need 5.5, I would say try to test against 5.4 as well, even if it means you need a lower version. I would do the following: If the travis php version is 5.4 then remove guzzle6-adapter and install guzzle5-adapter. The guesser will always find the class for you. This way you can test against 5.4. See how it is done in [php-http/discovery](https://github.com/php-http/discovery/blob/master/.travis.yml#L23)

sagikazarmark: Not that in this case you have to require a message implementation as well.

ddeboer: Good idea, going to do that.

src/ProxyClient/Request/RequestQueue.php
```diff
@@ -0,0 +1,232 @@
...
+
+use FOS\HttpCache\Exception\InvalidUrlException;
+use Http\Discovery\UriFactoryDiscovery;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * A queue of requests to be sent to the HTTP caching server
+ */
+class RequestQueue implements \Countable
```
ddeboer: Is this a good name? This class replicates each invalidation request over all proxy instances. Perhaps `RequestRouter`? `RequestReplicator` (hmm, sounds scary)?

sagikazarmark: I would play with the word 'batch' somehow. Also, if it is connected to invalidation, why isn't that word in the name? (like InvalidationQueue?)

tests/Unit/Test/Proxy/VarnishProxyTest.php
```diff
@@ -13,7 +13,7 @@
 
 use FOS\HttpCache\Test\Proxy\VarnishProxy;
 
-class VarnishTestCaseTest extends \PHPUnit_Framework_TestCase
+class VarnishProxyTest extends \PHPUnit_Framework_TestCase
```
ddeboer: Fixing the class name.

.travis.yml
```diff
@@ -24,9 +24,14 @@ branches:
     # Build maintenance branches for older releases if needed. such branches should be named like "1.2"
     - '/^\d+\.\d+$/'
 
+install:
```
ddeboer: Composer setup should be done in `install`, not `before_script`.

sagikazarmark: :+1:

.travis.yml
```diff
@@ -24,9 +24,14 @@ branches:
...
+  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" ]]; then composer remove "php-http/guzzle6-adapter" --dev --no-update; fi
+  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" ]]; then composer require "php-http/guzzle5-adapter:0.1.*@dev" --dev --no-update; fi
```
sagikazarmark: Guzzle 5 adapter 0.1 is also tagged.

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -120,189 +69,46 @@ protected function getBaseUrl()
...
-     * Get default scheme
-     *
-     * @return string
-     */
-    protected function getDefaultScheme()
+    
+    protected function initQueue(array $servers, $baseUrl)
     {
-        return 'http';
+        $this->queue = new RequestQueue($servers, $baseUrl);
```
ddeboer: No objections to constructing the queue object here? It would be better (in terms of DI) to pass one to the AbstractProxyClient’s constructor, but I think that may be confusing for users.

composer.json
dbu: we can always create a branch from the latest 1.4 tag if there is
something we need to fix urgently.

CHANGELOG.md
```diff
@@ -3,11 +3,21 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+Unreleased
```
dbu: i would label this 2.0.0  - we know already that it will be in that version.

composer.json
```diff
@@ -20,15 +20,34 @@
...
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/ddeboer/adapter-client.git"
+        },
+        {
+            "type": "vcs",
+            "url": "https://github.com/php-http/client.git"
```
dbu: is this and the helper not in packagist yet?

composer.json
```diff
@@ -20,15 +20,34 @@
             "homepage": "https://github.com/friendsofsymfony/FOSHttpCache/contributors"
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/ddeboer/adapter-client.git"
```
dbu: TODO: remove before merging this one, once this has been merged upstream

src/ProxyClient/Request/RequestQueue.php
```diff
@@ -0,0 +1,221 @@
...
+                    if ($this->baseUri->getPath() !== '') {
+                        $path = $this->baseUri->getPath() . '/' . ltrim($uri->getPath(), '/');
+                        $uri = $uri->withPath($path);
+                    }
+                }
+            }
+            
+            // Close connections to make sure invalidation (PURGE/BAN) requests
+            // will not interfere with content (GET) requests.
+            $request = $request->withUri($uri)->withHeader('Connection', 'Close');
```
ddeboer: So this has resolved the timing issues we had. I found that the issue only happens when a GET and PURGE request were sent to Varnish *over the same connection*. Closing each connection may be somewhat slower, but we already advise users to not purge in real time but after the response was send to the client (`kernel.terminate` in the case of Symfony).

.travis.yml
```diff
@@ -24,9 +24,14 @@ branches:
     # Build maintenance branches for older releases if needed. such branches should be named like "1.2"
     - '/^\d+\.\d+$/'
 
+install:
+  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" ]]; then composer remove "php-http/guzzle6-adapter" --dev --no-update; fi
+  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" ]]; then composer require "php-http/guzzle5-adapter" --dev --no-update; fi
+  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" ]]; then composer require "guzzlehttp/psr7" --dev --no-update; fi
```
dbu: i think guzzlehttp/psr7 is listed in composer.json anyways - still needed?

src/ProxyClient/Request/RequestQueue.php
```diff
@@ -0,0 +1,221 @@
...
+use Http\Discovery\UriFactoryDiscovery;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Stores each invalidation request and replicates it over all HTTP cache servers
+ *
+ * @author David de Boer <david@ddeboer.nl>
+ */
+class RequestQueue implements \Countable
```
dbu: i was thinking RequestCollection, but it is more than a collection. RequestReplicator sounds not that wrong. or something else that explains this kind of duplicates the requests in the collection for each server, without implying that it would be responsible of actually *sending* the request. RequestDuplicator is strange because its not really duplicate, and its not distribute as each request has to go to every server...

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -120,196 +69,58 @@ protected function getBaseUrl()
...
      *
-     * @return string
+     * @param array  $servers
+     * @param string $baseUri
      */
-    protected function getDefaultScheme()
+    protected function initQueue(array $servers, $baseUri)
     {
-        return 'http';
+        $this->queue = new RequestQueue($servers, $baseUri);
```
dbu: i don't see why we would want to inject this. this is no external dependency so it seems fine to me. and there is still the option to overwrite the method for customization.

src/ProxyClient/Varnish.php
```diff
@@ -133,26 +142,28 @@ public function refresh($url, array $headers = array())
...
     /**
      * {@inheritdoc}
+     *
+     * Additionally, store whether a base URI was set
      */
-    protected function getAllowedSchemes()
+    protected function initQueue(array $servers, $baseUri)
     {
-        return array('http');
+        $this->queue = new RequestQueue($servers, $baseUri);
```
dbu: why not `parent::initQueue`?

src/ProxyClient/Varnish.php
```diff
@@ -133,26 +142,28 @@ public function refresh($url, array $headers = array())
...
+        $this->baseUriSet = $baseUri !== null;
```
dbu: could we instead check that information on the queue?

and why is that only relevant for varnish but not for the other clients? don't all clients need to know to what server they should talk?

dbu: but if its correct, i would set the value in the constructor instead, it has nothing to do with initializing the queue.

src/ProxyClient/Request/RequestQueue.php
dbu: isn't this closing the connection after *each* request? that would be bad, if we ban/purge/refresh a couple of things. could we manually close the connection from client side, after sending all requests? or somehow add the connection close header only to the last request?

otherwise i agree with the idea of doing this. interesting find, but indeed seems a good idea then.

composer.json
ddeboer: Removed this, as we’re not using this after all.

composer.json
```diff
@@ -20,15 +20,26 @@
             "homepage": "https://github.com/friendsofsymfony/FOSHttpCache/contributors"
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/ddeboer/discovery.git"
```
ddeboer: TODO: remove when https://github.com/php-http/discovery/pull/9 is merged and tagged.

src/ProxyClient/AbstractProxyClient.php
ddeboer: Agreed.

src/ProxyClient/Request/RequestQueue.php
ddeboer: I’ll leave it at `RequestQueue` for now, as it corresponds to `queueRequest()`. If we come up with something better later, we can always change it as long as we’re < 2.0.

src/ProxyClient/Request/RequestQueue.php
ddeboer: Going to try to close the connection only on the last request. But that would leave the connection open if execution fails before the last request was sent.

dbu: can we ensure it is closed from client side?

src/ProxyClient/Varnish.php
ddeboer: I’m not too happy with this myself, but we need it for Varnish to [throw the `MissingHostException()`](https://github.com/FriendsOfSymfony/FOSHttpCache/pull/173/files#diff-006816ea733cc0564d6a5b345ed47929R153). That exception is thrown when you try to invalidate a relative URL but have no base URI set. This makes no sense in Varnish, because the hash always includes the full URL, so you’ll be purging something non-existent. Does this work the same for NGINX and Symfony HttpCache? If so, the exception could be moved to `RequestQueue`. Except perhaps for the BAN method check, because BANs don’t need a full URL.

src/ProxyClient/Request/RequestQueue.php
ddeboer: Sending `Connection: Close` only for the last invalidation request makes the timing issue come back, so as far as I can see we really need to close *each* invalidation request’s connection. Doing this from the client-side won’t work either, because that requires cURL/Guzzle-specifics, and we’re trying to get rid of those. 

All in all, I guess the performance impact won’t be so big: unless your sending invalidating many URLs per second, which defeats the purpose of caching those URLs anyway.

sagikazarmark: > Doing this from the client-side won’t work either, because that requires cURL/Guzzle-specifics

Why is that?

src/ProxyClient/Varnish.php
dbu: purge is a native thing in varnish. i think in nginx and symfony http cache its a bit more "by hand" so it might be possible to invalidate without specifying the domain, i think.

then lets just move the assignement into the constructor of the varnish client.

src/ProxyClient/Request/RequestQueue.php
dbu: okay.

src/ProxyClient/Varnish.php
ddeboer: Moved to the constructor.

ddeboer:
This is now ready for final review and merge.

.travis.yml
```diff
@@ -12,11 +11,13 @@ env:
   - VARNISH_VERSION=4.0
 
 matrix:
+  allow_failures:
+    # HHVM support for CURLOPT_PROTOCOLS was added recently but not yet
```
dbu: is this a new problem? could we use some older guzzle / different psr-7 when running with hhvm?

.travis.yml
```diff
@@ -12,11 +11,13 @@ env:
...
+    # available on Travis
+    - php: hhvm
   include:
-    - php: 5.5
-      env: VARNISH_VERSION=3.0
```
dbu: do we completely drop varnish 3 support? otherwise we should keep this. maybe as a second --prefer-lowest run.

.travis.yml
```diff
@@ -36,7 +41,6 @@ before_script:
   # Install NGINX
   - sh ./tests/install-nginx.sh
   # Starting webserver
-  - sh -c "if [ '$TRAVIS_PHP_VERSION' = '5.3' ]; then ./tests/ci/install-apache.sh; fi"
```
dbu: lets also delete the bash file

CHANGELOG.md
```diff
@@ -3,11 +3,21 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+2.0.0 (unreleased)
+------------------
+
+* Decoupled from the Guzzle HTTP client.
```
dbu: we could say what we use instead.

CHANGELOG.md
```diff
@@ -3,11 +3,21 @@ Changelog
...
+* In ProxyTestCase, `getHttpClient()` has been replaced with `getHttpAdapter()`;
+  added HTTP method parameter to `getResponse()`.
+
+
 1.4.0
 -----
 
-* Added symfony/http-kernel [HttpCache client](http://foshttpcache.readthedocs.org/en/latest/proxy-clients.html#symfony-client).
-* Added [SymfonyTestCase](http://foshttpcache.readthedocs.org/en/latest/testing-your-application.html#symfonytestcase). 
+* Added symfony/http-kernel [HttpCache client](http://foshttpcache.readthedocs.org/en/stable/proxy-clients.html#symfony-client).
```
dbu: even this will at some point no longer work. can we have a 1.4 thing on readthedocs?

also, can we add some warning on the `latest` doc that this is now for the not-yet-stable 2.0 version? otherwise people with bookmarks will get confused.

doc/proxy-clients.rst
```diff
@@ -1,15 +1,71 @@
...
+
+    $ composer require guzzlehttp/psr7
+
+Alternatively:
+
+.. code-block:: bash
+
+    $ composer require zendframework/zend-diactoros
+
+.. _custom HTTP client:
```
dbu: should this anchor be called "custom HTTP adapter"?

doc/proxy-clients.rst
```diff
@@ -56,9 +112,13 @@ is available as the second parameter::
     $nginx = new Nginx($servers, 'my-cool-app.com');
 
 If you have configured NGINX to support purge requests at a separate location,
-supply that location to the class as the third parameter::
+call `setPurgeLocation()`::
+
+    use FOS\HttpCache\ProxyClient\Nginx;
+
+    $nginx = new Nginx($servers, $baseUri);
+    $nginx->setPurgeLocation(
```
dbu: i think there should be an example param and closing bracket for setPurgeLocation

src/Exception/InvalidUrlException.php
```diff
@@ -20,47 +20,30 @@ class InvalidUrlException extends InvalidArgumentException implements HttpCacheE
      * @param string $url    The invalid URL.
      * @param string $reason Further explanation why the URL was invalid (optional)
      *
-     * @return InvalidUrlException
+     * @return self
      */
     public static function invalidUrl($url, $reason = null)
     {
-        $msg = sprintf('URL "%s" is invalid.', $url);
+        $msg = sprintf('URL "%s" is invalid. ', $url);
```
dbu: The whitespace should go into the `if`, before "Reason"

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -120,196 +69,58 @@ protected function getBaseUrl()
...
-                );
-            } else {
-                // Unexpected exception type
-                $e = $exception;
+            $this->httpAdapter->sendRequests($queue->all());
+        } catch (MultiHttpAdapterException $e) {
+            $collection = new ExceptionCollection();
+            foreach ($e->getExceptions() as $exception) {
+                // A workaround for php-http currently lacking differentiation
+                // between client, server and networking errors.
```
dbu: is there a github issue we can link to?

this indeed looks rather lacking. can we also check Http status codes? or are HTTP errors not handled here at all?

src/ProxyClient/Nginx.php
```diff
@@ -109,7 +84,7 @@ private function buildPurgeUrl($url)
             $pathStartAt = strpos($url, $urlParts['path']);
             $purgeUrl = substr($url, 0, $pathStartAt).$this->purgeLocation.substr($url, $pathStartAt);
         } else {
-            $purgeUrl = $this->getBaseUrl().$this->purgeLocation.$url;
+            $purgeUrl = $this->purgeLocation.$url;
```
dbu: the base url is now in the http adapter, right? otherwise this could be a problem in some situations

src/Test/HttpClient/MockHttpAdapter.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+namespace FOS\HttpCache\Test\HttpClient;
+
+use Http\Adapter\Exception\MultiHttpAdapterException;
+use Http\Adapter\HttpAdapter;
+use GuzzleHttp\Psr7\Response;
+use Psr\Http\Message\RequestInterface;
+
+class MockHttpAdapter implements HttpAdapter
```
dbu: can you add a class docblock to explain what this is?

dbu: and does php-http not provide something like that? we might contribute it there actually - this has nothing to do with cache invalidation...

src/Test/PHPUnit/AbstractCacheConstraint.php
```diff
@@ -53,8 +53,8 @@ protected function matches($other)
                 )
             );
         }
-
-        return $this->getValue() === (string) $other->getHeader($this->header);
+
```
dbu: extra whitespace

src/Test/PHPUnit/AbstractCacheConstraint.php
```diff
@@ -62,6 +62,6 @@ protected function matches($other)
      */
     protected function failureDescription($other)
     {
-        return (string) $other . ' ' . $this->toString();
+        return 'the responses are identical';
```
dbu: would it not be "are *not* identical"? and could we use a message transformer here to give more hints what is going wrong?

src/Test/ProxyTestCase.php
```diff
@@ -64,32 +68,34 @@ public static function isCacheMiss()
...
+     * @return ResponseInterface
      */
-    public function getResponse($url, array $headers = array(), $options = array())
+    public function getResponse($uri, array $headers = [], $method = 'GET')
     {
-        return $this->getHttpClient()->get($url, $headers, $options)->send();
+        // Close connections to make sure invalidation (PURGE/BAN) requests will
+        // not interfere with content (GET) requests.
+        $headers['Connection'] = 'Close';
+        $request = $this->createRequest($method, $uri, $headers);
```
dbu: do we still need this here? i thought you changed that in the proxy clients themselves?

src/Test/ProxyTestCase.php
```diff
@@ -125,6 +131,40 @@ protected function getHostName()
...
+     * @param string $method
+     * @param string $uri
+     * @param array  $headers
+     *
+     * @return RequestInterface
+     * @throws \Exception
+     */
+    protected function createRequest($method, $uri, $headers)
+    {
+        $uri = UriFactoryDiscovery::find()->createUri($uri);
```
dbu: should we move this to a protected method so it can be customized? just the discovery call i mean.

tests/Functional/SymfonyProxyClientTest.php
```diff
@@ -69,7 +69,11 @@ public function testRefresh()
         $this->getProxyClient()->refresh('/symfony.php/cache')->flush();
         usleep(100);
         $refreshed = $this->getResponse('/symfony.php/cache');
-        $this->assertGreaterThan((float) $response->getBody(true), (float) $refreshed->getBody(true));
+
+        $originalTimestamp = (float)(string) $response->getBody();
+        $refreshedTimestamp = (float)(string) $refreshed->getBody();
+        
+        $this->assertGreaterThan($originalTimestamp, $refreshedTimestamp);
```
dbu: this is basically the same code as above for nginx, but it looks differently. could we even unify that? though not really part of this PR, but lets take note somewhere

tests/Unit/CacheInvalidatorTest.php
```diff
@@ -184,8 +184,29 @@ public function testMethodException()
...
+            ->shouldReceive('getStatusCode')->andReturn(403)
+            ->shouldReceive('getReasonPhrase')->andReturn('Forbidden')
+            ->getMock();
+        $adapterException = \Mockery::mock('\Http\Adapter\Exception\HttpAdapterException')
+            ->shouldReceive('getMessage')->andReturn('Forbidden')
+            ->shouldReceive('getRequest')->andReturn($failedRequest)
+            ->shouldReceive('getResponse')->andReturn($response)
+            ->getMock()
+        ;
+        $responseException = ProxyResponseException::proxyResponse($adapterException);
```
dbu: this did not exactly get easier... is there anything we can do to make it less complicated to mock http requests?

dbu:
read through all of it again and added quite a few comments - sorry about that. but its mostly details or things not directly related to the PR. i propose to move most of them into issues so we can merge this soon - its getting big this PR.

.travis.yml
sagikazarmark: As I can see Varnish 3 became THE lowest supported one.

src/Test/HttpClient/MockHttpAdapter.php
sagikazarmark: :+1: @ddeboer Should I create a repo for that? :wink:

tests/Unit/CacheInvalidatorTest.php
sagikazarmark: I was also thinking about how we could mock requests more easily. Maybe a library that creates requests and responses from fixtures?

For example PHP-VCR does this, but that is buried deeply in it. I have the following plan: Have a fixture generator and parser. Each objects can be created from or saved to a fixture using multiple serialization logic. For example I would like to see a message saved as it arrives (using the HTTP message format), but it could be saved do YML or JSON as well.

What do you think?

sagikazarmark: This library could be used to create mocking plugin for testing frameworks, like phpunit or phpspec.

dbu: that sounds interesting!
i guess we do not need to fix this before merging, but can replace it
once its ready. some MessageInterface to json / from json would be very
nice for general purpose. maybe a testing library together with the mock
adapter?

php-vcr is quite cool, but operating on a low level. we use it in a
project with lots of functional tests about things talking to a REST
api. but here it would be overkill. and its only for functional testing,
not in the spirit of unit testing i think.

src/Test/HttpClient/MockHttpAdapter.php
ddeboer: @sagikazarmark, please do. This was really only meant as a first pass of such an adapter, but I agree it should be in php-http.

tests/Unit/CacheInvalidatorTest.php
sagikazarmark: PHP-VCR was just an example for fixture generation. This tool should specifically be PSR-7 dependent. From some point of view this can be low level as well.

The testing library sounds great, but I think this fixture library could be used elsewhere as well. For example  @jeromegamez has an awesome Tape Recorder library. Maybe it could use this tool as well.

src/Test/HttpClient/MockHttpAdapter.php
sagikazarmark: @ddeboer Done: https://github.com/php-http/mock-adapter

.travis.yml
ddeboer: Varnish 3.0 is still included: see https://github.com/FriendsOfSymfony/FOSHttpCache/pull/173/files#diff-354f30a63fb0907d4ad57269548329e3R20.

CHANGELOG.md
ddeboer: Added 1.4.0 to the docs. If we decide to bring out more updates for 1.0, this may become cumbersome. In that case, an 1.0 branch (and building the docs for that) are preferred.

.travis.yml
ddeboer: Trying Guzzle 5 with HVVM Now.

.travis.yml
ddeboer: Done.

tests/Unit/CacheInvalidatorTest.php
adri: Hey, great thoughts here. I'm a maintainer of php-vcr.  Thanks @dbu for pinging me about this discussion.

A MessageInterface to JSON/XML would be great. I was also thinking of moving php-vcr internally over to use PSR-7, but it makes more sense to create another library which can be used everywhere. 

Php-vcr has an btw [integration with phpunit](https://github.com/php-vcr/phpunit-testlistener-vcr) (but at the moment not phpspec).

doc/proxy-clients.rst
ddeboer: Well, you don’t really create a custom HTTP adapter, but construct a client with custom options and pass that to one of php-http’s adapters.

src/Test/HttpClient/MockHttpAdapter.php
ddeboer: Docblock added. I propose to merge this PR first, then move the mock to php-http.

src/Test/PHPUnit/AbstractCacheConstraint.php
ddeboer: PHPUnit will render this as ‘failed asserting that…’.

Unlike Guzzle 3, PSR-7’s `ResponseMessage` has no `__toString()`, so we have to do it ourselves. The problem is that just dumping out the full response message (headers and body) can make the failure description quite illegible. Should we show just the headers? Showing only the cache/miss header is not very useful either: if you assert it’s a miss and it’s a hit it’s quite obvious that the header says miss. :wink:

src/Test/ProxyTestCase.php
ddeboer: You mean the `Connection: Close` header? Yes, we do.

tests/Functional/SymfonyProxyClientTest.php
ddeboer: #213.

.travis.yml
ddeboer: Okay, Guzzle 5 solves the `CURLOPT_PROTOCOLS` issue, but introduces [another one](https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/jobs/66920349). Any clue?

src/ProxyClient/AbstractProxyClient.php
ddeboer: https://github.com/php-http/adapter/issues/26.

php-http configures Guzzle so as to not throw exceptions for 4xx/5xx status codes, so the only exceptions thrown here are really timeouts or other transfer exceptions before any response was returned. Looking at the [Guzzle docs](http://guzzle.readthedocs.org/en/latest/quickstart.html#exceptions) I’d say the `$exception` will always return null for `getResponse()`.

ddeboer:
@dbu No problem. On the contrary, thanks for the thorough review. :smile:

src/ProxyClient/AbstractProxyClient.php
sagikazarmark: Actually no, that option cannot be overridden. That would allow adapters to behave differently.

I am not quite sure how exceptions should be handled. Currently no exception is thrown if a response is available.

.travis.yml
ddeboer: Probably solved by by just constructing, not mocking exceptions.

src/ProxyClient/Nginx.php
ddeboer: Indeed, it's in the adapter.

src/Test/HttpClient/MockHttpAdapter.php
dbu: agreed

src/Test/PHPUnit/AbstractCacheConstraint.php
dbu: i see. hm, there is no verbosity level we can check, right? then lets keep
like this. i was thinking about a miss because its an error response and
similar problems

src/Test/ProxyTestCase.php
dbu: ah because we do not use the request queue? should we maybe use it to be closer to reality?

src/ProxyClient/AbstractProxyClient.php
dbu: then we either have to check http status or return the response to the
client to verify it. i think exception make more sense in our Situation:
the caller is not doing http but wants something invalidated

doc/proxy-clients.rst
dbu: could we say "configure HTTP adapter" then? i guess doing this also avoids the autodiscovery, in case there are multiple adapters in the codebase and i want to force a specific one.

src/Test/ProxyTestCase.php
ddeboer: This is the test HTTP adapter (`getHttpAdapter()`), which fakes requests to the application, not the caching proxy (`getProxyClient()`) that sends invalidation requests. So I don’t think it should go through the proxy client’s RequestQueue.

src/Test/PHPUnit/AbstractCacheConstraint.php
ddeboer: No verbosity level, but good point about the error status codes: you do want to know about that. I improved the assertion message somewhat so it now says: ‘Failed asserting that response (with status code 500) is a cache hit’.

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -120,196 +70,84 @@ protected function getBaseUrl()
...
-                    $exception
-                );
-            } else {
-                // Unexpected exception type
-                $e = $exception;
+        foreach ($responses as $response) {
+            if ($response->getStatusCode() >= 400
+                && $response->getStatusCode() < 600
+            ) {
+                $exceptions[] = ProxyResponseException::proxyResponse($response);
```
ddeboer: Now throwing exceptions again for responses with error status codes. Unfortunately, I see no easy way to get the corresponding request. Although Guzzle should return responses in the same order as the requests, things get difficult when the responses are mixed with exceptions (for timeout errors).

src/Test/ProxyTestCase.php
dbu: okay.

src/Test/PHPUnit/AbstractCacheConstraint.php
```diff
@@ -62,6 +62,10 @@ protected function matches($other)
      */
     protected function failureDescription($other)
     {
-        return (string) $other . ' ' . $this->toString();
+        return sprintf(
+            'response (with status code %s) %s',
+            $other->getStatusCode(),
+            $this->toString()
```
dbu: did you do this on purpose? it was "are identical"...

src/ProxyClient/AbstractProxyClient.php
dbu: well thats already a lot better.

dbu:
there was a test failure.

apart from that i am fine with squash and merge if you want. won't be much online the next days so go ahead if you want.

src/Test/PHPUnit/AbstractCacheConstraint.php
dbu: i think this is the last open question. anything else we need to discuss before merging?

ddeboer: This is to improve the failure messages: `Failed asserting that response (with status code 400) is a hit`.

vidarl:
I realize this PR was merged quite some time ago but, I have a question

> i propose we add the sleep (lets see if 100ms is also reliably enough) and open an issue at the varnish project to ask them if that is expected behaviour.

Did you open an issue on varnish project regarding this?

I experienced the same problem with v1.4 and came up with the following workaround in varnish.vcl instead of using a `sleep` in my tests:
```
sub vcl_deliver {
    (...)
    // This is the foshttpcache way of doing it, but it is not bulletproof.
    // Seems like for about 50000 usec after a purge, X-Varnish may also contain br_req ID and we'll falsely have a HIT flag
//      if (resp.http.X-Varnish ~ " ") {
//          set resp.http.X-Cache = "HIT";
//      } else {
//          set resp.http.X-Cache = "MISS";
//      }

    if (obj.hits > 0) {
      set resp.http.X-Cache = "HIT";
    } else {
      set resp.http.X-Cache = "MISS";
    }
```

I anyway don' know why you don't use obj.hits in order to detect a HIT or not as virtually everybody else do that, AFAIK

dbu:
the obj.hits worked well with varnish 3. varnish 4 seems not to set obj.hits reliable anymore, see for example: https://coderwall.com/p/zksgmq/vanish-4-effectively-detect-hit-miss-in-vcl

if this was ammended in later versions of varnish 4, do you have some source about that? if we can avoid the sleep in favor of something more reliable, we would prefer that as well.

vidarl:
I am using varnish 4.0.3 and have never experienced any issue with obj.hits.
Might be different story with 4.1.3 then?

ddeboer:
The change in `obj.hits` was introduced with Varnish 4.0’s [split between frontend and backend threads](https://www.varnish-cache.org/docs/4.0/whats-new/changes.html#client-backend-split). This [may have been fixed](https://github.com/varnishcache/varnish-cache/commit/01b4e4a4d3a674f76fc0c38c1dd8e7485999cff6) in 4.0.3 however.

That would allow us to revert the changes we made in our VCLs (and docs) for 4.0. @vidarl Do you have time to experiment with changing the VCLs and running the FOSHttpCache tests to see if you get the same results for hits/misses?

We

PR #174, opened 2015-02-26
do not expose cache tags to the internet

dbu:
we missed this, leading to showing cache tags to the internet. this is no security risk, but not elegant.


ddeboer:
Good one, thanks!

PR #178, opened 2015-03-13
need a way to access the TagHandler to refactor CacheManager in a BC way

dbu:
the only alternative i can see is duplicating the TagHandler property in the CacheManager, but that feels horrible... still maybe that is the way to go? see also https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/182


dbu:
in the end i don't need this

PR #181, opened 2015-04-01
Fixed linebreak

Nyholm:


doc/symfony-cache-configuration.rst
```diff
@@ -168,7 +168,8 @@ options through the constructor:
 
     With Apache, you can do this for example in a ``.htaccess`` file:
```
ddeboer: I guess this should end in `::` instead to make the snippet below a proper code sample.

ddeboer:
Thanks! While you’re at it, could you also change the colon line ending to make this into a proper code snippet?

Nyholm:
Of course. I was not familiar with that syntax. 
I've updated the file.

dbu:
thanks @Nyholm - are you giving the FOSHttpCache a try or already using it?

Nyholm:
Thank you for merging. 
I'm giving it a try. I was stuck for some time due a [bug in SensioFrameworkExtraBundle](https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/349) but apart from that the integration was easy.

dbu:
cool, glad you like it. the framework extra bundle thing sounds like a
problem indeed. might even be a security issue when using the user context.

PR #193, opened 2015-05-08
adding client for the symfony built-in reverse proxy HttpCache

### LINKED ISSUE #192: Symfony HttpCache invalidation client
See https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/209#issuecomment-100185710. The client probably shouldn’t extend AbstractProxyClient, as it won’t be a HTTP client.

dbu:
fix #192


doc/proxy-clients.rst
```diff
@@ -35,16 +35,16 @@ include that port in the base URL::
...
-    To use the client, you need to :doc:`configure Varnish <varnish-configuration>` accordingly.
+    To make invalidation work, you need to :doc:`configure Varnish <varnish-configuration>` accordingly.
 
 NGINX Client
 ~~~~~~~~~~~~
 
 At minimum, supply an array containing IPs or hostnames of the NGINX servers
 that you want to send invalidation requests to. Make sure to include the port
-NGINX runs on if it is not port 80::
+NGINX runs on if it is not the default::
```
dbu: default depends on http vs https. for varnish there is only http.

tests/Unit/ProxyClient/AbstractProxyClientTest.php
```diff
@@ -0,0 +1,282 @@
...
+use Guzzle\Http\Exception\CurlException;
+use Guzzle\Http\Exception\MultiTransferException;
+use Guzzle\Http\Exception\RequestException;
+use Guzzle\Plugin\Mock\MockPlugin;
+use Guzzle\Http\Message\Response;
+use Guzzle\Http\Message\Request;
+use \Mockery;
+
+/**
+ * Testing the base methods of the proxy client, using the Varnish client as concrete class.
```
dbu: this is cheating a little bit, but i felt its overkill to do a test client just for this.

tests/Unit/ProxyClient/SymfonyTest.php
```diff
@@ -0,0 +1,112 @@
...
+class SymfonyTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var MockPlugin
+     */
+    private $mock;
+
+    private $client;
+
+    public function testPurge()
```
dbu: this is pure copy-paste. not sure how we can avoid it however.

dbu:
i guess we really should do some functional tests around this whole stuff :-/

composer.json
```diff
@@ -31,7 +31,8 @@
         "mockery/mockery": "*",
         "monolog/monolog": "*",
         "symfony/process": "~2.3",
-        "symfony/http-kernel": "~2.3"
+        "symfony/http-kernel": "~2.3",
+        "symfony/config": "~2.3"
```
dbu: this is a hidden dependency of the HttpKernel class

dbu:
there we go, added functional tests. with the webserver already running, this actually looks quite simple.

src/Test/Proxy/SymfonyProxy.php
```diff
@@ -0,0 +1,59 @@
...
+/*
+ * This file is part of the FOSHttpCache package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\Test\Proxy;
```
ddeboer: Newline missing.

src/Test/Proxy/SymfonyProxy.php
```diff
@@ -0,0 +1,59 @@
...
+use Symfony\Component\Filesystem\Filesystem;
```
ddeboer: Tests fail because of a missing dependency. I guess we should add it to our `composer.json`?

src/Test/SymfonyTestCase.php
```diff
@@ -0,0 +1,95 @@
...
+ */
+abstract class SymfonyTestCase extends ProxyTestCase
+{
+    /**
+     * @var Symfony
+     */
+    protected $proxyClient;
+
+    /**
+     * @var
```
ddeboer: Missing type.

tests/Unit/ProxyClient/AbstractProxyClientTest.php
ddeboer: Alternatively, we could have an abstract `createClient` factory method that the VarnishTest and SymfonyTest implement to return the proper proxy client.

ddeboer:
Cool! Let’s get the build running and merge this. :)

src/Test/Proxy/SymfonyProxy.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Clear all cached content from the proxy server
+     */
+    public function clear()
+    {
+        if (is_dir($this->getCacheDir())) {
+            $fs = new Filesystem();
+            $fs->remove($this->getCacheDir());
```
dbu: ah, i had that around because i accidentally required symfony/config which depends on the filesystem. this is the only place we use it. not sure if its worth it to have that. we could also do `system('rm -R '.$this->getCacheDir());` (and `DEL /S` for windows).

one thing i just worry here is that on misconfiguration, we could accidentally attempt to delete the whole filesystem or the home folder or cwd - this is not related to how we delete. should we sanity check the cache dir? as in: must not be empty or "/"?

src/Test/Proxy/SymfonyProxy.php
dbu: fixed

tests/Unit/ProxyClient/AbstractProxyClientTest.php
dbu: yeah. but its rather complicated and duplicating the tests that are run. tests don't take that long so it would not hurt. want to do it? otherwise i propose we leave as is for now.

ddeboer: Let’s leave it for now. I have some plans for moving test logic into traits for 2.0, so I’ll tackle this then, too.

ddeboer:
Hm, weird hhvm failure. Any idea what’s going on?

dbu:
i spent quite some time investigating. i created a branch where i log a lot of things. 

```
curl -v http://localhost:8080/cache.php
- works -
curl -v http://localhost:8080/symfony.php/cache
- connection killed after headers -
127.0.0.1:8080 ::1 - - [29/May/2015:13:32:12 +0000] "GET /cache.php HTTP/1.1" 200 276 "-" "curl/7.22.0 (x86_64-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 librtmp/2.3"
127.0.0.1:8080 ::1 - - [29/May/2015:13:32:12 +0000] "GET /symfony.php/cache HTTP/1.1" 200 385 "-" "curl/7.22.0 (x86_64-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 
```

locally, i had a fatal error about the event dispatcher being null during kernel->terminate(). on the server, i do not have that, but i see `Log file not specified under daemon mode.\n\n` - do we maybe launch hhvm in a way that makes it not log anything?

i have to give up now, if you have any ideas please continue in #200 (my idea there is to just figure out why things fail, then make the solution here and discard the other branch)

dbu:
argh. shall we just skip those tests when running under hhvm so that we can move on? i am pretty much clueless but feel the problem really is related to our hhvm setup on travis.

ddeboer:
Perhaps allow_failures for hhvm?

dbu:
> Perhaps allow_failures for hhvm?

I would prefer skip in the client test so gravis gets still red when there
are other issues

ddeboer:
Fine with me!

dbu:
okay! this should now get green i hope.

can you do a last review @ddeboer ?

i will then take stuff from that debug branch and do a PR to remove the skipping so we can periodically check if the problem continues to exist.

tests/Functional/Fixtures/Symfony/AppCache.php
```diff
@@ -0,0 +1,28 @@
...
+    {
+        $response = parent::handle($request, $type, $catch);
+
+        if ($response->headers->has('X-Symfony-Cache')) {
+            if (false !== strpos($response->headers->get('X-Symfony-Cache'), 'miss')) {
+                $state = 'MISS';
+            } elseif (false !== strpos($response->headers->get('X-Symfony-Cache'), 'fresh')) {
+                $state = 'HIT';
+            } else {
+                $state = 'UNDETERMINED';
```
ddeboer: This is a bit weird: one would expect either a hit or a miss. Is this state ever reached?

dbu: well, the response could miss that header, or it could contain neither miss nor fresh. i preferred this over defaulting to MISS or HIT in that case so that we see that there is a problem.

ddeboer: Fair enough!

PR #194, opened 2015-05-08
adding more builds to the matrix

dbu:


.travis.yml
```diff
@@ -6,14 +6,17 @@ php:
...
-  - VARNISH_VERSION=3.0
+  - VARNISH_VERSION=4.0
 
 matrix:
   include:
     - php: 5.5
-      env: VARNISH_VERSION=4.0
+      env: VARNISH_VERSION=3.0
+    - php: 5.3
+      env: SYMFONY_VERSION=2.3.* COMPOSER_FLAGS="--prefer-lowest"
```
ddeboer: It seems `env` overrides the global `env` completely, so we need another `VARNISH_VERSION=4.0` (or `3.0`) here.

dbu: ah thats the problem! i was tinking the php 5.3.3 box had an issue - it has issues with not supporting ssl in php, so i switched to php 5.3 box... but i think you are right.

ddeboer:
Okay, Varnish now installs correctly, but unfortunately tests have started failing. Do you have any clue? Is it one of our vendors, of which we include lower versions here?

dbu:
somehow our webserver setup did not work well with the php 5.3.3 box. i go with the php 5.3 box now, that is a lot better than nothing.

PR #195, opened 2015-05-09
Some small doc improvements

ddeboer:


doc/conf.py
```diff
@@ -110,10 +110,11 @@
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
-#html_title = None
+# Don't show version in title
+html_title = "FOSHttpCache Documentation"
```
ddeboer: This hides the version from the title when viewing the docs. Currently foshttpcache.readthedocs.org is listed as ‘FOSHttpCache 1.0.0 documentation’ which may be somewhat confusing as we’re already at 1.3.0.

doc/conf.py
```diff
@@ -110,10 +110,11 @@
...
 
 # A shorter title for the navigation bar.  Default is the same as html_title.
-#html_short_title = None
+html_short_title = FOSHttpCache
```
dbu: seems the example was misleading. this needs quotes i think. travis doc building fails.

doc/conf.py
dbu: agreed. we could make it "FOSHttpCache Documentation (current)" though,
to make it more obvious, if you want.

ddeboer:
Added the quotes, this should build now.

doc/conf.py
ddeboer: As we’re only maintaining one major version now, I left that out for the time being. When we start having docs for version 2 as well as 1, we can always add it in.

dbu:
:+1:

PR #196, opened 2015-05-15
Cleanup setup

dbu:
while working on the symfony client, i realized that we are rather implicit in how the symfony cache works.

however, this is a BC break for people that implemented their own HttpCache rather than extending ours. not sure if this can go into 1.x or needs to wait for 2.0 - but even with the trait, we still want this, as there are no typehints on traits.


src/SymfonyCache/CacheEvent.php
```diff
@@ -39,10 +38,10 @@ class CacheEvent extends Event
     private $response;
 
     /**
-     * @param HttpCache $kernel  The kernel raising with this event.
-     * @param Request   $request The request being processed.
+     * @param CacheManipulationInterface $kernel  The kernel raising with this event.
+     * @param Request                    $request The request being processed.
      */
-    public function __construct(HttpCache $kernel, Request $request)
+    public function __construct(CacheManipulationInterface $kernel, Request $request)
```
dbu: this is the actual BC break. should we just additionally accept a HttpCache without interface until 2.0?

ddeboer: Yes, as it’s a BC break I would prefer to postpone this to 2.0. Should we leave this PR open till then?

dbu: i could remove the typehint and do an instanceof check in the constructor for either HttpCache *or* the interface.

ddeboer: :+1:

And maybe comment it to be something that can be taken out for 2.0?

dbu:
ok, adjusted to be BC and also updated the documentation. if you have no better idea for the name of the CacheManipulationInterface, this would be mergeable now.

src/SymfonyCache/CacheManipulationInterface.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\HttpCache\StoreInterface;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+/**
+ * Interface for extra requirements to do refresh operations with the cache.
+ */
+interface CacheManipulationInterface extends HttpKernelInterface
```
ddeboer: What about `CacheInvalidationInterface` or `InvalidationCacheInterface`?

dbu:
good idea, renamed the interface to CacheInvalidationInterface.

ddeboer:
Cool, thanks!

PR #197, opened 2015-05-23
Exclude files and dirs when building packages

ddeboer:
This will exclude some files and dirs when creating releases on GitHub. Smaller packages means faster downloads through Composer (using `--prefer-dist``) for our users. :wink:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+.gitattributes export-ignore
+.gitignore export-ignore
+.scrutinizer.yml export-ignore
+.travis.yml export-ignore
+phpunit.xml.dist export-ignore
+tests/ export-ignore


.gitattributes
```diff
@@ -0,0 +1,6 @@
+.gitattributes export-ignore
+.gitignore export-ignore
+.scrutinizer.yml
```
dbu: do we not need to say export-ignore here too?

dbu:
oh right, good idea! i guess one could also ignore the documentation folder - rst source code is not that helpful anyways.

.gitattributes
ddeboer: Whoops yes.

ddeboer:
Yeah, I wasn’t sure about the docs. I like to have them downloaded in case I end up coding somewhere without a proper internet connection (in the train, for instance). Although raw rst is not very legible, it’s better than nothing.

dbu:
docs could be a considerable part of the whole memory footprint, and there is always prefer-source for people who want everything included. but lets start with this. do you also copy this into the bundle? no need for a PR i guess, just push to master sounds fine to me.

PR #201, opened 2015-05-31
fix test setup with symfony backend

dbu:
follow up of #193

### DIFF
--- a/tests/Functional/Fixtures/Symfony/AppKernel.php
+++ b/tests/Functional/Fixtures/Symfony/AppKernel.php
@@ -4,13 +4,18 @@ namespace FOS\HttpCache\Tests\Functional\Fixtures\Symfony;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\HttpKernel;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
-class AppKernel extends HttpKernel
+/**
+ * A simplistic kernel that is actually the whole application.
+ */
+class AppKernel implements HttpKernelInterface
 {
-    public function __construct() {}
-
+    /**
+     * This simplistic kernel handles the request immediately inline.
+     *
+     * {@inheritDoc}
+     */
     public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQUEST, $catch = true)
     {
         switch ($request->getPathInfo()) {
--- a/tests/Functional/SymfonyProxyClientTest.php
+++ b/tests/Functional/SymfonyProxyClientTest.php
@@ -20,20 +20,6 @@ use FOS\HttpCache\Test\SymfonyTestCase;
  */
 class SymfonyProxyClientTest extends SymfonyTestCase
 {
-    public function setUp()
-    {
-        /*
-         * We did not figure out how to run this test with hhvm. Help welcome!
-         * On travis, connection is closed after the headers are sent without providing
-         * anything in either the hhvm or apache log.
-         */
-        if (defined('HHVM_VERSION')) {
-            $this->markTestSkipped('Test not working with hhvm as backend server');
-        }
-
-        parent::setUp();
-    }
-
     public function testPurge()
     {
         $this->assertMiss($this->getResponse('/symfony.php/cache'));


ddeboer:
Nice, you’re going to solve this one? :+1:

.travis.yml
```diff
@@ -1,23 +1,11 @@
...
 env:
   - VARNISH_VERSION=4.0
 
-matrix:
-  include:
-    - php: 5.5
-      env: VARNISH_VERSION=3.0
-    - php: 5.3
-      env: SYMFONY_VERSION=2.3.* VARNISH_VERSION=4.0 COMPOSER_FLAGS="--prefer-lowest"
-
```
dbu: changes in this file need to be reverted, just to have faster cycles with travis.

dbu:
not sure. i just wanted to get started on it. if the next run does not provide helpful information i will leave it be for now.

i now have a hhvm setup with apache 2.2 and it ends up in 
```
Fatal error: Uncaught exception 'BadMethodCallException' with message 'Call to a member function dispatch() on a non-object (NULL)' in /vagrant_data/FOSHttpCache/vendor/symfony/http-kernel/Symfony/Component/HttpKernel/HttpKernel.php:86
Stack trace:
#0 /vagrant_data/FOSHttpCache/vendor/symfony/http-kernel/Symfony/Component/HttpKernel/HttpCache/HttpCache.php(245): Symfony\\Component\\HttpKernel\\HttpKernel->terminate()
#1 /vagrant_data/FOSHttpCache/tests/Functional/Fixtures/web/symfony.php(24): Symfony\\Component\\HttpKernel\\HttpCache\\HttpCache->terminate()
#2 {main}
```

tests/Functional/Fixtures/web/symfony.php
```diff
@@ -21,4 +21,4 @@
 $request = Request::createFromGlobals();
 $response = $kernel->handle($request);
 $response->send();
-$kernel->terminate($request, $response);
+//$kernel->terminate($request, $response);
```
dbu: hm, so indeed at the time we try to terminate the kernel, the dispatcher inside the kernel is set to null. i have no idea why or even how this can happen.

it would sound like a general symfony + hhvm problem, but google does not bring up the problem.

```
Fatal error: Uncaught exception 'BadMethodCallException' with message 'Call to a member function dispatch() on a non-object (NULL)' in /vagrant_data/FOSHttpCache/vendor/symfony/http-kernel/Symfony/Component/HttpKernel/HttpKernel.php:86
Stack trace:
#0 /vagrant_data/FOSHttpCache/vendor/symfony/http-kernel/Symfony/Component/HttpKernel/HttpCache/HttpCache.php(245): Symfony\\Component\\HttpKernel\\HttpKernel->terminate()
#1 /vagrant_data/FOSHttpCache/tests/Functional/Fixtures/web/symfony.php(24): Symfony\\Component\\HttpKernel\\HttpCache\\HttpCache->terminate()
#2 {main}
```

ddeboer: Very weird. But this is the last problem, right? Otherwise it's green.

dbu: i think this is the only problem. i commented this line as i did not manage to get hhvm to log anything on travis. locally, it logged what i pasted in my comment. and i really can't explain why that could ever happen. i checked the codebase, nothing is assigning to $this->dispatcher, also not the extending classes...

xabbuh: I'm curious if changing the code as suggested in symfony/symfony-standard#658 does the trick.

dbu: it makes the tests green, but only because the response is sent before hhvm crashes... maybe we do not need to bother for FOSHttpCache, if we do not rely on any kernel terminate events.

dbu: oh hell, i dug some more. the dispatcher is always null. in symfony, you create a Kernel which gets HttpKernel from the container. the event dispatcher is created and injected by the container... i need to fix the setup of the kernel then.

tests/Functional/Fixtures/Symfony/AppKernel.php
```diff
@@ -4,13 +4,18 @@
...
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\HttpKernel;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
-class AppKernel extends HttpKernel
+/**
+ * A simplistic kernel that is actually the whole application.
+ */
+class AppKernel implements HttpKernelInterface
```
dbu: so actually extending the HttpKernel was only creating problems. the simplistic kernel just needs a handle method!

dbu:
this hopefully will get green now.

i think the problem exists on all php versions, but php still sends the response while hhvm immediately aborts on such a fatal error and does not flush the response body, which made us see the mistake. having terminate made no sense, and was not supposed to work (i had an empty constructor to not call the parent constructor which would expect the event dispatcher... so it was all my fault and i got confused)

ddeboer:
Nice one! Glad this is solved. :smile:

PR #212, opened 2015-06-10
Set Scrutinizer external coverage timeout

ddeboer:
See https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/216.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -2,4 +2,5 @@ imports:
   - php
 
 tools:
-  external_code_coverage: true
+  external_code_coverage:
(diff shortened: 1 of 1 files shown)


dbu:
:+1:

stof:
My comment on https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/221 would apply here too

dbu:
ups, saw that comment too late. but it seems to work now.

ddeboer:
@stof Not quite, as here we still run code coverage on all builds. We should change that, though.

PR #214, opened 2015-06-16
Symfony HttpCache listener for custom ttl header

### LINKED ISSUE #151: Symfony HttpCache subscriber for custom proxy ttl
have a look at https://github.com/sulu-cmf/sulu/blob/develop/src/Sulu/Component/HttpCache/HttpCache.php and extract that into a subscriber, with the constant for the header as default value.

add new events to the HttpCache as needed in order to be able to do this.

dbu:
fix #151

i guess its ok if this only goes to version 2, its not urgent but for completeness.

TODO
* [x] Document
* [x] Functional test

i opt for including this here rather than the bundle, for consistency of where we have http cache listeners. we could also move the vcl and documentation for the varnish custom ttl here - that would also make it easy to test that vcl as we already have the whole setup here.

### DIFF
--- a/doc/testing-your-application.rst
+++ b/doc/testing-your-application.rst
@@ -98,8 +98,10 @@ Make sure ``symfony/process`` is available in your project:
 
     $ composer require symfony/process
 
-Then set your Varnish configuration (VCL) file. All available configuration
-parameters are shown below.
+Then set your Varnish configuration (VCL) file. Configuration is handled either
+by overwriting the getter or by defining a PHP constant. You can set the
+constants in your ``phpunit.xml`` or in the bootstrap file. Available
+configuration parameters are:
 
 ======================= ========================= ================================================== ===========================================
 Constant                Getter                    Default                                            Description
@@ -109,9 +111,21 @@ Constant                Getter                    Default                       
 ``VARNISH_PORT``        ``getCachingProxyPort()`` ``6181``                                           port Varnish listens on
 ``VARNISH_MGMT_PORT``   ``getVarnishMgmtPort()``  ``6182``                                           Varnish management port
 ``VARNISH_CACHE_DIR``   ``getCacheDir()``         ``sys_get_temp_dir()`` + ``/foshttpcache-varnish`` directory to use for cache
-``VARNISH_VERSION``     ``getVarnishVersion()``   ``4``                                              installed varnish application version
 ``WEB_SERVER_HOSTNAME`` ``getHostName()``                                                            hostname your application can be reached at
 ======================= ========================= ================================================== ===========================================
+
+The Varnish version is controlled by an environment variable (in case you want
+to test both Varnish 3 and 4 on a continuous integration system). See the
+``.travis.yml`` of the FOSHttpCache git repository for an example.
+
+==================== ========================= ======= ===========================================
+Environment Variable Getter                    Default Description
+==================== ========================= ======= ===========================================
+``VARNISH_VERSION``  ``getVarnishVersion()``   ``4``   version of varnish application that is used
+==================== ========================= ======= ===========================================
+
+See ``tests/bootstrap.php`` for an example how this repository uses the version
+information to set the right ``VARNISH_FILE`` constant.
 
 Enable Assertions
 '''''''''''''''''
--- a/doc/varnish-configuration.rst
+++ b/doc/varnish-configuration.rst
@@ -10,11 +10,11 @@ and 4.
 Basic Varnish Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To invalidate cached objects in Varnish, begin by adding an
-`ACL <https://www.varnish-cache.org/docs/3.0/tutorial/vcl.html#example-3-acls>`_
-to your Varnish configuration. This ACL determines which IPs are allowed to
-issue invalidation requests. Let’s call the ACL `invalidators`. The ACL below
-will be used throughout the Varnish examples on this page.
+To invalidate cached objects in Varnish, begin by adding an `ACL`_ (for Varnish
+3 see `ACL for Varnish 3`_) to your Varnish configuration. This ACL determines
+which IPs are allowed to issue invalidation requests. To use the provided
+configuration fragments, this ACL has to be named ``invalidators``. The most
+simple ACL, valid for both Varnish 4 and Varnish 3, looks as follows:
 
 .. code-block:: varnish4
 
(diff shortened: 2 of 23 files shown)


ddeboer:
:+1:

Agreed on having it in the library: everything which is not strictly (full-stack) Symfony should go here.

dbu:
cool. can you please review the implementation - would prefer to fix implementation before writing documentation and more tests ;-)
@dantleech @chirimoya can you please also have a look? i chose a different approach than you did in sulu, but would hope that once this is tagged stable you can switch to this over your implementation. if there are good reasons to prefer your implementation, please tell them - we can still change things here.

src/SymfonyCache/ReverseProxyTtlListener.php
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @author David Buchmann <mail@davidbu.ch>
+ *
+ * {@inheritdoc}
+ */
+class ReverseProxyTtlListener implements EventSubscriberInterface
+{
+    private $ttlHeader;
+
+    const SMAXAGE_BACKUP = 'FOS-Smaxage-Backup';
```
dantleech: Should this also be configurable? Not important I guess.

chirimoya:
+1

src/SymfonyCache/ReverseProxyTtlListener.php
dbu: > Should this also be configurable? Not important I guess.

its only used inside this class so i think there is no need

src/SymfonyCache/ReverseProxyTtlListener.php
```diff
@@ -0,0 +1,89 @@
...
+
+    public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL')
```
ddeboer: Please docblock explaining the `$ttlHeader` param.

src/SymfonyCache/ReverseProxyTtlListener.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    public function useCustomTtl(CacheEvent $e)
+    {
+        $response = $e->getResponse();
+        if (!$response->headers->has($this->ttlHeader)) {
+            return;
+        }
+        $backup = $response->headers->hasCacheControlDirective('s-maxage')
+            ? $response->headers->getCacheControlDirective('s-maxage')
+            : 'false'
```
ddeboer: Wouldn’t just `getCacheControlDirective()` ?: false` work, too?

src/SymfonyCache/ReverseProxyTtlListener.php
```diff
@@ -0,0 +1,89 @@
...
+            $smaxage = $response->headers->get(static::SMAXAGE_BACKUP);
+            if ('false' === $smaxage) {
+                $response->headers->removeCacheControlDirective('s-maxage');
+            } else {
+                $response->headers->addCacheControlDirective('s-maxage', $smaxage);
+            }
+        }
+        $response->headers->remove($this->ttlHeader);
+        $response->headers->remove(static::SMAXAGE_BACKUP);
+
```
ddeboer: empty line may be removed

ddeboer:
Let’s pull this into 2.0. So needs a rebase. :wink:

src/SymfonyCache/ReverseProxyTtlListener.php
dbu: no, because then we would lose the `s-maxage: 0` case and consider that as s-maxage not set

ddeboer:
Please add docs and a test.

dbu:
ok, added documentation and a test. also some random cleanup in the doc. notably i started to write the calls to our subs into the corresponding doc section, rather than the extremly flaky includes.

doc/testing-your-application.rst
```diff
@@ -109,10 +111,22 @@ Constant                Getter                    Default
 ``VARNISH_PORT``        ``getCachingProxyPort()`` ``6181``                                           port Varnish listens on
 ``VARNISH_MGMT_PORT``   ``getVarnishMgmtPort()``  ``6182``                                           Varnish management port
 ``VARNISH_CACHE_DIR``   ``getCacheDir()``         ``sys_get_temp_dir()`` + ``/foshttpcache-varnish`` directory to use for cache
-``VARNISH_VERSION``     ``getVarnishVersion()``   ``4``                                              installed varnish application version
```
dbu: this was rather puzzling me with my varnish 3 locally. i think it does make sense to use an environment over a constant here.

tests/bootstrap.php
```diff
@@ -16,7 +16,7 @@
 
 if (!defined('VARNISH_FILE')) {
     if (getenv('VARNISH_VERSION')
-        && (0 === strncmp('3.', getenv('VARNISH_VERSION'), 2))
+        && (0 === strncmp('3', getenv('VARNISH_VERSION'), 1))
```
dbu: imo this was too restrictive.

doc/testing-your-application.rst
ddeboer: We need the constant as well because the lib’s tests run against both Varnish versions.

doc/varnish-configuration.rst
```diff
@@ -10,11 +10,11 @@ and 4.
 Basic Varnish Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To invalidate cached objects in Varnish, begin by adding an
-`ACL <https://www.varnish-cache.org/docs/3.0/tutorial/vcl.html#example-3-acls>`_
-to your Varnish configuration. This ACL determines which IPs are allowed to
-issue invalidation requests. Let’s call the ACL `invalidators`. The ACL below
-will be used throughout the Varnish examples on this page.
+To invalidate cached objects in Varnish, begin by adding an `ACL` (for Varnish
+3 see `ACL for Varnish 3`_) to your Varnish configuration. This ACL determines
```
ddeboer: We should have a Varnish 3 tab below with the ACL config.

doc/varnish-configuration.rst
```diff
@@ -36,111 +36,162 @@ will be used throughout the Varnish examples on this page.
 Provided Vcl Subroutines
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-In order to ease configuration we provide a set of vcl subroutines in the resources/config directory.
-These can be directly included into ``your_varnish.vcl`` and the needed subroutines called
-from the respective vcl_* subroutines.
+In order to ease configuration we provide a set of vcl subroutines in the
```
ddeboer: `vcl` -> `VCL`

doc/varnish-configuration.rst
```diff
@@ -36,111 +36,162 @@ will be used throughout the Varnish examples on this page.
...
-.. important::
-    When including one of the provided vcl you need to call all the defined subroutines
-    or your configuration will not be valid.
+.. tip::
+    When including one of the provided vcl, you need to call all the defined
+    subroutines or your configuration will not be valid.
 
-    See the respective sections below on how to configure usage of each of the provided vcl's.
+    See the respective sections below on how to configure usage of each of the
+    provided vcl's.
```
ddeboer: `VCLs` (these mistakes were here before, but I just noticed them)

doc/varnish-configuration.rst
```diff
@@ -36,111 +36,162 @@ will be used throughout the Varnish examples on this page.
...
-        :language: varnish4
-        :lines: 17,19,21-22
-    .. literalinclude:: ../tests/Functional/Fixtures/varnish-3/fos.vcl
-        :language: varnish3
-        :lines: 15,17,19-20,25-32
+        include "path-to-config/varnish-4/fos_purge.vcl";
+
+        sub vcl_recv {
+            call fos_purge_recv;
+        }
```
ddeboer: :+1: on writing out VCL instead of the includes. Let’s keep the includes for larger VCL sections only.

doc/varnish-configuration.rst
```diff
@@ -36,111 +36,162 @@ will be used throughout the Varnish examples on this page.
...
+        sub vcl_fetch {
+            call fos_ban_fetch;
+        }
+
+        sub vcl_deliver {
+            call fos_ban_deliver;
+        }
+
+Varnish contains a `ban lurker`_ that crawls the content to eventually throw
+out banned data even when it’s not requested by any client. Read more on
```
ddeboer: The reference to the ban lurker is a bit out of the blue, here. Perhaps we should add that the FOSHttpCache VCL subroutines set ban lurker-friendly headers (in other words, is compatible with Varnish's ban lurker)?

doc/varnish-configuration.rst
```diff
@@ -36,111 +36,162 @@ will be used throughout the Varnish examples on this page.
...
+`handling BAN requests`_ in the Varnish documentation (for Varnish 3, see
+`banning for Varnish 3`_).
 
 .. _varnish_tagging:
 
 Tagging
 ~~~~~~~
 
-If you have included fos_ban.vcl, tagging will be automatically enabled using a ``X-Cache-Tags`` header :ref:`cache tagging <tags>`.
+Feature: :ref:`cache tagging <tags>`
```
ddeboer: :+1: on references to the feature sections.

tests/bootstrap.php
ddeboer: :+1:

doc/testing-your-application.rst
dbu: i am not removing anything except wrong documentation. i added documentation for the environment variable and keep the getter that is using the env variable. we use the env variable both in VarnishTestCase and tests/bootstrap.php

we could alternatively look at both or have tests/bootstrap.php copy the env variable into a constant. but that all feels a bit clumsy and just as confusing.

doc/varnish-configuration.rst
dbu: there is no difference actually. i just wanted to link to the proper version of the documentation. not sure how we should handle that. duplicating the doc config with no difference seems a bit strange. should i just say that this did not change?

doc/varnish-configuration.rst
dbu: yeah, i normalized to lowercase... fixed to uppercase now, preferring that as well.

doc/varnish-configuration.rst
dbu: agreed, when there is actual logic in the VCL, we should keep including them and not copy. but here it felt tedious and there where slight mistakes already (calling debug, missing paranthesis)

doc/varnish-configuration.rst
dbu: good idea.

dbu:
updated

resources/config/varnish-4/fos_custom_ttl.vcl
```diff
@@ -0,0 +1,36 @@
...
+         *
+         * As of Varnish 4.0, inline C is disabled by default. To use this
+         * feature, you need to add `-p vcc_allow_inline_c=on` to your Varnish
+         * startup command.
+         */
+        C{
+            const struct gethdr_s hdr = { HDR_BERESP, "\024X-Reverse-Proxy-TTL:" };
+            ttl = VRT_GetHdr(ctx, &hdr);
+            VRT_l_beresp_ttl(ctx, atoi(ttl));
+        }C
```
dbu: @ddeboer do you see what is wrong here? i have the include above, and the code as you did in the playground. but tests are still not happy. any ideas? can i not include from an included file?

ddeboer: You’re missing `char *ttl;`

dbu: oh well. seems i got confused by reading the diff and then just not thinking. thanks!

doc/varnish-configuration.rst
dbu: this question is still open. @ddeboer want me to do anything more here?

dbu:
if this is green, i want to squash before we merge.

final review anybody?

ddeboer:
Looks good to me. Go ahead and squash. :smile:

doc/varnish-configuration.rst
dbu: okay, updated. its actually no longer just an example...

dbu:
squashed and some last cleanup. please check and merge if you are happy.

ddeboer:
Thanks!

Issue #213, opened 2015-06-15
Reduce code duplication in tests


ddeboer:
https://github.com/FriendsOfSymfony/FOSHttpCache/pull/173/files#r32400969

@dbu I think we have already talked about this elsewhere, when you created the `AbstractProxyClientTest`.

dbu:
see #320

PR #218, opened 2015-06-20
Provide test functionality in traits

### LINKED ISSUE #202: Restructure abstract test class into trait(s)
So users can have their own abstract test class but still use our Varnish/Nginx bootstrapping and custom assertions.

ddeboer:
Fix #202.

**Advantage**: users no longer have to extend our test classes, so they are free to extend their own (or their framework’s). Optionally, they can continue to extend `SymfonyTestCase`/`VarnishTestCase`/`NginxTest`: these classes are there now purely for convenience, as they contain no logic but just collect the appropriate traits.

**Disadvantage**: we have some code duplication in the getters, e.g. `getHostName()`, `getCachingProxyPort()`.

Todo (if accepted):
- [x] update docs on how to include traits


src/Test/AppClient.php
```diff
@@ -0,0 +1,114 @@
...
+use Http\Discovery\MessageFactoryDiscovery;
+use Http\Discovery\UriFactoryDiscovery;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Provides a HTTP client for getting responses from your application
+ */
+trait AppClient
```
ddeboer: Is this a good name? Perhaps just `HttpClient` or `AppHttpClient` or `HttpAdapter`? I think the name should show that this is *not* a client for your proxy, but a client to simulate requests to your application.

src/Test/AppClient.php
```diff
@@ -0,0 +1,114 @@
...
+     * @param string $uriString
+     *
+     * @return UriInterface
+     */
+    protected function createUri($uriString)
+    {
+        return UriFactoryDiscovery::find()->createUri($uriString);
+    }
+
+    abstract protected function getHostName();
```
ddeboer: Users need to either implement this method themselves, or (preferably) use one of the `VarnishTest/NginxTest/SymfonyTest` traits.

src/Test/CacheAssertions.php
```diff
@@ -0,0 +1,54 @@
...
+namespace FOS\HttpCache\Test;
+
+use FOS\HttpCache\Test\PHPUnit\IsCacheHitConstraint;
+use FOS\HttpCache\Test\PHPUnit\IsCacheMissConstraint;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Provides cache hit/miss assertions to PHPUnit tests
+ */
+trait CacheAssertions
```
ddeboer: I did not add the `Trait` suffix, which goes against [Symfony’s coding standards](http://symfony.com/doc/current/contributing/code/standards.html#naming-conventions) but makes [more sense](http://verraes.net/2013/09/sensible-interfaces/) to me.

dbu: yeah, old discussion. the same goes for `Interface` suffixes. i am fine with this.

src/Test/AppClient.php
dbu: i wonder if it makes sense that all of this is a trait. could we have something like a client provider instead?

src/Test/HttpCaller.php
```diff
@@ -0,0 +1,60 @@
...
+ * Provides a method for getting responses from your application
+ */
+trait HttpCaller
+{
+    /**
+     * HTTP adapter for requests to the application
+     *
+     * @var HttpClient
+     */
+    protected $client;
```
dbu: can this be private? what does it mean exactly when a trait has a private property, is it hidden from the class that uses the trait?

from @stof in https://github.com/FriendsOfSymfony/FOSHttpCache/pull/224/files#r33294271

src/Test/HttpClient.php
```diff
@@ -0,0 +1,132 @@
...
+
+        return $this->getHttpAdapter()->sendRequest($request);
+    }
+
+    /**
+     * Get HTTP adapter for your application
+     *
+     * @return HttpAdapter
+     */
+    protected function getHttpAdapter()
```
stof: these methods should be private IMO

dbu:
@ddeboer can you wrap this one up? make things private where possible and update the testing doc? otherwise looks good to me.

ddeboer:
Docs updated.

doc/testing-your-application.rst
```diff
@@ -151,59 +140,91 @@ Constant                Getter                    Default
                                                                                                    running PHPUnit.
 ======================= ========================= ================================================ ===========================================
 
-Enable Assertions
-'''''''''''''''''
+HttpCaller Trait
+~~~~~~~~~~~~~~~~
+
+Provides your tests with a ``getResponse`` method, which retrieves a URI from
+your application through the HTTP cache::
```
dbu: maybe "through a real HTTP call to go through the HTTP cache" to be more clear?

doc/testing-your-application.rst
```diff
@@ -4,22 +4,30 @@ Testing Your Application
...
-* independent tests: all previously cached content is removed in the tests
-  ``setUp`` method. The way this is done depends on which reverse proxy you use;
-* an instance of this library’s client that is configured to talk to your
-  reverse proxy server. See reverse proxy specific sections for details;
-* convenience methods for executing HTTP requests to your application:
-  ``$this->getHttpAdapter()`` and ``$this->getResponse()``;
-* custom assertions ``assertHit`` and ``assertMiss`` for validating a cache
+By running your tests against a live instance of your reverse proxy, you can
+validate the caching headers that your application sets, and the invalidation
+rules that it defines.
```
dbu: i would say something like "you can validate that the configuration of your caching proxy works correctly with the caching headers and invalidation requests your application is doing."

oh btw, do we call this thing reverse proxy or caching proxy now?

doc/testing-your-application.rst
```diff
@@ -4,22 +4,30 @@ Testing Your Application
...
+
+The FOSHttpCache library provides traits and base test classes to help you write
+functional tests. Using the traits, you can extend your own (or your
```
dbu: i would call this "integration tests" instead of "functional tests"

doc/testing-your-application.rst
```diff
@@ -4,22 +4,30 @@ Testing Your Application
...
+framework’s) base test classes. For convenience, you can also extend the
+FOSHttpCache base test classes, as they include a sensible set of traits by
```
dbu: is there more than one base class we provide? otherwise name it explicitly.

doc/testing-your-application.rst
```diff
@@ -4,22 +4,30 @@ Testing Your Application
...
+default.
+
+By using the traits, you get:
+
+* independent tests: all previously cached content is removed in the test’s
+  ``setUp()`` method;
+* an instance of this library’s proxy client that is configured to talk to your
+  proxy server;
```
dbu: ... for invalidation requests;

doc/testing-your-application.rst
```diff
@@ -4,22 +4,30 @@ Testing Your Application
...
+* a convenience method for executing (cached) HTTP requests to your application:
```
dbu: for executing HTTP requests to your caching proxy:

doc/testing-your-application.rst
```diff
@@ -56,89 +64,70 @@ You can override getters in your test class in the following way::
...
 
-Configuration
-'''''''''''''
+Caching Proxy Server Traits
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+FOSHttpCache provides three caching proxy traits that:
+
+* if necessary, start your caching proxy server before running the tests
+* clear any cached content between tests, so you can be sure they are isolated
```
dbu: between tests, to guarantee test isolation;

doc/testing-your-application.rst
```diff
@@ -56,89 +64,70 @@ You can override getters in your test class in the following way::
...
+------
 
-Configuration
-'''''''''''''
+Caching Proxy Server Traits
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+FOSHttpCache provides three caching proxy traits that:
+
+* if necessary, start your caching proxy server before running the tests
```
dbu: needs `;` at end of line

doc/testing-your-application.rst
```diff
@@ -56,89 +64,70 @@ You can override getters in your test class in the following way::
...
 ``VARNISH_VERSION``     ``getVarnishVersion()``   ``4``                                              installed varnish application version
-``WEB_SERVER_HOSTNAME`` ``getHostName()``                                                            hostname your application can be reached at
 ======================= ========================= ================================================== ===========================================
 
-Enable Assertions
-'''''''''''''''''
-
-For the `assertHit` and `assertMiss` assertions to work, you need to add a
-:ref:`custom X-Cache header <varnish_debugging>` to responses served
-by your Varnish.
```
dbu: do we not need this anymore, or is it handled elsewhere?

doc/testing-your-application.rst
```diff
@@ -151,59 +140,91 @@ Constant                Getter                    Default
...
+Provides your tests with a ``getResponse`` method, which retrieves a URI from
+your application through the HTTP cache::
 
-For the `assertHit` and `assertMiss` assertions to work, you need to add debug
-information in your AppCache. Create the cache kernel with the option
-``'debug' => true`` and add the following to your ``AppCache``::
+    use FOS\HttpCache\Test\HttpCaller;
 
-    public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQUEST, $catch = true)
+    class YourTest extends \PHPUnit_Framework_TestCase
```
dbu: don't we need something about the traits here?

doc/testing-your-application.rst
```diff
@@ -151,59 +140,91 @@ Constant                Getter                    Default
...
-'''''''''''''''''
+HttpCaller Trait
+~~~~~~~~~~~~~~~~
+
+Provides your tests with a ``getResponse`` method, which retrieves a URI from
+your application through the HTTP cache::
 
-For the `assertHit` and `assertMiss` assertions to work, you need to add debug
-information in your AppCache. Create the cache kernel with the option
-``'debug' => true`` and add the following to your ``AppCache``::
```
dbu: don't we need this anymore?

dbu: ah sorry, moved below

doc/testing-your-application.rst
dbu: ah, explained below

doc/testing-your-application.rst
```diff
@@ -151,59 +140,91 @@ Constant                Getter                    Default
...
-            // Varnish is restarted, so you don’t have to worry about previously
-            // cached content. Before continuing, the VarnishTestCase waits for
-            // Varnish to become available.
+            // The caching proxy is (re)started, so you don’t have to worry
+            // about  previously cached content. Before continuing, the
+            // VarnishTest/ NginxTest trait waits for the caching proxy to
+            // become available.
 
-            // Retrieve an URL from your application
+            // Retrieve a URL from your application
```
dbu: i think "an URL" was correct.

src/Test/CacheAssertions.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+namespace FOS\HttpCache\Test;
+
+use FOS\HttpCache\Test\PHPUnit\IsCacheHitConstraint;
+use FOS\HttpCache\Test\PHPUnit\IsCacheMissConstraint;
+use Psr\Http\Message\ResponseInterface;
+
+/**
+ * Provides cache hit/miss assertions to PHPUnit tests
```
dbu: we should repeat the note about the X-Cache header here, pointing to the doc.

dbu: and also that using this without a trait that resets the cache after each test is likely to create a problem with process isolation.

src/Test/ProxyTestCase.php
```diff
@@ -11,190 +11,15 @@
...
 use Psr\Http\Message\RequestInterface;
 use Psr\Http\Message\ResponseInterface;
 use Psr\Http\Message\UriInterface;
 
 /**
- * Abstract caching proxy test case
- *
+ * Abstract test that provides an application HTTP client and cache assertions
  */
 abstract class ProxyTestCase extends \PHPUnit_Framework_TestCase
```
dbu: should we drop this class? for just the two use trait statements it seems a bit overcomplicated.

src/Test/VarnishTestCase.php
```diff
@@ -11,152 +11,10 @@
...
- *                      (default varnishd)
- * VARNISH_PORT         port Varnish listens on (default 6181)
- * VARNISH_MGMT_PORT    Varnish management port (default 6182)
- * VARNISH_CACHE_DIR    directory to use for cache
- *                      (default sys_get_temp_dir() + '/foshttpcache-varnish')
- * VARNISH_VERSION      Version of varnish: 3 or 4, defaults to 4.
- * WEB_SERVER_HOSTNAME  hostname/IP your application can be reached at (required)
+ * Abstract test that contains traits necessary for running tests against Varnish
  */
 abstract class VarnishTestCase extends ProxyTestCase
```
dbu: does it make sense to provide this test case? it is basically an empty class, so we could just tell people to use the trait instead of having two alternate way of doing things.

ddeboer: Yeah, this base classes is there for convenience only: when you extend this base class all traits that you might need are included by default, so you don’t need to worry about forgetting one.

doc/testing-your-application.rst
ddeboer: Good one, let’s stick to caching proxy (server).

doc/testing-your-application.rst
ddeboer: Also `NginxTest` and `SymfonyTest`. They are just trait bags now, so just remove them?

doc/testing-your-application.rst
ddeboer: But we call our tests ‘functional tests’ (they are even in the tests/Functional directory).

doc/testing-your-application.rst
ddeboer: No, a/an depends on the *sound* that follows it (here *you*), not the letter.

doc/testing-your-application.rst
ddeboer: I think the trait is used here and described above.

src/Test/CacheAssertions.php
ddeboer: Done.

doc/testing-your-application.rst
dbu: okay. then we should move those tests that are integration tests. imho:

* Unit test: tests a single method of a class
* Functional test: Tests the interaction of the application, e.g.
WebTestCase
* Integration test: Tests the interaction of several different systems,
e.g. a PHP application and Varnish

doc/testing-your-application.rst
dbu: i will remind some english speaker next time he complains about german
being complicated :-D

src/Test/VarnishTestCase.php
dbu: yep. well okay, lets keep them. when writing several test classes it
will still be handy not to need to `use` all traits all the time.

doc/testing-your-application.rst
dbu: lets say: "you can also extend the FOSHttpCache base test class suitable for your caching proxy, as"...

doc/testing-your-application.rst
dbu: but we miss a `use HttpCaller;` line inside the class to be complete.

src/Test/VarnishTestCase.php
dbu: can we eliminate the ProxyTestCase to reduce the depth of inheritance by one level? otherwise this seems ready to merge.

doc/testing-your-application.rst
dbu: what about this one?

doc/testing-your-application.rst
dbu: well, lets leave this one as is. good enough.

dbu:
this does not fix #213, right? can we wrap this one up and then work on 213?

src/Test/VarnishTestCase.php
ddeboer: Done.

ddeboer:
Updated; now ready to merge.

Indeed, this PR enables users of the lib to test their own applications without having to extend our test classes by using our treats instead. It does not fix #213, as that is about code duplication in the code that tests the lib itself.

doc/symfony-cache-configuration.rst
```diff
@@ -179,4 +179,34 @@ the ``session_name_prefix`` option) in the requests to the backend. If you need
...
+                $state = 'UNDETERMINED';
+            }
+            $response->headers->set('X-Cache', $state);
+        }
+
+        return $response;
+    }
+
+The ``UNDETERMINED`` state should never happen. If it does, it means that your
+HttpCache is not correctly set into debug mode.
```
dbu: HttpCache? also, should we make this into another subscriber that one can simply activate in app_test.php?

src/Test/VarnishTestCase.php
```diff
@@ -11,152 +11,12 @@
...
- * VARNISH_PORT         port Varnish listens on (default 6181)
- * VARNISH_MGMT_PORT    Varnish management port (default 6182)
- * VARNISH_CACHE_DIR    directory to use for cache
- *                      (default sys_get_temp_dir() + '/foshttpcache-varnish')
- * VARNISH_VERSION      Version of varnish: 3 or 4, defaults to 4.
- * WEB_SERVER_HOSTNAME  hostname/IP your application can be reached at (required)
+ * Abstract test that contains traits necessary for running tests against Varnish
  */
-abstract class VarnishTestCase extends ProxyTestCase
+abstract class VarnishTestCase
```
stof: this is broken. It needs to extend from the PHPUnit_Framework_TestCase

src/Test/ProxyTestCase.php
```diff
@@ -1,200 +0,0 @@
...
-use Http\Discovery\UriFactoryDiscovery;
-use Psr\Http\Message\RequestInterface;
-use Psr\Http\Message\ResponseInterface;
-use Psr\Http\Message\UriInterface;
-
-/**
- * Abstract caching proxy test case
- *
- */
-abstract class ProxyTestCase extends \PHPUnit_Framework_TestCase
```
stof: removing this class is a BC break if someone was extending it directly when writing their tests

dbu: added extending the phpunit test case to the 3 proxy specific base cases. we are in the process of moving to version 2 so removing unnecessary classes is ok. i will add a note to the changelog.

src/Test/VarnishTestCase.php
dbu: it seems to not be needed, at least the travis tests did run through without. oh, or maybe phpunit now ignores classes called Test if they dont extend the phpunit TestCase class.

but anyways i added it back to avoid confusion.

dbu:
@ddeboer i rebased and did some cleanup 

only open point from my side is https://github.com/FriendsOfSymfony/FOSHttpCache/pull/218#discussion_r36509179 - but i think we should just clean up the text to make it better understandable. providing a trait can happen in a separate PR.

src/Test/VarnishTestCase.php
stof: well, if it does not extend the base testcase class, it won't run as it won't be a PHPUnit testcase.

https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/jobs/74553567 was running 78 tests while https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/jobs/74556258 ran 112 tests

dbu: oh, indeed. now that i added it back, we are back at 112 tests.

doc/symfony-cache-configuration.rst
ddeboer: I’m not sure about this; we already had this in the docs and I just moved it. :+1: on a `DebugSubscriber`, but let’s do that in a separate PR.

dbu: created #238 for that

PR #219, opened 2015-06-24
update phpdoc for TagHandler::invalidateTags

dbu:
fix https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/228

### DIFF
--- a/src/Handler/TagHandler.php
+++ b/src/Handler/TagHandler.php
@@ -106,6 +106,11 @@ class TagHandler
      * Invalidate cache entries that contain any of the specified tags in their
      * tag header.
      *
+     * The cache manager is told to invalidate all content with the specified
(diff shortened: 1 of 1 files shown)


src/Handler/TagHandler.php
```diff
@@ -106,6 +106,10 @@ public function addTags(array $tags)
      * Invalidate cache entries that contain any of the specified tags in their
      * tag header.
      *
+     * The cache manager is told to invalidate all content with the specified tags
+     * and invalidation is dispatched on request termination. To immediately
```
ddeboer: Not just request termination, right? Also command termination, so more generally, kernel termination.

dbu: true, kernel terminate would be the right thing to say here. maybe even be explicit that commands are covered as well.

dbu:
improved the wording

PR #224, opened 2015-06-25
separate http client from trait

dbu:


src/Test/HttpCaller.php
```diff
@@ -0,0 +1,60 @@
...
+ * Provides a method for getting responses from your application
+ */
+trait HttpCaller
+{
+    /**
+     * HTTP adapter for requests to the application
+     *
+     * @var HttpClient
+     */
+    protected $client;
```
stof: does it need to be protected ?

PR #227, opened 2015-06-25
Use short array syntax

ddeboer:
Just because we can. :smile:


doc/proxy-clients.rst
```diff
@@ -193,7 +193,7 @@ You can specify HTTP headers as the second argument to ``purge()``.
 For instance::
 
     $client
-        ->purge('/some/path', array('X-Foo' => 'bar')
```
ddeboer: Found some syntax errors here. :winkK

PR #233, opened 2015-08-02
Provide Symfony HttpCache as trait

### LINKED ISSUE #191: Symfony HttpCache trait instead of class
When we stop php 5.3 support, we should convert the EventDispatchingHttpCache into a trait to reuse the logic with the FOSHttpCacheBundle kernel which extends a different HttpCache class and thus currently needs to copy the methods.

### LINKED ISSUE #238: debug subscriber for symfony http cache
The debug subscriber shown in doc/symfony-cache-configuration.rst should be a real subscriber with tests instead.

ddeboer:
Fix #191, fix #238.

* Provide event dispatching HttpCache as trait
* Move debug header code to a subscriber

- [x] Unit tests for debug subscriber
- [x] Update docs. mention the unit test for event dispatching kernel, should be used whenever using this kernel. and actually most custom extensions to HttpCache should probably be done as event subscribers in this scenario, to keep things understandable.

### DIFF
--- a/src/SymfonyCache/EventDispatchingHttpCache.php
+++ b/src/SymfonyCache/EventDispatchingHttpCache.php
@@ -11,27 +11,36 @@
 
 namespace FOS\HttpCache\SymfonyCache;
 
-use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\HttpCache\HttpCache;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
 /**
- * Base class for enhanced Symfony reverse proxy based on the symfony component.
+ * Trait for enhanced Symfony reverse proxy based on the symfony kernel component.
  *
- * <b>When using FOSHttpCacheBundle, look at FOS\HttpCacheBundle\HttpCache instead.</b>
+ * Your kernel needs to implement CacheInvalidatorInterface and redeclare the
+ * fetch method as public. (The latter is needed because the trait declaring it
+ * public does not satisfy the interface for whatever reason. See also
+ * http://stackoverflow.com/questions/31877844/php-trait-exposing-a-method-and-interfaces )
  *
- * This kernel supports event subscribers that can act on the events defined in
- * FOS\HttpCache\SymfonyCache\Events and may alter the request flow.
+ * CacheInvalidator kernels support event subscribers that can act on the
+ * events defined in FOS\HttpCache\SymfonyCache\Events and may alter the
+ * request flow.
+ *
+ * If your kernel overwrites any of the methods defined in this trait, make
+ * sure to also call the trait method. You might get into issues with the order
+ * of events, in which case you will need to copy event triggering into your
+ * kernel.
  *
  * @author Jérôme Vieilledent <lolautruche@gmail.com> (courtesy of eZ Systems AS)
+ * @author David Buchmann <mail@davidbu.ch>
  *
  * {@inheritdoc}
  */
-abstract class EventDispatchingHttpCache extends HttpCache implements CacheInvalidationInterface
+trait EventDispatchingHttpCache
 {
     /**
      * @var EventDispatcherInterface
@@ -69,17 +78,13 @@ abstract class EventDispatchingHttpCache extends HttpCache implements CacheInval
      */
     public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQUEST, $catch = true)
     {
-        if ($this->getEventDispatcher()->hasListeners(Events::PRE_HANDLE)) {
-            $event = new CacheEvent($this, $request);
-            $this->getEventDispatcher()->dispatch(Events::PRE_HANDLE, $event);
-            if ($event->getResponse()) {
-                return $this->dispatchPostHandle($request, $event->getResponse());
-            }
+        if ($response = $this->dispatch(Events::PRE_HANDLE, $request)) {
+            return $this->dispatch(Events::POST_HANDLE, $request, $response);
         }
 
         $response = parent::handle($request, $type, $catch);
 
-        return $this->dispatchPostHandle($request, $response);
+        return $this->dispatch(Events::POST_HANDLE, $request, $response);
     }
 
     /**
@@ -89,42 +94,9 @@ abstract class EventDispatchingHttpCache extends HttpCache implements CacheInval
      */
     protected function store(Request $request, Response $response)
     {
-        if ($this->getEventDispatcher()->hasListeners(Events::PRE_STORE)) {
-            $event = new CacheEvent($this, $request, $response);
-            $this->getEventDispatcher()->dispatch(Events::PRE_STORE, $event);
-            $response = $event->getResponse();
-        }
+        $response = $this->dispatch(Events::PRE_STORE, $request, $response);
 
         parent::store($request, $response);
-    }
-
-    /**
-     * Dispatch the POST_HANDLE event if needed.
-     *
-     * @param Request  $request
-     * @param Response $response
-     *
-     * @return Response The response to return which might be altered by a POST_HANDLE listener.
-     */
-    private function dispatchPostHandle(Request $request, Response $response)
-    {
-        if ($this->getEventDispatcher()->hasListeners(Events::POST_HANDLE)) {
-            $event = new CacheEvent($this, $request, $response);
-            $this->getEventDispatcher()->dispatch(Events::POST_HANDLE, $event);
-            $response = $event->getResponse();
-        }
-
-        return $response;
-    }
-
-    /**
-     * Made public to allow event subscribers to do refresh operations.
-     *
-     * {@inheritDoc}
-     */
-    public function fetch(Request $request, $catch = false)
-    {
-        return parent::fetch($request, $catch);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


tests/Functional/Fixtures/Symfony/AppCache.php
```diff
@@ -25,4 +29,14 @@ public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQ
 
         return $response;
     }
+
+    /**
+     * Made public to allow event subscribers to do refresh operations.
+     *
+     * {@inheritDoc}
+     */
+    public function fetch(Request $request, $catch = false)
```
ddeboer: This needs to be overridden here because of trait [precedence](http://php.net/manual/en/language.oop5.traits.php#language.oop5.traits.precedence).

tests/Unit/SymfonyCache/EventDispatchingHttpCacheTest.php
```diff
@@ -142,6 +144,16 @@ public function testAbortInvalidate()
     }
 }
 
+class AppCache extends HttpCache implements CacheInvalidationInterface
+{
+    use EventDispatchingHttpCache;
+
+    public function fetch(Request $request, $catch = false)
+    {
+        return $this->fetch($request, $catch);
```
dbu: if this is not creating an endless loop, you probably need to do the same in the other HttpCache above

tests/Functional/Fixtures/Symfony/AppCache.php
```diff
@@ -25,4 +29,14 @@ public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQ
...
+    {
+        return parent::fetch($request, $catch);
```
dbu: i suspect that parent:: does not go to the trait but directly the parent. do we need to remap fetch from the trait to eventDispatchingFetch? or even change the trait method as it will never work when there is a name collision? looks like the most tricky bit of the trait, but don't see how we could avoid that.

tests/Functional/Fixtures/Symfony/AppCache.php
dbu: strange that a trait can't change visibility of a method - or is there a way that would allow it?

dbu:
very cool! this will get us rid of a lot of duplication needs.

tests/Functional/Fixtures/Symfony/AppCache.php
dbu: so actually it does change visibility but there seems to be some bug with interface vs trait somewhere: http://stackoverflow.com/questions/31877844/php-trait-exposing-a-method-and-interfaces

dbu:
@ddeboer i investigated a bit and at least found a workaround, albeit not an elegant one. there is one test reliably failing now: SymfonyProxyClientTest::testRefresh

i have to leave now, if you want to further investigate, please do.

tests/Functional/Fixtures/Symfony/AppCache.php
stof: parent is indeed not going to the trait

tests/Functional/Fixtures/Symfony/AppCache.php
dbu: @stof this is the solution i came up with, because of the problem i explain on stackoverflow. its annoying but i guess its really a bug (or illogic decision) of traits vs interfaces.

tests/Functional/Fixtures/Symfony/AppCache.php
```diff
@@ -2,12 +2,17 @@
...
 use FOS\HttpCache\SymfonyCache\EventDispatchingHttpCache;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\HttpCache\HttpCache;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
-class AppCache extends EventDispatchingHttpCache
+class AppCache extends HttpCache implements CacheInvalidationInterface
 {
+    // http://stackoverflow.com/questions/31877844/php-trait-exposing-a-method-and-interfaces
+    use EventDispatchingHttpCache {fetch as public eventTriggeringFetch;}
```
stof: is it needed to be public ?

dbu: yes, because of https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/SymfonyCache/CacheInvalidationInterface.php or more specifically because of https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/SymfonyCache/RefreshSubscriber.php

src/SymfonyCache/EventDispatchingHttpCache.php
```diff
@@ -89,28 +93,25 @@ public function handle(Request $request, $type = HttpKernelInterface::MASTER_REQ
...
-     * @param Response $response
+     * @param string        $name     Name of the event to trigger. One of the constants in FOS\HttpCache\SymfonyCache\Events
+     * @param Request       $request
+     * @param Response|null $response If already available
      *
-     * @return Response The response to return which might be altered by a POST_HANDLE listener.
+     * @return Response The response to return, which might be provided/altered by a listener.
      */
-    private function dispatchPostHandle(Request $request, Response $response)
+    protected function dispatch($name, Request $request, Response $response = null)
```
dbu: made this protected so that kernels using the trait can also use it, if they need to expand on a method and still need the pre and post event.

src/SymfonyCache/EventDispatchingHttpCache.php
```diff
@@ -118,28 +119,14 @@ private function dispatchPostHandle(Request $request, Response $response)
     }
 
     /**
-     * Made public to allow event subscribers to do refresh operations.
-     *
-     * {@inheritDoc}
-     */
-    public function fetch(Request $request, $catch = false)
```
dbu: removing as it does not work anyways

src/Test/PHPUnit/AbstractCacheConstraint.php
```diff
@@ -41,17 +43,25 @@ public function __construct($header = null)
 
     /**
      * {@inheritdoc}
+     *
+     * @param Response $other The guzzle response object
      */
     protected function matches($other)
     {
+        if (!$other instanceof ResponseInterface) {
+            throw new \RuntimeException(sprintf('Expected a GuzzleHttp\Psr7\Response but got %s', get_class($other)));
```
dbu: i realized we are assuming things here that could lead to fatal errors during the tests.

src/Test/PHPUnit/AbstractCacheConstraint.php
```diff
@@ -41,17 +43,25 @@ public function __construct($header = null)
...
-                    . 'to set the header with cache hit/miss status.',
-                    $this->header
-                )
+            $message = sprintf(
+                'Response has no "%s" header. Configure your caching proxy '
+                . 'to set the header with cache hit/miss status.',
+                $this->header
             );
+            if (200 !== $other->getStatusCode()) {
+                $message .= sprintf("\nStatus code of response is %s.", $other->getStatusCode());
```
dbu: and an error inside the kernel could abort the request. this is an additional hint if we expect a successful response but did not get one

dbu:
so this would be ready for a code review. doc is missing and some unit tests for the debug subscriber.

dbu:
@lolautruche can you or somebody else at ez have a look at this? it will affect ez i think, as you use the symfony http cache. i hope its an improvement. BC might be a bit tricky however - but then supporting both FOSHttpCache 1.x and 2.x would be really tricky i think, hope you do not need that.

tests/Functional/Fixtures/Symfony/AppCache.php
stof: none of this is calling ``eventTriggeringFetch``

dbu: this is cleaned up, thanks.

ddeboer:
@dbu Thanks for fixing the test failure. Can you squash your commits? 

@lolautruche Will this change become a problem for eZ? Please keep in mind this change will only make it to the 2.0 version of the library.

dbu:
squashed and rebased. lets give @lolautruche a few more days to have a look at this before merging.

src/SymfonyCache/EventDispatchingHttpCache.php
```diff
@@ -11,27 +11,35 @@
...
+use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
 /**
- * Base class for enhanced Symfony reverse proxy based on the symfony component.
+ * Trait for enhanced Symfony reverse proxy based on the symfony kernel component.
  *
- * <b>When using FOSHttpCacheBundle, look at FOS\HttpCacheBundle\HttpCache instead.</b>
+ * Your kernel needs to implement CacheInvalidatorInterface and redeclare the
+ * fetch method as public. (The later is needed because the trait declaring it
```
lolautruche: s/later/latter

lolautruche:
Hi

AFAICS, we can be able to use the trait without any hassle for the end developer since we rely on [our own abstract `HttpCache`](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/HttpCache.php). We'd do the change here to use the trait, which is fine since we require PHP 5.4+.
As for v2.0 of FOSHttpCache, any ETA? I also need to look into the changelog to be sure our usage is not broken :wink: .

Thanks for noticing (and sorry for the delay).

lolautruche:
Ping @andrerom for review as well

dbu:
> Thanks for noticing (and sorry for the delay).

thanks for the review. so it seems it will also get a bit more elegant
for ez with the trait.

> As for v2.0 of FOSHttpCache, any ETA? I also need to look into the
> changelog to be sure our usage is not broken :wink: .

i think we need to wait for php-http (a http client abstraction so that
we don't depend on any specific implementation like a specific guzzle
version) to stabilize before we can call FOSHttpCache stable. the plan
there is to go alpha in the next days afaik, so i would hope we can
start tagging alpha in october and hopefully release end of october.

lolautruche:
> i think we need to wait for php-http

You mean the PHP extension?

dbu:
@lolautruche no, this thing: http://php-http.readthedocs.org/en/latest/ - it will get a more unique name soon as php-http is not google-friendly.

lolautruche:
Ah OK, cool

andrerom:
> Ping @andrerom for review as well

Looks ok for me, sensible to move this into trait anyway.

> i think we need to wait for php-http (a http client abstraction so that
we don't depend on any specific implementation like a specific guzzle
version) to stabilize before we can call FOSHttpCache stable

make sense for 2.0


Side: @dbu / @ddeboer any chance of some feedback on the abstract tagging (#237) direction so we can have at least some abstraction in 2.0 to build upon for Sf HttpCache improvements?

dbu:
thanks. so lets get this merged.

@ddeboer any idea why hhvm fails: https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/jobs/82876644 ? i already restarted the test, so its not random failure. but it looks like a varnish config issue - but as its specifically hhvm this is strange.

ddeboer:
It’s probably not the Varnish config per se, but HHVM not serving the responses correctly: I’m seeing 404s and 503s (backend fetch failed).

dbu:
i just restarted the last build of master in https://travis-ci.org/FriendsOfSymfony/FriendsOfSymfony/FOSHttpCache/jobs/77916788

it now fails with the same pattern. i think this means that travis updated their hhvm and we see some regression. ok if we merge as-is and look into fixing hhvm in a separate pull request?

dbu:
we see the same test failure on master as well. @ddeboer wdyt, should we just merge this? and would you have some time to look into what hhvm is unhappy about?

dbu:
fixed the last typo from the comments. @ddeboer is it ok if we merge this? hhvm fails the same way as on master, so imho we should not block this merge. and fix hhvm against master, or disable it if we can't fix it.

PR #235, opened 2015-08-06
Use the HttpFoundation API to read the header in the test kernel

stof:


dbu:
thanks, that is cleaner.

PR #236, opened 2015-08-07
Fix the upload of the code coverage to Scrutinizer

stof:
PHP 7 and HHVM are not able to generate code coverage on Travis. If ocular is called in this case, it will report to Scrutinizer that the code coverage has not been collected, which will cancel the scrutinizer analysis if it is the first Travis job to finish (which is likely as generating code coverage would slow down the other jobs).


dbu:
thanks a lot!

Issue #230, opened 2015-07-04
Add PSR-7 TagHandler::tagResponse()


ddeboer:
Currently, we only have `tagResponse()` in the bundle (for Symfony responses). Now PSR-7 has been accepted, however, we could also add this method to the library, setting tags on a PSR-7 response. 

It’s then up for debate whether we want to override this in the bundle, or use Symfony’s PSR-7 bridge to convert between the two response types.

dbu:
ah, indeed. this will become tricky with type hinting in the method
however. best would be if we find a way that works for both PSR 7 and
symfony. in a non-surprising way. maybe no typehint but instanceof checks?

ddeboer:
Well, so far the library does not depend on Symfony requests/responses. I suggest we keep it that way and implement `tagResponse()` in the library (PSR-7 response) and override/decorate it in the bundle (Symfony response).

dbu:
agreed. but we need to not have a type hint on `tagResponse` or we get
into a conflict with
https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/Handler/TagHandler.php#L20
(will need to change the bundle method signature too, of course).

then in the bundle we check instanceof symfony Response and if so handle
it, otherwise call the lib, which checks instanceof psr7 Response and
handles or fails on anything else.

PR #237, opened 2015-08-09
Add abstracted tag handling

### LINKED ISSUE #234: Support Tags with Symfony Http cache
Opening this to continue the conversation of adding such a feature.
In our (eZ) case we would like to use tags both when configured with Varnish and when using Sf Cache.

I'm not 100% up to speed on EventDispatchingHttpCache as introduced earlier this year, but possible idea as discussed with @lolautruche was to use hardlinks (see https://github.com/symfony/symfony/pull/15458) to place the cache in several locations representing the tags, to be able to efficiently do reverse lookup on <del>"banning", or *to allow stale/grace cache;* expiry by tag</del> <ins>expiry</ins>. Discussed that with @lsmith77 some time ago and decided to take the discussion here.

andrerom:
Aims to allow support for tagging also on Symfony HTTP Cache and
potentially other caches by limiting the feature to focus on only
tags and not random headers.

Closes #234 

* [x] explain changes in CHANGELOG.md
* [x] Update documentation


README.md
```diff
@@ -12,6 +12,9 @@ This library integrates your PHP applications with HTTP caching proxies such as
 Use this library to send invalidation requests from your application to the caching proxy
 and to test your caching and invalidation code against a Varnish setup.
 
+It does this by abstracting some caching concepts and attempting to make sure these
+can be supported across Varnish, Nginx and Symfony HTTP Cache.
```
dbu: i suggest: 

... and Symfony `HttpCache`.

src/Handler/TagHandler.php
```diff
@@ -117,9 +119,14 @@ public function addTags(array $tags)
...
-        $tagExpression = sprintf('(%s)(,.+)?$', implode('|', array_map('preg_quote', $this->escapeTags($tags))));
-        $headers = [$this->tagsHeader => $tagExpression];
-        $this->invalidator->invalidate($headers);
+        if (empty($tags)) {
+            return $this;
+        }
+
+        $this->invalidator->invalidateTags(
+            $this->escapeTags($tags),
+            $this->tagsHeader
```
dbu: i think we should move this out of here and the CacheInvalidator. if a header is needed, its up to the client to know that, its an implementation detail specific to the caching proxy.

src/ProxyClient/Invalidation/TagsInterface.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Remove/Expire cache objects based on cache tags
+     *
+     * @param array $tags Tags that should be removed/expired from the cache
+     * @param string $tagsHeader Name of the HTTP header used to provide the tags.
+     *
+     * @return $this
+     */
+    public function tags(array $tags, $tagsHeader = 'X-Cache-Tags');
```
dbu: i think we should call this invalidateTags or something. the other methods are nice short ban/purge/refresh, but all of those words are verbs, while tag is not a verb and does not say whether you want to add tags to something or to invalidate the tags. i prefer verbose over implicit/contextual here.

src/CacheInvalidator.php
```diff
@@ -198,6 +206,27 @@ public function invalidate(array $headers)
...
+     * @see TagsInterface::tags()
+     *
+     * @param array $tags Tags that should be removed/expired from the cache
+     * @param string $tagsHeader Name of the HTTP header used to provide the tags
+     *
+     * @throws UnsupportedProxyOperationException If HTTP cache does not support Tags invalidation
+     *
+     * @return $this
+     */
+    public function invalidateTags(array $tags, $tagsHeader = 'X-Cache-Tags')
```
dbu: leaving $tagsHeader out of here would also have the benefit of keeping this method the same as we had in version 1 (and deprecated because the TagHandler was supposed to do this).

src/Handler/TagHandler.php
```diff
@@ -117,9 +119,14 @@ public function addTags(array $tags)
...
     {
-        $tagExpression = sprintf('(%s)(,.+)?$', implode('|', array_map('preg_quote', $this->escapeTags($tags))));
-        $headers = [$this->tagsHeader => $tagExpression];
-        $this->invalidator->invalidate($headers);
+        if (empty($tags)) {
+            return $this;
+        }
+
+        $this->invalidator->invalidateTags(
+            $this->escapeTags($tags),
```
dbu: there is an issue here: the TagHandler does the escaping of the tags, but people could start calling CacheInvalidator::invalidateTags directly. at this point, the tag handler is down to a collection of tags. not sure if that is actually worth anything anymore. 

and actually, the escaping is again caching proxy specific. so instead, we should move escaping to the varnish client and ditch the tag handler completely i think. the added value of something that accumulates tags that we will invalidate is very small - one can just invalidate several times if needed, or build something to aggregate the tags.

dbu:
@ddeboer can you please also give your input here? there are some architectural decisions to take, would appreciate your opinion.

src/Handler/TagHandler.php
andrerom: Ok, but then we remove the api's here as well for it right?

src/ProxyClient/Invalidation/TagsInterface.php
andrerom: +1

src/Handler/TagHandler.php
andrerom: I'm fine with that, had the same concern.

But then we agree that we also remove `getTagsHeaderValue` to make this only focus on tags then, right?

dbu: ah, monday morning... now i realize that TagHandler is also about adding
tags to responses. i guess we want to keep that part. do you plan to use
a header for tagging responses with HttpCache as well?

either way we would want to keep that part. which raises the question
where the escaping should be. maybe still in the TagHandler but public
so that caching clients can use them? or should we keep invalidateTags
on TagHandler in the end and document on CacheInvalidator that the end
user is not allowed to call invalidateTags there, mark the public method
as `@private` in phpdoc?

andrerom: > do you plan to use a header for tagging responses with HttpCache as well?

Kind of, but I planning to ask you and @ddeboer how this can be done in a clean way.
Could also be that the ProxyClients -> drivers gets involved in generating the output.

andrerom:
Did an attempt to move tag header name, value and escaping down to cache invalidations, looks a bit better. I guess this implies changes is needed on FOSHttpCacheBundle as well now.

README.md
```diff
@@ -12,6 +12,9 @@ This library integrates your PHP applications with HTTP caching proxies such as
 Use this library to send invalidation requests from your application to the caching proxy
 and to test your caching and invalidation code against a Varnish setup.
 
+It does this by abstracting some caching concepts and attempting to make sure these
+can be supported across Varnish, Nginx and Symfony HTTPCache.
```
dbu: i think we wrote HttpCache elsewhere (the php class is written in camelcase)

dbu:
at this point, the only thing the TagHandler is doing is collecting the tags for tagging a response. that is really not much. i think we should either let TagHandler talk to the proxy client directly or get rid of the handler and move all to CacheInvalidator. i tend to keep TagHandler because its a whole topic of its own and not only about invalidation. I wonder if the TagInterface should be outside the Invalidation namespace, because it is not only about invalidating tags but also about tagging responses.

andrerom:
> I wonder if the TagInterface should be outside the Invalidation namespace, because it is not only about invalidating tags but also about tagging responses.

ok, so then I'll skip the header related parts added to CacheInvalidator as well right?

dbu:
sorry, still trying to get some input from @ddeboer what he thinks about
this one.

can we wait another day or two? if he does not respond by then we just
move forward with what we think is cleanest.

andrerom:
ok, no problem. Probably vacation :)

ddeboer:
That's right, I'm on holidays. Back in civilisation with proper internet connection in a week. :)

dbu:
@andrerom sorry to keep you waiting, but i would prefer to wait for the other david to be back before we take a decision on how to do this one. its a rather important architectural decision.

andrerom:
No problem, opened a possible followup issue to discuss in the mean time (#241).

andrerom:
@ddeboer any views on this topic?

ilukac:
This looks promising. With this feature we could better optimise invalidation...

src/ProxyClient/AbstractProxyClient.php
```diff
@@ -150,4 +150,22 @@ private function handleErrorResponses(array $responses)
...
+    /**
+     * Make sure that the tags are valid.
+     *
+     * Escapes `,` and `\n` (newline) characters.
+     *
+     * @param array $tags The tags to escape.
+     *
+     * @return array Sane tags.
+     */
+    protected function escapeTags(array $tags)
```
ddeboer: Why not private?

src/ProxyClient/Varnish.php
```diff
@@ -89,6 +97,35 @@ public function setDefaultBanHeader($name, $value)
     /**
      * {@inheritdoc}
      */
+    public function invalidateTags(array $tags)
+    {
+        $tagExpression = sprintf('(%s)(,.+)?$', implode('|', array_map('preg_quote', $this->escapeTags($tags))));
+
+        return $this->ban([$this->tagsHeader => $tagExpression]);
+    }
```
ddeboer: Why were these methods added back here, to the `CacheInvalidator`? Previously, we moved them out to the `TagHandler`.

tests/Unit/ProxyClient/VarnishTest.php
```diff
@@ -99,6 +99,43 @@ public function testBanPathEmptyHost()
...
+
+        $requests = $this->getRequests();
+
+        $this->assertCount(1, $requests);
+        $this->assertEquals('BAN', $requests[0]->getMethod());
+
+        $this->assertEquals('(post\-1|post\-type\-3)(,.+)?$', $requests[0]->getHeaderLine('X-Cache-Tags'));
+        $this->assertEquals('fos.lo', $requests[0]->getHeaderLine('Host'));
+
+        // This being taken into account, might not be expected from user POV?
```
ddeboer: How do you mean exactly?

README.md
ddeboer: Correct.

README.md
```diff
@@ -22,6 +25,7 @@ Features
 
 * Send [cache invalidation requests](http://foshttpcache.readthedocs.org/en/stable/cache-invalidator.html)
   with minimal impact on performance.
+* Cache tagging abstraction use of BAN with Varnish and allowing support for other caching proxies in the future.
```
ddeboer: This sentence is a bit weird, but we can correct it later.

ddeboer:
Very interesting! I agree with the general approach: we’ve been following Varnish terminology in differentiating purge, refresh and ban, but nothing is stopping is from introducing the separate tags type.

src/ProxyClient/AbstractProxyClient.php
andrerom: It can be

tests/Unit/ProxyClient/VarnishTest.php
andrerom: That tag invilaidation, now separate from ban, takes the extra ban options into account because of code reuse.

src/ProxyClient/AbstractProxyClient.php
andrerom: But this, and maybe some other parts will be needed for HttpCache as well

andrerom:
> @andrerom Can you elaborate a bit on why you did this? Especially since @dbu and I decided earlier to move tag-specific logic out of the CacheInvalidator and into the TagHandler

I'll need to read up on the discussions above where @dbu and me hit some issues with first approach, but think it was among other things because some of the escaping was clearly varnish specific.

ddeboer:
Right, in the discussion (on outdated diffs) I read:

> i think we should move this out of here and the `CacheInvalidator`. if a header is needed, its up to the client to know that, its an implementation detail specific to the caching proxy.

@dbu Did you mean here to move the tag header logic into CacheInvalidator? If we’re saying it’s *client*-specific, it should go into the proxy client classes (`Varnish`/`Nginx`/`Symfony`/`AbstractProxyClient`). After all, we use a header for tagging in Varnish, but we won’t be using one in Symfony HttpCache (if I understand #234 correctly).

In the current solution, `getTagsHeaderValue` (e.g.) is duplicated in `CacheInvalidator` and `TagHandler`.

src/ProxyClient/AbstractProxyClient.php
ddeboer: Good point.

ddeboer: Good point.

dbu:
@ddeboer that is why i hoped for your input. it started to look here like having a TagHandler is a bad idea after all, because things would be easier in CacheInvalidator when talking directly to the proxy client. the proxy client actually needs to know about tagging, if we want to abstract from using ban and leave that to the client.

so maybe we should just drop the TagHandler again? we introduced it in one of the 1.x versions and deprecated tag methods on CacheInvalidator - but it seems to me now that we where going the wrong way there... 

i wonder if/how the proxy client should add tags to responses. should we have a registerTags method that gets the response object passed but that may also do other stuff than adding a header? @lolautruche how would you pass the tags to use on a response to the symfony HttpCache? in a header as well? or are there other ways?

andrerom:
Read up a bit on the code now again.

One option *which might make sense:
- we remove `TagHandler->invalidateTags()`
- and instead of invalidator we inject ProxyClient
- thus we can also remove `getTagsHeaderValue()` and `getTagsHeaderName()` from invalidator

This way `TagHandler` is a higher level api for keeping track of tags and output tags in response, but all invalidation (and only invalidation) is kept on `CacheInvalidator`.

dbu:
so invalidateTags would happen on the CacheInvalidator, while the header
things happen on TagHandler? from the point of view of invalidating,
this is logic, but from the point of view of handling tags, its
confusing that its different services.

what if we had invalidateTags on TagHandler, but directly forward it to
the ProxyClient? there is no need to go through the CacheInvalidator for
this i guess.

so its purely a question of what is less confusing. hm, given that
invalidateTags is expecting an array of tags to invalidate and is not
using the tags previously registered in the TagHandler with addTag, it
could make sense to have the invalidation method completely in
CacheInvalidator. and TagHandler is purely a tool to help tagging
responses and has nothing to do with invalidating the tags.

andrerom:
> so its purely a question of what is less confusing. hm, given that
invalidateTags is expecting an array of tags to invalidate and is not
using the tags previously registered in the TagHandler with addTag, it
could make sense to have the invalidation method completely in
CacheInvalidator. and TagHandler is purely a tool to help tagging
responses and has nothing to do with invalidating the tags.

:+1:

dbu:
ping @ddeboer @lolautruche

dbu:
ping @ddeboer @lolautruche

lolautruche:
Fine by me

ddeboer:
Agreed! @andrerom Could you update the PR? It's about time we merge this. ;)

andrerom:
Yes, I'll try to make time, was waiting for approval ;)

dbu:
cool! sorry for the long wait.

what exactly did we agree on right now? to keep a special class to tag responses but have tag invalidation in the CacheInvalidator?

do we want to call that thing ResponseTagger then? the word Handler is in the namespace already and not adding much imho. ResponseTagHandler is very long. TagHandler is kind of lying.

andrerom:
this:

> given that
invalidateTags is expecting an array of tags to invalidate and is not
using the tags previously registered in the TagHandler with addTag, it
could make sense to have the invalidation method completely in
CacheInvalidator. and TagHandler is purely a tool to help tagging
responses and has nothing to do with invalidating the tags.

I'll ping for specifics once I have updated PR to settle the details.

dbu:
:+1:

andrerom:
Updated, @dbu open question is how to handle proxy clients that don't support tagging, right now it in practice behaves like before, just instead of exception PHP will fatal instead because of wrong type.

tests/Unit/ProxyClient/VarnishTest.php
```diff
@@ -99,6 +99,43 @@ public function testBanPathEmptyHost()
...
+
+        $requests = $this->getRequests();
+
+        $this->assertCount(1, $requests);
+        $this->assertEquals('BAN', $requests[0]->getMethod());
+
+        $this->assertEquals('(post\-1|post\-type\-3)(,.+)?$', $requests[0]->getHeaderLine('X-Cache-Tags'));
+        $this->assertEquals('fos.lo', $requests[0]->getHeaderLine('Host'));
+
+        // That default BANs is taken into account, might not be expected from user POV when invalidating specific tags?
```
andrerom: @ddeboer question  here clarified a bit, but still open.

tests/Unit/ProxyClient/VarnishTest.php
andrerom: updated: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/237#discussion_r45044351

src/Handler/TagHandler.php
```diff
@@ -11,26 +11,21 @@
...
+use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface as ProxyClientInvalidationTagsInterface;
 
 /**
  * Handler for cache tagging.
  *
  * @author David de Boer <david@driebit.nl>
  * @author David Buchmann <mail@davidbu.ch>
+ * @author André Rømcke <ar@ez.no>
  */
 class TagHandler
```
dbu: what do you think about calling this ResponseTagger?

src/Handler/TagHandler.php
```diff
@@ -11,26 +11,21 @@
 
 namespace FOS\HttpCache\Handler;
 
-use FOS\HttpCache\CacheInvalidator;
-use FOS\HttpCache\Exception\UnsupportedProxyOperationException;
+use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface as ProxyClientInvalidationTagsInterface;
```
dbu: why do we need to alias this?

src/ProxyClient/Invalidation/TagsInterface.php
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Get the value for the HTTP tag header.
+     *
+     * This concatenates all tags and ensures correct encoding.
+     *
+     * @param array $tags
+     *
+     * @return string
+     */
+    public function getTagsHeaderValue(array $tags);
```
dbu: is this not too specific? for HttpCache, this could be something else than a string, or not?

src/ProxyClient/Varnish.php
```diff
@@ -89,6 +97,35 @@ public function setDefaultBanHeader($name, $value)
...
+     * Get the HTTP header name that will hold cache tags.
+     *
+     * @return string
+     */
+    public function getTagsHeaderName()
+    {
+        return $this->tagsHeader;
+    }
+
+
```
dbu: extra blank line, please remove

src/Handler/TagHandler.php
```diff
@@ -73,7 +61,11 @@ public function getTagsHeaderName()
      */
     public function getTagsHeaderValue()
     {
-        return implode(',', array_unique($this->escapeTags($this->tags)));
+        if (empty($this->tags)) {
+            return '';
+        }
+
+        return $this->client->getTagsHeaderValue($this->tags);
```
dbu: do we need to forward this call to the client? but if we do, i think we should remove the check for no tags and let the client handle this. (currently it should work fine, implode of an empty array returns an empty string)

src/Handler/TagHandler.php
```diff
@@ -83,7 +75,7 @@ public function getTagsHeaderValue()
      */
     public function hasTags()
     {
-        return 0 < count($this->tags);
+        return !empty($this->tags);
```
dbu: why this change? performance?

src/CacheInvalidator.php
```diff
@@ -198,6 +206,26 @@ public function invalidate(array $headers)
...
+     *
+     * @return $this
+     */
+    public function invalidateTags(array $tags)
+    {
+        if (!$this->cache instanceof TagsInterface) {
+            throw UnsupportedProxyOperationException::cacheDoesNotImplement('Tags');
+        }
+        $this->cache->invalidateTags($tags);
+        return $this;
```
dbu: please have an empty line before the return statement for readability.

tests/Unit/ProxyClient/VarnishTest.php
dbu: sorry, i don't understand what exactly you mean. during the tag invalidation, do we send other ban requests as well? or is the thing that we only BAN if e.g. the domain works?

tests/Unit/ProxyClient/VarnishTest.php
```diff
@@ -142,7 +179,7 @@ protected function setUp()
     }
 
     /**
-     * @return array|RequestInterface[]
+     * @return array|\Psr\Http\Message\RequestInterface[]
```
dbu: lets use a `use` for this

dbu:
very cool! i commented a couple of details, but overall i am happy with this!

src/Handler/TagHandler.php
andrerom: Would make sense

src/Handler/TagHandler.php
andrerom: no need, will fix

src/ProxyClient/Invalidation/TagsInterface.php
andrerom: suggestions?

src/Handler/TagHandler.php
andrerom: the logic of moving it to client was that client is only one knowing how header should look like to be able to work. But I have no problems standardizing header name and moving that back to ResponseTagger.

src/Handler/TagHandler.php
andrerom: consistency with `if (empty($this->tags)) {`, and perf, but neglectable, so can remove the change.

tests/Unit/ProxyClient/VarnishTest.php
andrerom: see assertions under, there are bans going on that are not specified among the tags, this is a FOSHttpCache BAN feature, and my question is if it still makes sense on `invalidateTags ` or not.

src/Handler/TagHandler.php
andrerom: > but if we do, i think we should remove the check for no tags and let the client handle this. 

fixed in  51ce1cc

src/ResponseTagger.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Get the HTTP header name that will hold cache tags.
+     *
+     * @return string
+     */
+    public function getTagsHeaderName()
+    {
+        return $this->client->getTagsHeaderName();
```
andrerom: *Continuing discussion from https://github.com/FriendsOfSymfony/FOSHttpCache/pull/237#discussion_r45050050*

>> Do we need to forward this call to the client?
> the logic of moving it to client was that only client knows how it needs to be escaped for specifics on proxy

However maybe this is not needed, constraints I know about:
- For Sf I was kind of thinking the header name needs to be hard coded, but I guess it does not have to do that
- For Nginx I have yet to look into which plugins would make sense to enable this, and which constraints they will enforce

So maybe this can be fully handled in ResponseTagger, `escapeTags ` is rather generic anyway.
However be aware both Proxy and ResponseTagger will need to know about header then, but not a blocker, can use same param in bundle.

andrerom:
@dbu updated ;)

src/ProxyClient/Invalidation/TagsInterface.php
dbu: my question is whether saying that the tags are set in a header is too specific. but actually if they are not, we would need to move the whole tag response logic in here as well. lets keep this as it is now.

src/ResponseTagger.php
dbu: lets keep it like this. we want to have a default header name and if there are several places, users that do not use the bundle have a dangerous live.

tests/Unit/ProxyClient/VarnishTest.php
dbu: i think at least for varnish, the configuration is assuming all those headers are set, so i think it is correct to assume them. also, i could have a multi domain setup where i want to only ban per hostname. maybe ids are only unique per domain, not globally.

dbu:
i think we are all good now with the code!

can you please add a note to the changelog about the rename and moving around of responsibilities, and update the .rst documentation?

andrerom:
First pass on doc done.

doc/response-tagging.rst
```diff
@@ -75,21 +74,21 @@ Only ``/one`` will stay in the cache.
 
 .. note::
 
-    Don't forget to call :ref:`flush <flush>` on the ``CacheInvalidator`` to commit the invalidations
-    to the caching proxy.
-
-.. _custom_tags_header:
```
dbu: removing this leads to "/home/travis/build/FriendsOfSymfony/FOSHttpCache/doc/varnish-configuration.rst:180: WARNING: undefined label: custom_tags_header (if the link has no caption the label must precede a section header)"

please re-add it right before the "Custom Tags Header" subtitle.

doc/cache-invalidator.rst
```diff
@@ -82,6 +82,32 @@ Refresh a URL with added header(s)::
...
+
+.. note::
+
+    Make sure to :doc:`configure your proxy <proxy-configuration>` for tagging first,
+    in the case of Varnish this is powered by banning.
+
+By taking advantage of :doc:`response tagging <response-tagging>`, invalidation
+all URLs that has been tagged with a given tag becomes possible.
+
+
```
dbu: extra blank line

doc/cache-invalidator.rst
```diff
@@ -82,6 +82,32 @@ Refresh a URL with added header(s)::
...
+Invalidating by Tags
+--------------------
+
+.. note::
+
+    Make sure to :doc:`configure your proxy <proxy-configuration>` for tagging first,
+    in the case of Varnish this is powered by banning.
+
+By taking advantage of :doc:`response tagging <response-tagging>`, invalidation
+all URLs that has been tagged with a given tag becomes possible.
```
dbu: maybe something like this?

```
When you are using :doc:`response tagging <response-tagging>`, you can invalidate all requests that where tagged with a specific label.
```

doc/cache-invalidator.rst
```diff
@@ -82,6 +82,32 @@ Refresh a URL with added header(s)::
...
+
+    $cacheInvalidator->invalidateTags(['blog-post-44'])->flush();
+
+See below for the :ref:`flush() <flush>` method.
+
+Invalidate several tags::
+
+    $cacheInvalidator->invalidateTags(['type-65', 'location-3'])
+    ->flush()
+    ;
```
dbu: can we layout this as:

```php
$cacheInvalidator
    ->invalidateTags(['type-65', 'location-3'])
   ->flush()
;
```

doc/response-tagging.rst
```diff
@@ -1,17 +1,17 @@
-Extra Invalidation Handlers
-===========================
+Response Tagging
+================
 
-This library provides decorators that build on top of the ``CacheInvalidator``
-to simplify common operations.
+This library provides a service to handle tagging of responses ``ResponseTagger``
```
dbu: s/service/helper/ ?

doc/response-tagging.rst
```diff
@@ -1,17 +1,17 @@
...
+to simplify use of tags.
 
 .. _tags:
 
-Tag Handler
------------
+Response Tagger
+---------------
```
dbu: i don't think we need two titles for response tagging. this document was about all handlers, but there never was more than 1 so i agree it makes more sense this way. but one title and introduction is enough

doc/response-tagging.rst
```diff
@@ -1,17 +1,17 @@
...
 
-The tag handler helps you to mark responses with tags that you can later use to
-invalidate all cache entries with that tag. Tag invalidation works only with a
-``CacheInvalidator`` that supports ``CacheInvalidator::INVALIDATE``.
+The ResponseTagger helps you to mark responses with tags that you can later use to
+invalidate all cache entries with that tag. Response Tagging and tag invalidation
+require a proxy client that implements ``TagsInterface``.
```
dbu: please put double backticks around the class name ResponseTagger

doc/response-tagging.rst
```diff
@@ -1,17 +1,17 @@
...
 
-Tag Handler
------------
+Response Tagger
+---------------
 
-The tag handler helps you to mark responses with tags that you can later use to
-invalidate all cache entries with that tag. Tag invalidation works only with a
-``CacheInvalidator`` that supports ``CacheInvalidator::INVALIDATE``.
+The ResponseTagger helps you to mark responses with tags that you can later use to
```
dbu: please put double backticks around the class name ResponseTagger

doc/response-tagging.rst
dbu: i think this introduction is enough, the first line after the main title is saying kind of the same

doc/response-tagging.rst
```diff
@@ -20,29 +20,28 @@ Setup
 
     Make sure to :doc:`configure your proxy <proxy-configuration>` for tagging first.
 
-The tag handler is a decorator around the ``CacheInvalidator``. After
-:doc:`creating the invalidator <cache-invalidator>` with a proxy client
-that implements the ``BanInterface``, instantiate the ``TagHandler``::
+The response tagger is a decorator around a proxy client that implements
+the ``TagsInterface``, handling tags for the current response::
```
dbu: "handling adding tags to responses"?

doc/response-tagging.rst
```diff
@@ -20,29 +20,28 @@ Setup
...
+    // $proxyClient already created, implementing FOS\HttpCache\ProxyClient\Invalidation\TagsInterface
+    $responseTagger = new ResponseTagger($proxyClient);
 
 Usage
 ~~~~~
 
 With tags you can group related representations so it becomes easier to
 invalidate them. You will have to make sure your web application adds the
-correct tags on all responses. You can add tags to the handler using::
+correct tags on all responses. You can add tags to the response using::
```
dbu: this is not true, the tags are added into the handler (maybe call it ResponseTagger here). but to add them to the response, you need to use getTagsHeaderName|Value. or with #230 call tagResponse to tag a PSR-7 response. but the response is not tagged simply by calling this.

this is different in the symfony bundle, which we might add a note about somewhere here.

doc/response-tagging.rst
```diff
@@ -75,21 +74,21 @@ Only ``/one`` will stay in the cache.
 
 .. note::
 
-    Don't forget to call :ref:`flush <flush>` on the ``CacheInvalidator`` to commit the invalidations
-    to the caching proxy.
```
dbu: lets add the call to flush in the code example of invalidateTags

doc/response-tagging.rst
```diff
@@ -75,21 +74,21 @@ Only ``/one`` will stay in the cache.
...
-
-.. _custom_tags_header:
+    For further reading on tag invalidation see :doc:`cache-invalidator page <cache-invalidator>`.
+    For changing the cache header, :doc:`configure your proxy <proxy-clients>`
 
 Custom Tags Header
 ~~~~~~~~~~~~~~~~~~
 
+// TOOO: Move or change to reflect that this depends on Proxy Client.
```
dbu: lets move this to the client, yes. this section is not relevant here anymore.

doc/proxy-clients.rst
```diff
@@ -91,6 +91,11 @@ include that port in the base URL::
 
     $varnish = new Varnish($servers, 'my-cool-app.com:8080');
 
+Another optional parameter available on Varnish client is ``tagsHeader``, which allows you to
+change the default HTTP header used for tags, ``X-Cache-Tags``::
```
dbu: lets say "tagging" instead of "tags", and link that to the response tagger documentation.

doc/response-tagging.rst
andrerom: > this is different in the symfony bundle, which we might add a note about somewhere here.

this is mentioned in usage part already

dbu: ah, perfect. but lets talk about the ResponseTagger here, not about adding to the response, because that could lead to assume that the tags would magically happen to be in the resposne without doing anything :-)

doc/cache-invalidator.rst
andrerom: assuming you meant `requests => responses`

doc/response-tagging.rst
andrerom: I changed it to `_varnish_custom_tags_header` instead to refer to the right section.

doc/cache-invalidator.rst
dbu: yes i did. sorry.

andrerom:
@dbu second pass

doc/response-tagging.rst
```diff
@@ -0,0 +1,76 @@
+Response Tagging
+================
+
+This library provides a helper to handle tagging of responses ``ResponseTagger``
+to simplify use of tags. The ``ResponseTagger`` helps you keep track tags for a response, which can be
```
dbu: i think the first line can simply be removed, it does not say anything more than the second line.

doc/response-tagging.rst
```diff
@@ -0,0 +1,76 @@
...
+
+    header(sprintf('%s: %s'),
+        $responseTagger->getTagsHeaderName(),
+        $responseTagger->getTagsHeaderValue()
+    );
+
+.. tip::
+
+    If you are using Symfony with the FOSHttpCacheBundle_, the tag header is
+    set automatically. You also have `additional methods of defining tags`_ with
```
dbu: ... the tags added to `ResponseTagger` are added to the response automatically.

dbu:
cool! can you adress these last two details and squash commits? then we can merge this.

andrerom:
done

andrerom:
Failing on doc, as I don't know anything about rst, how is it supposed to work?

doc/proxy-clients.rst
```diff
@@ -91,6 +91,15 @@ include that port in the base URL::
 
     $varnish = new Varnish($servers, 'my-cool-app.com:8080');
 
+-.. _varnish_custom_tags_header:
```
dbu: you have a `-` too much. the line should start with `..`

andrerom: Thanks, copy past fail, hopefully fixed in bc36103

dbu:
great job, thanks a lot for your patience!

andrerom:
@dbu Likewise :)

ddeboer:
Thanks @andrerom and @dbu! :+1:

Issue #210, opened 2015-06-06
Documentation enhancements


ddeboer:
Some ideas for improving the docs in preparation for 2.0.

* [ ] 1. Add FAQ, e.g. 
    1. ‘Q. Why is nothing purged? A. Check that you are calling `flush()`’.
    2. ‘Q. Which invalidation methods are supported? A. Check table.’
    3. Checking Varnish logs while testing (`varnishlog -N /tmp/foshttpcache-varnish/_.vsm`).
    4. Varnish how to: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/227#issuecomment-114900370
* [x] 2. Rename ‘Caching Proxy Configuration’ to ‘Proxy Server Configuration’ (or `Caching Server Configuration`). This creates a clearer contrast with the next chapter, ‘Caching Proxy Clients’ (and perhaps rename that to ‘Proxy Clients’).
* [x] 3. ‘Caching Proxy Clients’: emphasise that only 1 client needs to be created.
* [ ] 4. Move the introductory chapter on cache invalidation to the end of the docs?
* [x] 5. Remove version constraint from `$ composer install` (and let Composer find the right one).
* [ ] 6. Add a glossary: caching proxy (server), ….

dbu:
+1 for the renaming. yeah lets just talk about a proxy and a proxy client. i think we can do most of that even for version 1.

only point i am not entirely sure is moving 4 to the end. i think the introduction does help people not familiar with it. and we have the install and config and overview chapter before that for the impatient. but i am not strictly against it, lets discuss in amsterdam ;-)

ddeboer:
* Move HTTP adapter customization to the bottom of the page

purplefish32:
How about : "How to purge/ban chache for a single user"

dbu:
@purplefish32 is that an entry for the FAQ? what exactly do you mean with "for a single user"?

purplefish32:
Yes I ment that as a FAQ entry, sorry that was not specific enough.
Purging/Baning a spécific users cached version of a page is hard to understand.

Example : A user is promoted to admin in a platform, he get lots of new shiny buttons in the menu. I am guessing you have to bust all the pages using a Vary on the PHPSESSID cookie. I dont know if this is the correct way to handle the problem or not, or if this is typicly the case for ESI.

dbu:
[...]

PR #240, opened 2015-08-15
do not specify version in composer install

dbu:
point 5 in #210

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -9,7 +9,7 @@ Composer_:
 
 .. code-block:: bash
 
-    $ composer require friendsofsymfony/http-cache:~2.0
+    $ composer require friendsofsymfony/http-cache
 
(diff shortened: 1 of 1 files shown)


doc/installation.rst
```diff
@@ -9,13 +9,15 @@ Composer_:
...
 .. note::
 
-    This library follows `Semantic Versioning`_. Because constraint ``~1.0``
-    will only increment the minor and patch numbers, it will not introduce BC
-    breaks.
+    This library follows `Semantic Versioning`_. Except for major versions, we
+    aim to not introduce BC breaks in new releases. You should still test your
+    application after upgrading though. What is a bug fix for somebody could
+    break something for others when they where (probably unawares) relying on
+    that bug.
```
dbu: this is kind of stating the obvious. we can also remove everything after "not to introduce BC breaks in new releases" if you prefer

PR #246, opened 2015-10-24
Try to fix HHVM build

### LINKED ISSUE #245: hhvm build broken
The master build that was green a month ago now fails with hhvm https://travis-ci.org/FriendsOfSymfony/FriendsOfSymfony/FOSHttpCache/jobs/77916788

It looks like something very basic broke on travis.

ddeboer:
Fix #245.

Wasn’t able to reproduce the build failures on my local HHVM 3.9. Travis currently installs 3.6, so trying a newer version.

### DIFF
--- a/tests/ci/travis-ci-apache-hhvm
+++ b/tests/ci/travis-ci-apache-hhvm
@@ -20,7 +20,7 @@ Listen 8080
       Alias /hhvm /hhvm
       Action hhvm-php-extension /hhvm virtual
 
-      FastCgiExternalServer /hhvm -host 127.0.0.1:9000 -pass-header Authorization -idle-timeout 300
+      FastCgiExternalServer /hhvm -host 127.0.0.1:9090 -pass-header Authorization -idle-timeout 300
   </IfModule>
 
 </VirtualHost>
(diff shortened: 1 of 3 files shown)


dbu:
did you try curl an url that should work to see the page hhvm outputs?

dbu:
okay, i think we are getting there. reverted some debug-only stuff. if its good i will squash the commits to get rid of this tragic history :-)

tests/ci/travis-ci-apache-hhvm
```diff
@@ -20,7 +20,7 @@ Listen 8080
       Alias /hhvm /hhvm
       Action hhvm-php-extension /hhvm virtual
 
-      FastCgiExternalServer /hhvm -host 127.0.0.1:9000 -pass-header Authorization -idle-timeout 300
+      FastCgiExternalServer /hhvm -host 127.0.0.1:9090 -pass-header Authorization -idle-timeout 300
```
dbu: looks like something started to pirate port 9000 from us. like this we are sure we use the hhvm we think we use.

ddeboer:
Good find! Glad we have some more HHVM debugging now as well: it will probably come in handy in the future. :wink:

PR #252, opened 2015-11-23
Add tagResponse method

### LINKED ISSUE #230: Add PSR-7 TagHandler::tagResponse()

ddeboer:
Fix #230.

### DIFF
--- a/src/ProxyClient/Invalidation/TagsInterface.php
+++ b/src/ProxyClient/Invalidation/TagsInterface.php
@@ -29,13 +29,11 @@ interface TagsInterface extends ProxyClientInterface
     public function invalidateTags(array $tags);
 
     /**
-     * Get the value for the HTTP tag header.
-     *
-     * This concatenates all tags and ensures correct encoding.
+     * Get escaped tags
      *
      * @param array $tags
      *
-     * @return string
+     * @return array
      */
     public function getTagsHeaderValue(array $tags);
 
--- a/src/ProxyClient/Varnish.php
+++ b/src/ProxyClient/Varnish.php
@@ -84,7 +84,7 @@ class Varnish extends AbstractProxyClient implements BanInterface, PurgeInterfac
      */
     public function getTagsHeaderValue(array $tags)
     {
-        return implode(',', array_unique($this->escapeTags($tags)));
+        return array_unique($this->escapeTags($tags));
     }
 
     /**
--- a/src/ResponseTagger.php
+++ b/src/ResponseTagger.php
@@ -11,8 +11,9 @@
 
 namespace FOS\HttpCache;
 
-use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
 use FOS\HttpCache\Exception\InvalidTagException;
+use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
+use Psr\Http\Message\ResponseInterface;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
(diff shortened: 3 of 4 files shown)


src/ProxyClient/Invalidation/TagsInterface.php
```diff
@@ -29,13 +29,11 @@
...
-     *
-     * This concatenates all tags and ensures correct encoding.
+     * Get escaped tags
      *
      * @param array $tags
      *
-     * @return string
+     * @return array
      */
     public function getTagsHeaderValue(array $tags);
```
ddeboer: When using PSR-7, I think it has become the responsibility of the HTTP messaging library chosen to format the header correctly. We may want to rename this method to `getEscapedTags()`.

src/ResponseTagger.php
```diff
@@ -89,4 +90,22 @@ public function addTags(array $tags)
...
+    /**
+     * Set tags on a response
+     *
+     * @param ResponseInterface $response Original response
+     * @param bool              $replace  Whether to replace the current tags
+     *                                    on the response
+     *
+     * @return ResponseInterface          Tagged response
+     */
+    public function tagResponse(ResponseInterface $response, $replace = false)
```
ddeboer: Remove this typehint (see #230) or add another method `tagSymfonyResponse` to the bundle’s ResponseTagger (previously TagHandler)?

src/ProxyClient/Invalidation/TagsInterface.php
dbu: if we return a tag array, what would this method do at all? can you check how escaping is handled? i guess one thing that a PSR-7 library won't care about is `,` in a header - but if there are any `,` in a tag, we need to escape them.

dbu: but if PSR-7 does some escaping, we don't want to escape things it would escape ourselves, or we risk double escaping.

src/ResponseTagger.php
dbu: lets call this one `tagPsrResponse` and add a tagSymfonyResponse. the implementation is different, maybe even the escaping is different.

src/ProxyClient/Varnish.php
```diff
@@ -109,7 +109,7 @@ public function invalidateTags(array $tags)
      */
     public function getTagsHeaderValue(array $tags)
     {
-        return implode(',', array_unique($this->escapeTags($tags)));
+        return array_unique($this->escapeTags($tags));
```
dbu: but would you expect the psr to accept an array as header value and combine the array with commas? i would not expect that. and you keep calling escapeTags so i don't see what is changing here.

dbu: maybe we should add a functional test to see if the integration works?

ddeboer: The difference is that we don't join the tags into one header ourselves, as PSR implementations already be do that (see ResponseInterface).

dbu: hm, but is comma separating defined? the way we invalidate e.g. in
varnish, it must be comma. it seems kind of strange that psr response
would handle this, as its not a question on the HTTP layer but on
application layer. (and comma is just our choice, we could have used any
character as long as we escape it when it occurs in tags)

ddeboer: A comma separated list is the standard HTTP way of having multiple values for a header. That applies here: we have multiple tags for the tag header.

dbu: fair enough.

src/ResponseTagger.php
dbu: lets rename this method and then merge. we still might refactor things for #265

dbu:
@ddeboer do you have some time this week to wrap this up?

ddeboer:
Will do so on Friday.

dbu:
i did a rebase to get rid of outdated php-http dependencies.

src/ResponseTagger.php
ddeboer: Actually, I prefer to keep the method name `tagResponse()`. As the library is now built around PSR-7, we can assume that we’re talking about a PSR-7 response here. 

In the bundle, then, we may want to rename its `tagResponse(SymfonyResponse)` method to `tagSymfonyResponse()`, but that requires no changes here.

@dbu If you agree, this PR is ready to merge right away.

ddeboer:
Thanks. See my comment above.

src/ResponseTagger.php
dbu: makes sense.

dbu:
yay!

PR #253, opened 2015-11-27
Tag Subscriber

### LINKED ISSUE #251: support tag invalidation with Symfony HttpCache
#237 separated tagging from BAN. we can now support tag invalidation with the symfony cache without needing full BAN support (#148).

/cc @andrerom

dantleech:
Would fix #251

This PR adds a framework for handling tagging and tag invalidation and an implementation for the Symfony HTTP cache.

httpcache

## Framework

Proxy clients may accept a `Tag\InvalidatorInterface` which can be used to fulfil the contract imposed by the `TagInterface`.

This PR provides 3 methods of invalidating tags; *response*, *remote* and *direct*:

- *response* method adds the tags-for-invalidation to the response;
- *remote* method sends a sepearate HTTP request to invalidate the tags (can be used with multi-server setup);
- *direct* uses the same instance of the TagManager as the HttpCache to directly invalidate tags.

Note that only the `remote` invalidator would work on multi-server/installation setups.

## Symfony HTTP Cache Implemnetation

This PR adds a `TagSubscriber` event subscriber to the event-dispatching Symfony `HttpCache`. This subscriber is
responsible for associating tags with responses and for invalidating tags (either from a request or from a response).

The subcsriber is dependent upon the `Tag\Manager\SymfonyCacheManager` to
perform the actual invalidation and association of tags.

## Tag Storage

An interface is provided for "tag storage" implementations. These implementations can be used by any system which needs
to associate tags with cache-entry-identifiers (i.e. URLs, hashes, whatever).

Currenltly a `DoctrineCache` implementation is provided which uses the Doctrine
Cache library - which allows the use of many different storage backends.

## Not Currently Implemented

- Tag content expiration: If all the contents that are associated with a given tag have expired, then the tag could be removed. We would need to serialize the expiry date of the content along with the content hash.

## Example

```php
$httpCache = // ...
$store = new Store($cacheDir);
$tagStorage = new DoctrineCache(new PhpFileCache($cacheDir . '/tags'));
$tagManager = new SymfonyCacheManager($tagStorage, $store);
$httpCache->addSubscriber(new TagSubscriber($tagManager));
```


src/SymfonyCache/Tag/ManagerInterface.php
```diff
@@ -0,0 +1,32 @@
...
+
+interface ManagerInterface
+{
+    /**
+     * Return the concrete cache paths for the given tag.
+     *
+     * @param string $tag
+     * @return string[]
+     */
+    public function getPathsForTag($tag);
```
dantleech: This is a utility method really for testing. Could be removed from the interface.

src/SymfonyCache/Tag/NullManager.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace FOS\HttpCache\SymfonyCache\Tag;
+
+class NullManager implements ManagerInterface
```
dantleech: Use this class in dev env.

src/SymfonyCache/Tag/RemotePurger.php
```diff
@@ -0,0 +1,42 @@
...
+                'method'  => TagSubscriber::INVALIDATE,
+                'header' => sprintf('%s: %s', TaggedCache::HEADER_TAGS, json_encode($tags))
+            )
+        );
+        $context = stream_context_create($request);
+
+        foreach ($this->purgeUrls as $purgeUrl) {
+            $contents = file_get_contents($purgeUrl, false, $context);
+
+            $return = json_decode($contents, true);
```
dantleech: I am returning status information in JSON, seemed easiest.

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,155 @@
...
+     * Name for HTTP header containing the tags (for both invalidation and
+     * initial tagging).
+     */
+    const HEADER_TAGS = 'X-TaggedCache-Tags';
+
+    /**
+     * Header which should contain the content digest produced by the Symfony
+     * HTTP cache.
+     */
+    const HEADER_CONTENT_DIGEST = 'X-Content-Digest';
```
dantleech: I initially coded these as options, as with the `UserContextSubscriber`, but it makes it awkward when you need these values from somewhere else, and would indicate the need for a `Configuration` object or something..

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * @var array
+     */
+    private $options = array();
```
dantleech: Not actually used.

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,150 @@
...
+     * Check to see if the request is an invalidation request, if so
+     * handle the invalidation.
+     *
+     * @param CacheEvent $event
+     */
+    public function preHandle(CacheEvent $event)
+    {
+        $request = $event->getRequest();
+
+        if (self::METHOD_INVALIDATE!== $request->getMethod()) {
```
dantleech: Formatting -- note that this method does not work for 2 reasons with the built-in PHP webserver:

1. It does not recognize custom HTTP methods.
2. It is single threaded, making a request on the webserver when it is serving will cause it to lockup.

Should document this.

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,150 @@
...
+        return new JsonResponse(array(
+            'Status' => 'PURGED',
+            'NbCacheEntries' => $nbCacheEntries
+        ));
+    }
+
+    private function handleTags(Response $response)
+    {
+        if (!$response->headers->has(self::HEADER_CONTENT_DIGEST)) {
+            throw new \RuntimeException(sprintf(
```
dantleech: Not really sure under which cicumstances the content digest header is sent -- we need it in order to link to the cache, and it seems to always be sent when the HTTP cache stores the page.

tests/Unit/SymfonyCache/TagSubscriberTest.php
```diff
@@ -0,0 +1,145 @@
...
+    private $tagManager;
+
+    /**
+     * @var CacheEvent
+     */
+    private $event;
+
+    public function setUp()
+    {
+        $this->tagManager = $this->prophesize('FOS\HttpCache\SymfonyCache\Tag\ManagerInterface');
```
dantleech: Using prophecy instead of PHPUnit Mock Object, hope thats ok.

composer.json
```diff
@@ -24,6 +24,7 @@
         "php": ">=5.4.8",
         "symfony/event-dispatcher": "~2.3",
         "symfony/options-resolver": "~2.3",
+        "symfony/filesystem": "~2.3",
```
dantleech: Should we depend on this or reimplement the code locally? The filesystem component provides a more resilient `symlink` method and recursive delete.

src/SymfonyCache/Tag/RemotePurger.php
ddeboer: Why not use the php-http client here instead of `file_get_contents`?

composer.json
ddeboer: Let’s not reinvent the wheel and just use symfony/filesystem. The only problem I have with this is that it is an *optional* dependency. @dbu Should we move symfony/filesystem to `require-dev` and add a note in the docs that users should install the dependency when they want to use tagging in Symfony Cache?

ddeboer:
Pretty cool, @dantleech! I think the `ManagerInterface` makes sense: it leaves open the option (as discussed in #251) to have different ways of invalidating tags with the Symfony Cache.

src/SymfonyCache/Tag/RemotePurger.php
dantleech: Can do - originally this code didn't have that dependency.

tests/Unit/SymfonyCache/TagSubscriberTest.php
ddeboer: Would you be okay with changing this to Mockery? That’s what we now use in this library’s test suites. It may be a bit confusing to have different ways of mocking.

dantleech: Yeah sure.

src/SymfonyCache/Tag/ManagerInterface.php
```diff
@@ -0,0 +1,31 @@
...
+    public function invalidateTags(array $tags);
+
+    /**
+     * Create a new tag for the given content digest.
+     *
+     * @param string $tag
+     * @param string $contentDigest
+     * @return void
+     */
+    public function createTag($tag, $contentDigest);
```
dantleech: Rename this -- we are associating a content digest with a tag which may already exist. Perhaps `tagContentDigest`

src/ProxyClient/Symfony.php
```diff
@@ -54,6 +58,7 @@ public function __construct(
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
             'purge_method' => PurgeSubscriber::DEFAULT_PURGE_METHOD,
+            'tags_invalidator' => new NullManager()
```
dantleech: Have made the tags invalidator an option on the `Symfony` proxy.

src/ProxyClient/Symfony.php
```diff
@@ -79,4 +84,34 @@ public function refresh($url, array $headers = [])
 
         return $this;
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function invalidateTags(array $tags)
+    {
+        $invalidationRequest = $this->options['tags_invalidator']->invalidateTags($tags);
```
dantleech: The tags invalidator optionally returns an `InvalidationRequest`. Invalidation can be performed directly upon the (tag) `ManagerInterface` instance, or on a dedicated `InvalidatorInterface` instance - i.e. the `RemoteInvalidator`.

src/SymfonyCache/Tag/RemoteInvalidator.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace FOS\HttpCache\SymfonyCache\Tag;
+
+use FOS\HttpCache\ProxyClient\Request\InvalidationRequest;
+use FOS\HttpCache\ProxyClient\Symfony;
+use FOS\HttpCache\SymfonyCache\TagSubscriber;
+
+class RemoteInvalidator implements InvalidatorInterface
```
dantleech: document

dantleech:
Integrated with the Symfony proxy client. Everything should basically be there now, excepting the documentation. Will next try and test it on my website.

src/ProxyClient/Symfony.php
```diff
@@ -23,9 +27,10 @@
  * @author David de Boer <david@driebit.nl>
  * @author David Buchmann <mail@davidbu.ch>
  */
-class Symfony extends AbstractProxyClient implements PurgeInterface, RefreshInterface
+class Symfony extends AbstractProxyClient implements PurgeInterface, RefreshInterface, TagsInterface
 {
     const HTTP_METHOD_REFRESH = 'GET';
+    const HTTP_METHOD_INVALIDATE = 'INVALIDATE';
```
dantleech: Don't think this is required, it already exists somewhere.

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    private function tagResponse(Response $response)
+    {
+        $contentDigest = $this->getContentDigestFromHeaders($response->headers);
+        $tags = $this->getTagsFromHeaders($response->headers);
+        $expiry = $this->getExpiryFromResponse($response);
+        $this->manager->tagCacheId($tags, $contentDigest, $expiry);
```
dantleech: `s/lifetime/expiry`

src/Tag/Manager/Symfony.php
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        // remove the tag directory
+        $this->tagStorage->removeTags($tags);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function tagCacheId(array $tags, $contentDigest, $lifetime)
```
dantleech: We discussed this as `tagResponse` and accepting the `Response` object, however it makes more sense to deal with extracting things from the Response in the TagSubscriber, which is coupled to the HTTP cache anyway. However we retain this method so as not to ensure that both invalidateTags and tagCacheID operate on the same tag storage.

src/Tag/Storage/DoctrineCache.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * TODO: The lifetime applies not to the tag but to the identifier - we
+     *       need to serialize the lifetime with the cache entry and store tags
+     *       indefintely.
+     *
+     * {@inheritDoc}
+     */
+    public function tagCacheId(array $tags, $identifier, $lifetime = null)
```
dantleech: As with the TODO, we will need to serialize the expiry time in addition to the cache ID and then we can ignore expired cache entries when invalidating.

This does mean however that we will need to prune our tag tree somehow from time to time.

src/Tag/StorageInterface.php
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * The identifier can be any scalar value which can be associated with a
+     * unique HTTP cache entry.
+     *
+     * @param string[] $tags
+     * @param mixed $identifier
+     * @param integer $expiry
+     * @return void
+     */
+    public function tagCacheId(array $tags, $identifier, $expiry = null);
```
dantleech: `s/expiry/lifetime`

src/Test/SymfonyTest.php
```diff
@@ -108,14 +108,21 @@ protected function getProxy()
      *
      * @return Symfony
      */
-    protected function getProxyClient()
+    protected function getProxyClient(array $options = array())
     {
+        $options = array_merge(
+            [
+                'purge_method' => 'NOTIFY'
```
dantleech: Not sure what this has to do with anything..

tests/Functional/Fixtures/Symfony/AppCache.php
```diff
@@ -27,6 +30,9 @@ public function __construct(HttpKernelInterface $kernel, StoreInterface $store,
         $this->addSubscriber(new PurgeSubscriber(['purge_method' => 'NOTIFY']));
         $this->addSubscriber(new RefreshSubscriber());
         $this->addSubscriber(new UserContextSubscriber());
+
+        $this->addSubscriber(new TagSubscriber($tagManager ?: new NullManager()));
```
dantleech: s/NullManager/NullTagManager

src/Test/SymfonyTest.php
dbu: this was/is used to test whether the configuration option to change the purge_method works. but indeed not sure why we would need that in the Test - having it in one unit test would be enough. we should not fix this in this PR - its quite big already.

dantleech: Actually I introduced the passing of `$options` to this method in order to be able to pass in the Invalidator.

dbu: we should only change the method when the test wants to check something with it.

src/Tag/Storage/DoctrineCache.php
dbu: maybe have a utitlity class that we can trigger in a symfony command in the bundle?

composer.json
dbu: i would prefer to move it to require-dev and document you need this if you want to use the symfony CacheKernel based tagging, yes. and have a class_exists check with the same information in the place where we would run into the problem if the user did not read the doc...

tests/Functional/Fixtures/web/symfony-tag.php
```diff
@@ -0,0 +1,19 @@
...
+use FOS\HttpCache\Tests\Functional\Fixtures\Symfony\AppKernel;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\HttpCache\Store;
+
+$loader = require_once __DIR__.'/../../../../vendor/autoload.php';
+
+$symfonyProxy = new SymfonyProxy();
+
+$kernel = new AppKernel();
+$kernel = new EventDispa
```
dbu: incomplete

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,144 @@
...
+            return;
+        }
+
+        $this->tagResponse($response);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    private function tagResponse(Response $response)
```
dbu: this does not actually tag the response, right? it records the tags from the response and should then unset the header tag. can we find a better name?

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,144 @@
...
+        $contentDigest = $this->getContentDigestFromHeaders($response->headers);
+        $tags = $this->getTagsFromHeaders($response->headers);
+        $expiry = $this->getExpiryFromResponse($response);
+        $this->manager->tagCacheId($tags, $contentDigest, $expiry);
+    }
+
+    /**
+     * Return the content digest from the headers.
+     * The content digest should be set by the Symfony HTTP cache before
+     * this method is invoked.
```
dbu: is that something that symfony HttpCache does by default or is it custom?

dbu: if its custom, we should be a bit more precise here.

src/SymfonyCache/TagSubscriber.php
```diff
@@ -0,0 +1,144 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\HeaderBag;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+use FOS\HttpCache\ProxyClient\Symfony;
+
+/**
+ * Adds tag invalidation capabilities to the Symfony HTTP cache
+ *
```
dbu: we should explain the concept (header with tag information as json encoded, store cache items by tag so we can invalidate all items with that tag) a bit in this docblock

src/SymfonyCache/TagSubscriber.php
dantleech: We sure can. extractTagsFromResponse or similar.

src/Tag/ManagerInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace FOS\HttpCache\Tag;
+
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Implementing classes are responsible for associating tags with cache entries
+ * and invalidating tags (and removing cache entries).
+ */
+interface ManagerInterface extends InvalidatorInterface
```
dbu: could we call this CacheTagStoreInterface?

src/Tag/StorageInterface.php
dbu: this is the same method signature as ManagerInterface has. what are the differences?

src/SymfonyCache/TagSubscriber.php
dantleech: AFAIK it is not configurable, it is by default. But I said `should` above because I am not currently sure. Will have to have a closer look.

src/Tag/ManagerInterface.php
dantleech: Not sure that is a better name.. this is also an invalidator and there already exists a `StorageInterface`.

src/Tag/StorageInterface.php
dantleech: This method would typically be a proxy method.

The reason for the separation is in the other method: `invalidateTags`. The Symfony Tag Manager *knows* about the Symfony HTTP cache, it is responsible for removing the cache entry. The `StorageInterface` is agnostic to the cache implementations.

But I wonder now if we should remove the `ManagerInterface` and use the `StorageInterface` directly in the class which actually tags the cache entries (in this case the `TagSubscriber`) -- BUT the problem then is that you could have 2 separate implementations of the StorageInterface being used instead of 1.

Having this proxy method ensures that there is a single API (although the `TagSubscriber` currently only uses the `tagCacheId` method, it could/will also need access to `invalidateTags` in the case that invalidation is done via. a HTTP request).

src/Tag/Manager/NullTagManager.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace FOS\HttpCache\Tag\Manager;
+
+/**
+ * Null tag manager - use this manager when no tagging is required.
+ */
+class NullTagManager
```
dantleech: Missing interface.

src/Tag/ManagerInterface.php
```diff
@@ -0,0 +1,29 @@
...
+     * considered as having already been invalidated by the caching proxy.
+     *
+     * NOTE: Often this method could simply be a proxy to StorageInterface#tagCacheId.
+     *
+     * @param string[] $tags
+     * @param mixed $cacheId
+     * @param int $lifetime
+     * @return void
+     */
+    public function tagCacheId(array $tags, $cacheId, $lifetime);
```
dantleech: change method signature: `$cacheId, array $tags, $lifetime = null`

src/Tag/Storage/DoctrineCache.php
dantleech: Do we need to store the lifetime of the content/cache response?

We can just have a command which iterates over the cache entries and removes the entries which no longer have a valid cache ID.

dantleech:
@ddeboer @dbu have updated and included a new description above.

andrerom:
Not jumping into the code here, but find it a bit worrying to use cache for the tags, several cache backends uses LRU algorithm and can evict at any point. As talked to @magnusnordlander here https://github.com/php-cache/cache/pull/32#issuecomment-184906668, safest would be to persist in database (DBAL) as safe default, it would anyway one be used on store and on invalidation.

dantleech:
@andrerom the DoctrineCache implementation could be replaced (just need to implement `TagStorageInterface`. I just used it because it is simple, and if using the file cache (there are a fair few backends), I don't think anything will ever be evicted.

But saying that, if we start adding new tag storage backends I also wonder if this is the best place for this feature, it can be implemented almost entirely as middle-ware.

dantleech:
I think I am going to create a standalone middleware specifically for tag handling with the Symfony HTTP cache, which could then perhaps be integrated here. Maybe perhaps it will lead to new ideas. If people want we can then create an FOS package for it.

dantleech:
have created a new standalone repository based on this PR: https://github.com/dantleech/sf-http-cache-tagging

It could be integrated here via. an Event Subscriber.

/cc @dbu @ddeboer @andrerom

Koc:
@dantleech is it better use Symfony's PSR6 implementation (with native support of tags) instead of Doctrine Cache?

dantleech:
@Kok yes, this PR predates the Symfony cache implementation, I will close it now as it is past its sell-by-date :)

PR #256, opened 2015-12-03
Upgrade Symfony dependencies to 3.0

ddeboer:


composer.json
```diff
@@ -22,8 +22,8 @@
     ],
     "require": {
         "php": ">=5.4.8",
-        "symfony/event-dispatcher": "~2.3",
-        "symfony/options-resolver": "~2.3",
+        "symfony/event-dispatcher": "^2.3||^3.0",
```
ddeboer: Also switching to caret here, as it’s the [recommended operator](https://getcomposer.org/doc/articles/versions.md#caret).

dbu:
hm, looks like composer is having a hard time making sense of this. or are we doing the mistake?

ddeboer:
Ah, the problem is not our requirement format, but the fact that some Symfony 3.0 packages require 5.5.

dbu:
yeah. but do we force symfony 3 here? i think composer should realize we
are on a too old php and pick an older symfony, no?

however i am ok if we just force it to 2.8 on php 5.4

ddeboer:
For one or another reason ``prefer-lowest` only started behaving properly (installing 2.3 on PHP 5.4) when I split `before_install` off from `install`. Anyway, it works now. :wink:

dbu:
great! should we do this pull request against the 1.3 branch so that we can do a dot release that supports symfony 3? 2.0 might still take a moment until its ready for release.

ddeboer:
Or just merge this to master and cherry-pick it into the 1.3 branch?

dbu:
i'd rather do it the other way round, to avoid merge conflicts.

ddeboer:
Okay, closed this and moved to #259.

PR #257, opened 2015-12-04
Upgrade php-http integration

dbu:
trying to adjust to the current versions of php-http.

* [x] Tests fail because of problem with error response conversion
* [x] Some TODO in the code


src/ProxyClient/Http/HttpAdapterInterface.php
```diff
@@ -0,0 +1,29 @@
...
+namespace FOS\HttpCache\ProxyClient\Http;
+
+use FOS\HttpCache\Exception\ExceptionCollection;
+
+/**
+ * An adapter to work with Httplug.
+ *
+ * @author David Buchmann <mail@davidbu.ch>
+ */
+interface HttpAdapterInterface
```
dbu: not sure if we should call this adapter, but turns out we do need some layer between the proxy client and Httplug, or we would need to duplicate logic or share it in a base class.

src/Test/HttpClient/MockHttpClient.php
```diff
@@ -2,21 +2,19 @@
...
 /**
- * HTTP adapter mock
+ * HTTP client mock
  *
  * This mock is most useful in tests. It does not send requests but stores them
  * for later retrieval. Additionally, you can set an exception to test
  * exception handling.
  */
-class MockHttpAdapter implements HttpAdapter
+class MockHttpClient implements HttpClient
```
dbu: @sagikazarmark could this class go into the php-http/utils ? its a utility to mock a Httplug client for unit testing that is more convenient than prophesize and phpunit mocks.

src/ProxyClient/Http/HttpAdapterBase.php
```diff
@@ -0,0 +1,184 @@
...
+                if ($this->baseUri->getPath() !== '') {
+                    $path = $this->baseUri->getPath() . '/' . ltrim($uri->getPath(), '/');
+                    $uri = $uri->withPath($path);
+                }
+            }
+        }
+
+        // Close connections to make sure invalidation (PURGE/BAN) requests
+        // will not interfere with content (GET) requests.
+        // TODO: or is this the job of a plugin?
```
dbu: not sure if this should be here or if we should delegate this to a plugin of Httplug (but how to make sure that plugin is in the chain, just trust the user?)

src/ProxyClient/Http/HttpAsyncAdapter.php
```diff
@@ -0,0 +1,96 @@
...
+     * Wait for all promises to resolve and gather exceptions.
+     */
+    public function flush()
+    {
+        $promises = $this->promises;
+        $this->promises = [];
+        foreach ($promises as $promise) {
+            $promise->wait();
+            // TODO: gather exceptions
+            // Handle all networking errors: php-http only throws an exception
```
dbu: actually we can use a plugin in the chain to make http error responses exceptions. we probably should ask the user to do that and this will become easy.

src/Test/HttpClient/MockHttpClient.php
sagikazarmark: IIRC the plan was having a separate mock-client package for it. It could go into the utils package, but what if we end up having 2 or 3 more testing utility? Should those also go into the utils package or rather into a separate testing-utils package? #toomuchpackages

src/ProxyClient/Http/HttpAdapterBase.php
sagikazarmark: Base URI is optional in the first place, isn't it? In this case I think you can rely on the user, since it is not required.

src/ProxyClient/Http/HttpAdapterInterface.php
sagikazarmark: I agree, it should not be Http related at all. For example in FXMLRPC, http based client is just a transfer object. It could be SSH as well, but that's xmlrpc specific.

So this should be a generic interface with a naming which explains what it does (Invalidator?). The actual implementation can be Http dependent. Even if it does not make sense to invalidate without http, from an abstract point of view, it doesn't matter.

src/ProxyClient/Http/InvalidationRequest.php
```diff
@@ -21,7 +21,7 @@
 class InvalidationRequest implements RequestInterface
 {
     use RequestDecorator;
```
dbu: looks like we are missing a require statement. this would be in https://github.com/php-http/message-decorator - i guess we still want to use that, right? maybe @sagikazarmark split that up after @ddeboer built the initial version of this?

src/ProxyClient/Http/HttpAsyncAdapter.php
sagikazarmark: Yeah, could be a plugin. At least that would be the nice solution.

src/ProxyClient/Http/InvalidationRequest.php
sagikazarmark: Require what exactly? Haven't touched the decorator since June.

dbu: we don't require php-http/message-decorator explicitly, so tests fail because our request class does not respect the interface. probably some other lib previously required the decorator or something.

sagikazarmark: Ah, I see. AFAIK message decorator was never required by anything, but I could be wrong.

src/Test/HttpClient/MockHttpClient.php
ddeboer: I suggest just adding it to utils for now, and splitting it off later (in a php-http 2.0 version) if things get too messy.

sagikazarmark: :+1:

ddeboer: I’ll open a PR on utils.

sagikazarmark: :+1:

ddeboer: Should it be a client or would a plugin be better?

sagikazarmark: That's a good question, I see sense in both. @dbu ?

dbu: i would go for client, then you don't need a "real" client and its much
simpler.

doc/proxy-clients.rst
```diff
@@ -28,6 +28,8 @@ You also need a `PSR-7 message implementation`_. If you use Guzzle 6, Guzzle’s
 implementation is already included. If you use another client, install one of
 the implementations. Recommended:
 
+TODO: i think we need the matching adapter for the factory, if guzzle6-adapter is not installed
```
dbu: @ddeboer you did the first iteration of this and wrote this doc. are you sure this will work without the php-http adapter for guzzle?

src/ProxyClient/Http/HttpAdapter.php
```diff
@@ -0,0 +1,293 @@
...
+                $promises[] = $this->httpClient->sendAsyncRequest($proxyRequest);
+            }
+        }
+
+        if (count($exceptions)) {
+            throw new ExceptionCollection($exceptions);
+        }
+
+        foreach ($promises as $promise) {
+            // TODO: or will the final version of promise return the value in wait()?
```
dbu: @joelwurtz what was the decision about the promises? will we return the result from `wait()` or not? in my situation here, providing callback methods would make little sense. the only reason we use the async client is so that we can send the requests in parallel. the flush() method itself stays sync.

src/ProxyClient/Nginx.php
```diff
@@ -49,7 +49,7 @@ public function setPurgeLocation($purgeLocation = '')
     public function refresh($url, array $headers = [])
     {
         $headers = array_merge($headers, [self::HTTP_HEADER_REFRESH => '1']);
-        $this->queueRequest(self::HTTP_METHOD_REFRESH, $url, $headers);
+        $this->httpAdapter->invalidate($this->messageFactory->createRequest(self::HTTP_METHOD_REFRESH, $url, $headers));
```
dbu: looking at this, we could also keep the queueRequest to make this less verbose. wdyt @ddeboer ?

src/Test/HttpClient.php
```diff
@@ -113,7 +113,6 @@ private function createRequest($method, $uri, $headers)
         return MessageFactoryDiscovery::find()->createRequest(
             $method,
             $uri,
-            '1.1',
```
dbu: this was reordered in the factory to avoid having to specify it when setting headers.

src/ProxyClient/Http/HttpAdapter.php
sagikazarmark: I would like to make it that way, PR is being prepared.

doc/proxy-clients.rst
```diff
@@ -276,6 +285,5 @@ Varnish client::
 Make sure to add any headers that you want to ban on to your
 :doc:`proxy configuration <proxy-configuration>`.
 
-.. _header: http://php.net/header
```
dbu: i did not see this being used anywhere in the doc.

src/ProxyClient/Varnish.php
```diff
@@ -60,14 +60,17 @@ class Varnish extends AbstractProxyClient implements BanInterface, PurgeInterfac
...
+     *
+     * @param string $tagsHeader Header for tags, defaults to X-Cache-Tags
      */
     public function __construct(
         array $servers,
         $baseUri = null,
-        HttpAdapter $httpAdapter = null,
+        HttpAsyncClient $httpClient = null,
+        MessageFactory $messageFactory = null,
         $tagsHeader = 'X-Cache-Tags'
```
dbu: should we go with an options parameter for this, like in the Symfony proxy client?

dbu:
@sagikazarmark when looking at the travis fails, i think we have a problem with the guzzle6 adapter. created https://github.com/php-http/guzzle6-adapter/issues/13 to discuss that further

src/ProxyClient/Http/HttpAdapter.php
```diff
@@ -0,0 +1,293 @@
...
+    {
+        $queue = $this->queue;
+        $this->queue = [];
+        /** @var Promise[] $promises */
+        $promises = [];
+
+        $exceptions = new ExceptionCollection();
+
+        foreach ($queue as $request) {
+            foreach ($this->multiplex($request) as $proxyRequest) {
```
ddeboer: I’m not sure ‘multiplex’ is the right terminology. Perhaps what you mean is [inverse multiplexing](https://en.wikipedia.org/wiki/Inverse_multiplexing)? Unfortunately, that sounds a little convoluted, so perhaps just `fanOut()` or something would be better?

src/ProxyClient/Nginx.php
ddeboer: A `queueRequest` in the `AbstractClient` could make sense, yes.

src/ProxyClient/Varnish.php
ddeboer: And move `baseUri` into that as well?

src/Test/HttpClient/MockHttpClient.php
ddeboer: Created https://github.com/php-http/httplug/issues/89 to discuss the mock client.

src/ProxyClient/Varnish.php
dbu: we could, but baseUri is something of the base client. do we want to mix
base client options and specific options? though maybe we do want, as
its optional and one day we might find more optional things for the base
client, at which point it would become tricky with parameter order. i
would keep http client, message factory and servers out. servers because
they are required, the client and factory because they seem more than a
simple option. agreed?

src/ProxyClient/Http/HttpAdapter.php
dbu: i am fine with fanOut

src/ProxyClient/Varnish.php
ddeboer: Yes.

doc/proxy-clients.rst
ddeboer: I’m not sure I follow:`php-http/guzzle6-adapter` installation is mentioned on line 25.

dbu: yes, we say "If you use another client, install one of the implementations." - this is confusing me. you would need to install an adapter in that case, right?

src/ProxyClient/Varnish.php
dbu: fixed

doc/proxy-clients.rst
ddeboer: This section is about the message implementation: ‘You also need…’. Guzzle 6 comes with guzzlehttp/psr7. What I mean by ‘If you use another client, install one of the implementations.’ is this: if you use another HTTP client instead of Guzzle 6, you need to include a PSR-7 implementation yourself. We recommend to require guzzlehttp/psr7, but you can also use e.g. Diactoros.

However, I agree that we can improve clarity here.

dbu:
the guzzle setup has been fixed: https://github.com/php-http/guzzle6-adapter/pull/19

ddeboer:
This is green for guzzle6-adapter now. Fixing guzzle5-adapter installation in https://github.com/php-http/guzzle5-adapter/pull/9.

.travis.yml
```diff
@@ -22,8 +22,8 @@ branches:
     - '/^\d+\.\d+$/'
 
 install:
-  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" || "$TRAVIS_PHP_VERSION" == "hhvm" ]]; then composer remove "php-http/guzzle6-adapter" --dev --no-update; fi
-  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" || "$TRAVIS_PHP_VERSION" == "hhvm" ]]; then composer require "php-http/guzzle5-adapter" --dev --no-update; fi
+  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" ]]; then composer remove "php-http/guzzle6-adapter" --dev --no-update; fi
+  - if [[ "$TRAVIS_PHP_VERSION" == "5.4" ]]; then composer require "php-http/guzzle5-adapter" --dev --no-update; fi
```
ddeboer: Guzzle 6 now correctly runs on hhvm, yay!

ddeboer:
Okay, everything is now green except PHP 5.4. This is caused by the current php-http/guzzle5-adapter not offering async. We could fix this, of course, but I’d actually opt for dropping 5.4 support: it’s been EOL for quite some time, and big ones like Symfony 3.0 don’t support it any longer either.

@dbu Do you agree?

dbu:
yup, sounds good. lets ditch 5.4 and go for 5.5. we are talking version
2.0 so lets not burden us with things that are already legacy.

composer.json
```diff
@@ -20,20 +20,22 @@
...
         "symfony/event-dispatcher": "^2.3||^3.0",
         "symfony/options-resolver": "^2.3||^3.0",
-        "psr/http-message-implementation": "~1.0",
-        "php-http/adapter-implementation": "^0.1.0",
-        "php-http/discovery": "^0.1.1",
-        "php-http/message-decorator": "^0.1.0"
+        "php-http/client-implementation": "^1.0.0",
+        "php-http/discovery": "^0.6.0",
+        "php-http/message": "^0.2.1",
+        "puli/cli": "^1.0.0-beta8"
```
ddeboer: Adding this as a dependency so php-http/discovery works out of the box.

doc/proxy-clients.rst
ddeboer: :+1:

ddeboer:
We have a green build. Scrutinizer may still be failing, but I’d like to fix that on a separate PR. Squashed commits, so this is ready for merging.

composer.json
sagikazarmark: I would rather add this as a development dependency and document that Puli is necessary. At least it seemed to me as a recommendation from Puli. Maybe @webmozart could help in in this question.

dbu: and puli is only required when not injecting a http client explicitly. maybe we should move discovery to optional. does discovery only work with puli but not require it? that would feel wrong to me.

ddeboer: @dbu Yes, Discovery should be optional with FOSHttpCache. We can make it so in two ways:

1. just move it from `require` to `suggest` and document it properly; and maybe add conditional checks around the Discovery calls in our code
2. remove all Discovery calls from our code and document that users should either pass an HTTP client explicitly or do the discovery in their own code.

Against 2 and @sagikazarmark’s idea to have users install Puli is that it becomes a lot more work for users to get started with FOSHttpCache. 

For discovery’s Puli dependency, see https://github.com/php-http/discovery/issues/36.

sagikazarmark: No, Puli is not optional. All the relevant Puli dependencies are required in the discover package. The cli is something that is up to the user: it can be installed as a phar globally in the prefix, or installed with composer in the application.

I guess there is nothing wrong in adding it as a dependency here. But I guess some time in the future having puli.phar in the users prefix will be as common as composer.

sagikazarmark: @ddeboer Usage of discovery depends on whether you want to provide zero-config usage. I think it is probably not so important in case of this library as you need some configuration apart from the client anyway.

Generally discovery should be avoided whenever possible.

composer.json
```diff
@@ -20,20 +20,22 @@
             "homepage": "https://github.com/friendsofsymfony/FOSHttpCache/contributors"
         }
     ],
+    "minimum-stability": "dev",
+    "prefer-stable": true,
     "require": {
-        "php": ">=5.4.8",
+        "php": ">=5.5.9",
```
andrerom: nitpick, but could be `^5.5.9|^7.0` to be more explicit on future versions.

composer.json
```diff
@@ -20,20 +20,22 @@
             "homepage": "https://github.com/friendsofsymfony/FOSHttpCache/contributors"
         }
     ],
+    "minimum-stability": "dev",
```
andrerom: is this needed because of puli dependency? could it be avoided by using `^1.0.0@beta` or similar?

sagikazarmark: It doesn't matter, it is not applied in packages where this is installed.

andrerom: I know, but while rather remote because of the "prefer-stable" flag, it looked like it might cause other versions to be pulled in on travis tests on this repo as opposed to what will be pulled in on project installs

sagikazarmark: > as opposed to what will be pulled in on project installs

For the time being php http packages are mostly unstable as well, so you are right, without relying on development versions, not the same packages will be installed. Wonder how we could improve that until stable versions are out.

ddeboer: This `minimum-stability` is necessary because some of the php-http packages themselves require unstable versions of other php-http packages. If your minimum-stability is stable (the default) Composer won’t find them.

When stable versions are available, they will be installed instead: see `"prefer-stable": true` one line below.

andrerom: > For the time being php http packages are mostly unstable as well,
> This minimum-stability is necessary because some of the php-http packages themselves require unstable versions of other php-http packages

Any chance those will be stable in the near future *(by the time 2.0 of FOSHttpCache is stable)*?
Otherwise might have to document which packages are unstable in order for people to specify every single one in project install to avoid having to set `minimum-stability` to dev.

sagikazarmark: Yes, that's the plan (since we are the author of php-http packages as well :smile: )

ddeboer: @andrerom Definitely. FOSHttpCache 2.0 will not ship before we can remove the `minimum-stability` flag.

composer.json
ddeboer: Good point, it seems to becoming the standard now.

composer.json
ddeboer: >  But I guess some time in the future having puli.phar in the users prefix will be as common as composer.

True, but we need to be as user/dev-friendly as possible in the meantime. :wink:

The best solution now is for us to make Discovery an optional dependency. Installing that would be easiest if Discovery required Puli.

sagikazarmark: > Installing that would be easiest if Discovery required Puli.

Not sure I understand this. Do you propose to install puli/cli in discovery? How would it be better than installing it in this package?

composer.json
andrerom: :+1:

composer.json
dbu: do usees technically need the puli cli installed in vendors? do we somehow trigger the cli? or will discovery inform the user about running puli cli when its not done?

sagikazarmark: cli is used in composer installation hook. It generates a factory class and a list of available bindings in all packages.

dbu: the composer hook is only there if explicitly added. so i guess we should not require the cli as `require` but only in `require-dev` if we use it. (all the runtime code requirements seem already there in php-http/discovery: https://github.com/php-http/discovery/blob/master/composer.json#L16)

composer.json
```diff
@@ -20,20 +20,22 @@
...
-        "php": ">=5.4.8",
+        "php": "^5.5.9||^7.0.0",
         "symfony/event-dispatcher": "^2.3||^3.0",
         "symfony/options-resolver": "^2.3||^3.0",
-        "psr/http-message-implementation": "~1.0",
-        "php-http/adapter-implementation": "^0.1.0",
-        "php-http/discovery": "^0.1.1",
-        "php-http/message-decorator": "^0.1.0"
+        "php-http/client-implementation": "^1.0.0",
+        "php-http/discovery": "^0.6.0",
```
dbu: if the discovery is not usable without running a puli cli command or adding that command to the project's post-install commands in composer, does it make sense to require the discovery at all? or should we just document how to use discovery and puli for 0-config when wanted?

composer.json
sagikazarmark: > the composer hook is only there if explicitly added

I probably used the wrong term. It runs whether you manually add it to the composer.json or not. It's probably not the same as hook is in composer terminology. That's why discovery requires the composer plugin.

Adding it to dev dependencies is wrong I think, because it has to run in production as well.

composer.json
sagikazarmark: That's what I would like to achieve: https://github.com/php-http/discovery/issues/33

composer.json
dbu: dev dependencies are dependencies we only need for developing this library, but not necessarily when using it in a project. (not really a question of dev vs production).

i am still confused:

1. can we have full 0 config with discovery (or will have soonish?). by that i mean its enough to add fos-http-cache and a HTTPlug client/adapter to my composer dependencies and discovery + puli do their stuff "magically" without my interaction? or do i need some manual step anyways?
2. if the answer to 1. is yes, do we need the puli cli for that, or is the cli for something else?

sagikazarmark: Puli CLI collects configuration (bindings, available clients in our case) from all your installed composer packages and stores them under the .puli directory. This is done during composer install normally (by the composer plugin required by discovery) or of course you can do it manually by running `puli build`.

In order to do it, puli cli must be available somewhere. Either in your prefix as an executable, or by installing the puli cli package in your project or library.

To achieve full zero config, you have two options:

1. assume puli cli is installed by the user (either the phar or the package installed in the application where this library is used)
2. Install puli cli with this library which means puli cli will be made sure to be installed. However this way you don't let the user choose the way. If the puli CLI is already available, you essentially force the user to install the dependency anyway. I don't say it is bad, but I guess it is not how puli is imagined to work. That's why I said that for the long-term puli executable can be expected to be installed by the user, but for now, we can't expect this.

dbu: so the composer plugin relies on the puli cli being either in the
vendors or installed globally?

what happens if it is not? will composer install explode, or is it
silently ignored and there are no resources? can discovery figure out
whether there are no resources because no adapter was installed or
because puli cli is missing? aka, can we properly report the problem to
the user?

sagikazarmark: When whe CLI is there it generates a class. I can't tell you what happens during composer installation if puli cli is missing, haven't tried it yet. But the discovery reports if the Puli Factory class, which is generated by the build command is missing.

ddeboer: Discovery will not work properly if puli/cli isn't installed (whether globally or in the vendors). We should solve this:

1. by making puli/cli a dependency of Discovery (for now)
2. when Puli has become more widespread, remove the dependency and note in the Discovery docs that Puli must be installed globally (as we now docs usually mention Compoer)
3. optionally, make Dependency an optional dependency of FOSHttpCache. 

I'm not sure about 3. On the one hand, it spares people the trouble of some extra deps (in particular, Puli). On the other, requiring Discovery manually is an extra step for our users, who may not really care about this HTTPlug thing as long as it works.

sagikazarmark: I also started to think that puli/cli should be a dependency of discovery, at least for now.

dbu: > I also started to think that puli/cli should be a dependency of
discovery, at least for now.

i think i understand now. thanks for the patience! agree we should depend
on it for discovery. maybe one day composer will detect a globally
installed puli cli like it detects php version and php extensions... its a
bit like with phpunit, but not only for testing. as discovery is all about
simplicity and easy first steps, lets depend on puli cli for now.

sagikazarmark: :+1:

ddeboer: @dbu No problem. :wink:

How will we then handle Discovery here?

* either we require it, no questions asked
* or we suggest it and add a note to our docs.

sagikazarmark: no questions asked :smile:

sagikazarmark: Ah, sorry, missed the question, I thought it's still discovery-puli.

dbu: imho it would make more sense if php-http/discovery would have the cli as
requirement than we have here. then discovery could be either suggestion or
require, but only one thing not two.

dbu: https://github.com/php-http/discovery/pull/39

dbu:
yay!

lets continue to tweak things, but i wanted to merge now as this PR was getting rather large

sagikazarmark:
Discovery 0.6.2 tagged

ddeboer:
:+1:

PR #259, opened 2015-12-04
Upgrade Symfony dependencies to 3.0

ddeboer:


dbu:
merged, thanks! but we really have to sort out the php 7 and hhvm mess. its like mostly a travis problem i fear, but still annyoing to have failing builds...

PR #262, opened 2015-12-05
[WIP] Fix Mockery TestListener failing on HHVM and PHP 7

ddeboer:


dbu:
looks like we need a newer version of mockery: "Class Mockery\Adapter\Phpunit\TestListener contains 1 abstract method and must therefore be declared abstract"

dbu:
hhvm seems to have a fundamental setup issue. if we don't find why that is, i am ok to just ignore it for now and wherever we say its supported say instead that it might work but if anybody is using hhvm they should send a PR to fix the build.

ddeboer:
Yeah, I'm still working on this one. We had a lot of different errors due to Travis' new legacy (i.e., non-container) build infrastructure but it seems we're now back on the old infrastructure. Really confusing.

ddeboer:
Just by restarting the build on Travis, the PHP 7 build became green again, even though we’re still using the same 0.9.4 Mockery version that is [green on master](https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/jobs/94631340).

Unfortunately, the restart also caused weird HHVM 404s again.

dbu:
probably travis fixed their PHP 7 setup then?
lets ignore the HHVM build fails and add a note to the doc then. if
somebody cares about HHVM, they should spend the effort to make the
tests work.

ddeboer:
Closing this, as we currently have [successful builds on all branches](https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/branches). We may want to come back to this PR when builds start failing again due to changes in Travis’s infrastructure.

PR #271, opened 2016-01-21
[cache tagging] add tags sanity check

### LINKED ISSUE #270: [cache tagging] add tags sanity check
would it make sense to add a sanity check to the [addTag](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/5a3d88ad9093f983601c629ca858943d76216b47/src/ResponseTagger.php#L88) function of the ``ResponseTagger``?

```php
$this->tags = array_merge($this->tags, array_filter($tags, 'strlen'));
```
would prevent an empty tag (``null``, ``''``, ``false``) from being added to the collection of tags.

generation of an empty tag can, for instance, occur if a configured expression function returns an empty string after evaluation

dbu:
fixes #270

### DIFF
--- a/src/ResponseTagger.php
+++ b/src/ResponseTagger.php
@@ -12,6 +12,8 @@
 namespace FOS\HttpCache;
 
 use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
+use FOS\HttpCache\Exception\InvalidTagException;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Service for Response cache tagging.
@@ -19,9 +21,16 @@ use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
  * @author David de Boer <david@driebit.nl>
  * @author David Buchmann <mail@davidbu.ch>
  * @author André Rømcke <ar@ez.no>
+ * @author Wicliff Wolda <wicliff.wolda@gmail.com>
  */
 class ResponseTagger
 {
+
+    /**
+     * @var array
+     */
+    private $options;
+
     /**
      * @var TagsInterface
      */
@@ -36,10 +45,21 @@ class ResponseTagger
      * Constructor
      *
      * @param TagsInterface $client
+     * @param array         $options supported options:
+     *                               - strict (bool) Default: false. If set to true, throws exception when adding empty tags.
      */
-    public function __construct(TagsInterface $client)
+    public function __construct(TagsInterface $client, array $options = array())
     {
         $this->client = $client;
+
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults(array(
+            'strict' => false,
+        ));
+
+        $resolver->setAllowedTypes('strict', 'bool');
+
+        $this->options = $resolver->resolve($options);
     }
 
     /**
@@ -81,11 +101,19 @@ class ResponseTagger
      *
      * @param array $tags List of tags to add.
      *
+     * @throws InvalidTagException
+     *
      * @return $this
      */
     public function addTags(array $tags)
     {
-        $this->tags = array_merge($this->tags, $tags);
+        $filtered = array_filter($tags, 'strlen');
+
+        if ($this->options['strict'] && array_diff($tags, $filtered)) {
+            throw new InvalidTagException('Empty tags are not allowed');
+        }
+
+        $this->tags = array_merge($this->tags, $filtered);
 
         return $this;
     }
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
 matrix:
   include:
     - php: 5.5
-      env: SYMFONY_VERSION=2.3.* VARNISH_VERSION=3.0 COMPOSER_FLAGS="--prefer-lowest"
+      env: SYMFONY_VERSION=2.7.* VARNISH_VERSION=3.0 COMPOSER_FLAGS="--prefer-lowest"
 
 branches:
   only:
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
     "require": {
         "php": "^5.5.9||^7.0.0",
         "symfony/event-dispatcher": "^2.3||^3.0",
-        "symfony/options-resolver": "^2.3||^3.0",
+        "symfony/options-resolver": "^2.7||^3.0",
         "php-http/client-implementation": "^1.0.0",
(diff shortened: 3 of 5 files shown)


src/ResponseTagger.php
```diff
@@ -36,10 +43,24 @@ class ResponseTagger
...
+     * @param array         $options
+     *
+     * @throws InvalidArgumentException
      */
-    public function __construct(TagsInterface $client)
+    public function __construct(TagsInterface $client, array $options = array())
     {
         $this->client = $client;
+
+        foreach ($options as $name => $value) {
```
dbu: please use the OptionsResolver class for this. its more readable and simpler.

src/ResponseTagger.php
```diff
@@ -85,7 +106,13 @@ public function hasTags()
      */
     public function addTags(array $tags)
     {
-        $this->tags = array_merge($this->tags, $tags);
+        $filtered = array_filter($tags, 'strlen');
+
+        if ($this->strict && array_diff($filtered, $tags)) {
+            throw new InvalidTagException('Empty tags are not supported');
+        } else {
```
dbu: no need for the else. in the if you throw an exception, so the execution flow is stopped.

src/ResponseTagger.php
```diff
@@ -36,10 +43,24 @@ class ResponseTagger
      * Constructor
      *
      * @param TagsInterface $client
+     * @param array         $options
```
dbu: the text part of the phpdoc should list the supported options and explain them (terse phrase is ok, something like "strict: if true, throws exceptions on empty tags")

dbu:
@wickedOne i just created a pull request from your diff. that is easier to comment :-)

i think you are on the right track. i added some comments about details, and we should have two tests for empty tags, one with strict, one without.

about the setup issues: puli seems to not be that stable yet, unfortunately. i will try to have a look today to fix things in the master branch. when that works, you can rebase this branch on the fixed master and it should install again.

wickedOne:
@dbu thanks for the pointers, will be taken into account when i continue to work on it.
might be a couple of days before i have enough time on my hands to get back to it though, probably next week…

wickedOne:
now using the options resolver and including some untested tests :sunglasses: 
i'll wait for the dependencies to be fixed to test the tests and see whether the coverage is sufficient

dbu:
okay, i pushed a fix for the setup to master (updating php-http/discovery to the latest version fixed the issues). can you please rebase this branch on master?

src/ResponseTagger.php
```diff
@@ -36,10 +50,31 @@ class ResponseTagger
...
+
+        $this->options = self::$resolver->resolve($options);
+    }
+
+    /**
+     * Configure options
+     *
+     * @param OptionsResolver $resolver
+     */
+    public function configureOptions(OptionsResolver $resolver)
```
dbu: why public and not private? and why a method? i would inline this in the constructor. and not use a static variable either, just a local variable. i don't expect there to be more than one response tagger.

wickedOne: so someone can override the method in an extend?

granted, not much going on in the costructor at the moment, but was thinking that it would be easier once more options were added to the ``ResponseTagger``

will fix this along with the rebase and the tests on tuesday

dbu: > so someone can override the method in an extend?

then protected would be enough. i'd move that to the constructor for
now. if somebody comes up with a good reason why they want to extend it,
we can still refactor to a protected method without causing any BC break
so they can extend it.

wickedOne:
sorry, took me a while to get puli to do my bidding…
rebased and fixed.

dbu:
should we do validation when invalidating tags as well? maybe build it into https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/AbstractProxyClient.php#L138 ?

src/ResponseTagger.php
```diff
@@ -36,10 +45,20 @@ class ResponseTagger
...
-    public function __construct(TagsInterface $client)
+    public function __construct(TagsInterface $client, array $options = array())
     {
         $this->client = $client;
+
+        $resolver = new OptionsResolver();
+
+        $resolver->setDefaults(array(
+            'strict' => false,
+        ));
```
dbu: lets add validation: `$resolver->setAllowedTypes('strict', 'bool');`

src/ResponseTagger.php
```diff
@@ -81,11 +100,19 @@ public function hasTags()
...
+     *
      * @return $this
      */
     public function addTags(array $tags)
     {
-        $this->tags = array_merge($this->tags, $tags);
+        $filtered = array_filter($tags, 'strlen');
+
+        if ($this->options['strict'] && array_diff($tags, $filtered)) {
+            throw new InvalidTagException('Empty tags are not supported');
```
dbu: s/supported/allowed/

dbu:
cool! have some last tweaks, then this is ready to merge. the idea of sanity-checking when invalidating tags would be for another PR i think.

src/ResponseTagger.php
wickedOne: hmz, this causes the symfony 2.3 build to break… 
``setAllowedTypes`` was refactored in version 2.6 and i can't seem to find a way around it which supports both the 2.3 and the >= 2.6 ``OptionsResolver``

src/ResponseTagger.php
```diff
@@ -36,10 +45,21 @@ class ResponseTagger
...
+    public function __construct(TagsInterface $client, array $options = array())
     {
         $this->client = $client;
+
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults(array(
+            'strict' => false,
+        ));
+
+        $resolver->setAllowedTypes('strict', 'bool');
```
dbu: ups. but this is a new major version, so lets just require at least 2.7 of options resolver in composer.json. please also adjust .travis.yml where it says symfony 2.3 atm.

wickedOne:
the ``php-http/message`` errors @ travis are due to my modifications?

dbu:
i hope b4fa70ae31174c51e094f6e896a33e04498423d2 fixes the build error. can you please rebase on master one more time? sorry about that - php-http was released yesterday and it seems some 0.x branches we rely on where never tagged so they stopped working.

wickedOne:
no problem, rebased with master…

dbu:
thanks a lot!

PR #273, opened 2016-01-28
[cache tagging] http_resp_hdr_len check

### LINKED ISSUE #269: [cache tagging] http_resp_hdr_len check
in order to prevent the [tag expression](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/f9769032eb43f6d1396cb60cf03550ea99257e60/src/ProxyClient/Varnish.php#L108) to exceed the configured [http_resp_hdr_len](https://www.varnish-cache.org/docs/3.0/reference/varnishd.html) value, i've implemented a check which will invalidate the tags in chunks if applicable.

would you guys be interested to have that in this library?

wickedOne:
fixes #269 

###### notes:
- as ``getTagsHeader`` and ``setTagsHeader`` are marked as depricated in the ``CacheInvalidator`` i didn't provide seperate functions for the ``headerLength``.
- as varnish' ``http_resp_hdr_len`` defaults to 8000 bytes, i defaulted the ``headerLength`` to 7500 bytes to be on the safe side

### DIFF
--- a/src/Handler/TagHandler.php
+++ b/src/Handler/TagHandler.php
@@ -33,6 +33,11 @@ class TagHandler
     private $tagsHeader;
 
     /**
+     * @var int
+     */
+    private $headerLength;
+
+    /**
      * @var array
      */
     private $tags = array();
@@ -40,18 +45,20 @@ class TagHandler
     /**
      * Constructor
      *
-     * @param CacheInvalidator $invalidator The invalidator instance.
-     * @param string           $tagsHeader  Header to use for tags, defaults to X-Cache-Tags.
+     * @param CacheInvalidator $invalidator  The invalidator instance.
+     * @param string           $tagsHeader   Header to use for tags, defaults to X-Cache-Tags.
+     * @param int              $headerLength Maximum header size in bytes, defaults to 7500.
      *
      * @throws UnsupportedProxyOperationException If CacheInvalidator does not support invalidate requests
      */
-    public function __construct(CacheInvalidator $invalidator, $tagsHeader = 'X-Cache-Tags')
+    public function __construct(CacheInvalidator $invalidator, $tagsHeader = 'X-Cache-Tags', $headerLength = 7500)
     {
         if (!$invalidator->supports(CacheInvalidator::INVALIDATE)) {
             throw UnsupportedProxyOperationException::cacheDoesNotImplement('BAN');
         }
         $this->invalidator = $invalidator;
         $this->tagsHeader = $tagsHeader;
+        $this->headerLength = $headerLength;
     }
 
     /**
@@ -62,6 +69,16 @@ class TagHandler
     public function getTagsHeaderName()
     {
         return $this->tagsHeader;
+    }
+
+    /**
+     * Get the maximum HTTP header length.
+     *
+     * @return int
+     */
+    public function getHeaderLength()
+    {
+        return $this->headerLength;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Handler/TagHandler.php
```diff
@@ -112,9 +129,26 @@ public function addTags(array $tags)
...
+        $escapedTags = array_map('preg_quote', $this->escapeTags($tags));
+
+        if (mb_strlen(implode('|', $escapedTags)) >= $this->headerLength) {
+            /*
+             * estimate the amount of tags to invalidate by dividing the max
+             * header length by the largest tag (minus 1 for the implode character)
+             */
+            $tagsize = max(array_map('mb_strlen', $escapedTags));
+            $elems = floor($this->headerLength / ($tagsize - 1)) ? : 1;
+
```
dbu: we could do `} else { $elems = count($escapedTags); }`

and then have the foreach outside the if/else, to avoid the 3 redundant lines of code in the else below.
that should not cost much extra processing power, and avoids code duplication. there would be only one chunk so the foreach would only be executed once.

src/Handler/TagHandler.php
```diff
@@ -40,18 +45,20 @@ class TagHandler
...
-     * @param CacheInvalidator $invalidator The invalidator instance.
-     * @param string           $tagsHeader  Header to use for tags, defaults to X-Cache-Tags.
+     * @param CacheInvalidator $invalidator  The invalidator instance.
+     * @param string           $tagsHeader   Header to use for tags, defaults to X-Cache-Tags.
+     * @param int              $headerLength Maximum header size in bytes, defaults to 7500.
      *
      * @throws UnsupportedProxyOperationException If CacheInvalidator does not support invalidate requests
      */
-    public function __construct(CacheInvalidator $invalidator, $tagsHeader = 'X-Cache-Tags')
+    public function __construct(CacheInvalidator $invalidator, $tagsHeader = 'X-Cache-Tags', $headerLength = 7500)
```
dbu: we should probably switch to an OptionsResolver for tags header and header length. more expressive and more flexible for future configuration options. we would have `array $options = []` here and then something like

```
$resolver->setDefaults([
    'tags_header' = 'X-Cache-Tags',
    'header_length' = 7500,
]);
$resolver->setAllowedType('tag_header', 'string');
$resolver->setAllowedType('header_length', 'int');

$options = $resolver->resolve($options);
$this->tagsHeader = $options['tags_header'];
$this->headerLength = $options['header_length'];
```

wickedOne: had a thought about that indeed, but as it was considered a bugfix i went with a construction which wouldn't break any current implementations…

nevertheless willing to change it to use an ``OptionsResolver``

dbu: oups, you are right. we should not change this in a bugfix.

will take a note for 2.0

src/Handler/TagHandler.php
dbu: what do you think about this idea?

wickedOne: yes, makes sense!

src/Handler/TagHandler.php
wickedOne: yes, will take that into account for the 2.0 pr (which has to be a separate pr anyway)

dbu: i created #274 for this.

and i want to try to merge 1.4 into master once this is merged, and find a solution. if you do a completely separate PR on 2.0, we can't update master from 1.4 anymore which could lead to loss of bugfixes.

wickedOne: is it possible to merge this into master? 
the master branch doesn't have a ``TagHandler.php`` anymore and tag invlidation has moved to the ``ProxyClient`` so figured i had to do a seperate pr to get this into the master…

dbu: my plan is to create a branch update-master from 1.4, merge master into it and solve conflicts. as part of that we would shuffle things around as needed. this amounts to about the same work as doing a separate PR, but has the benefit that git understands in the future how to merge 1.4 into master again.

wickedOne: ok, i didn't know that was possible…

moved the [``foreach``](https://github.com/FriendsOfSymfony/FOSHttpCache/pull/273#discussion_r51232912) outside the ``if`` statement like you suggested

dbu:
looks good, thanks a lot!

can you please add a note in the CHANGELOG.md file?

wickedOne:
done

wickedOne:
you think those travis errors related to this pull request? 
can't seem to replicate them…

dbu:
thanks! the test failure looks unrelated indeed. seems our way of starting servers has become unreliable somehow.

do you want to have a go at porting things to master? creating a branch from 1.4 and do a PR to master from that branch.

dbu:
i released https://github.com/FriendsOfSymfony/FOSHttpCache/releases/tag/1.4.2 with this fix.

wickedOne:
ok, will have a go at merging 1.4 into master

PR #275, opened 2016-02-02
Update master with 1.4.2

wickedOne:
this should be it…


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 Changelog
 =========
+1.4.2
+-----
+
+* The TagHandler constructor now accepts a ``headerLenght`` argument which will
+  cause it's ``invalidateTags`` function to invalidate in batches if the header
+  length exceeds this value.
```
dbu: please move this below the 2.0 section, above 1.4.0

src/CacheInvalidator.php
```diff
@@ -137,6 +137,7 @@ public function getEventDispatcher()
     }
 
     /**
+
```
dbu: unnecessary newline

dbu:
great, that was quick! looks good, commented a few details.

can you also add a note about this in the documentation? http://foshttpcache.readthedocs.org/en/stable/invalidation-handlers.html#custom-tags-header (that bit is not updated to say we need to pass an array, btw)

wickedOne:
all should be fixed by now. also added a tiny bit of documentation for the ``strict`` parameter of the ``ResponseTagger`` as i completely forgot that one as well…

dbu:
thanks a lot!

PR #278, opened 2016-02-23
fix user context caching documentation

dbu:
fix https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/285

i added this documentation in master only, and its simply incorrect and misleading. sorry about that.


pasdeloup:
Yes, it makes more sense to me, now I understand.
Thanks!
By the way, great job for this bundle.

PR #285, opened 2016-04-06
drop the x- prefix from our custom headers

### LINKED ISSUE #279: remove x- from our headers
[rfc6648](http://tools.ietf.org/html/rfc6648) says to not prefix headers with x- but if anything use a vendor prefix. lets change everything to fos- instead of x-

dbu:
fix #279

[rfc6648](http://tools.ietf.org/html/rfc6648) says to not prefix custom headers with x-. this PR removes all x- prefixes from default header names we use in this library.

X-Symfony-Cache is done by symfony, not by us.

Question: Is this a good idea and makes the library cleaner, or a pedantic change that is an unnecessary BC break and leads to tons of confusion? Upgrading applications have to also change all their VCL or nginx configuration or will end with broken systems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 2.0.0 (unreleased)
 ------------------
 
+* Change all our custom headers to not use a `X-` prefix to comply with
+  [RFC 6648](http://tools.ietf.org/html/rfc6648#section-3).
 * Replace hard coupling on Guzzle HTTP client with HTTP adapter. You now need
   to explicitly specify the adapter you want, see [installation instructions]
   (http://foshttpcache.readthedocs.org/en/stable/installation.html)
--- a/doc/cache-invalidator.rst
+++ b/doc/cache-invalidator.rst
@@ -148,9 +148,9 @@ You can also invalidate the cache based on any headers.
 Cache client implementations should fill up the headers to at least have the
 default headers always present to simplify the cache configuration rules.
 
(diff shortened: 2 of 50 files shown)


dbu:
@ddeboer seems i missed to adjust something, but what do you think in general, do we want to do this? i notice that i kind of like a prefix to see which headers are standard and which are done by us... but the rfc seems to say we should do it like this.

dbu:
@ddeboer ping

dbu:
ping @ddeboer

TomasVotruba:
Looks fine! Would that be a BC break, so aiming for next mayor realease?

dbu:
@TomasVotruba yes, this is a PR against master which will be 2.0. current stable versions are 1.x

TomasVotruba:
Allright then

dbu:
may i ask you for your opinion @Tobion @lsmith77 ?

Question: Is this a good idea and makes the library cleaner, or a pedantic change that is an unnecessary BC break and leads to tons of confusion? Upgrading applications have to also change all their VCL or nginx configuration or will end with broken systems.

lsmith77:
In general the RFC to deprecate "-X" prefix makes it clear that they talk about "new standards"
https://tools.ietf.org/html/rfc6648

ie. they do not say that existing (defacto) standards need to be updated but just that when defining new standards one should avoid the "-x" prefix.

So I would say its not necessary to push people away from "-x" in currently working scenarios.

dbu:
@lsmith77 thanks for this input. i think the discussion started when we discussed why our prefixes are X- and not something like FOS- . but at that point, i think i will simply drop this. i committed the very few general cleanups that where included in this PR to master.

PR #289, opened 2016-06-29
adds a noop client which implements all interfaces but …

gsdevme:
As discussed here https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/293

Each method is implemented from the interfaces and follows the docblocks, also using a constructor similar to the abstract. I can't write any tests for this though as phpunit is not included within composer require-dev :O

### DIFF
--- a/src/ProxyClient/Noop.php
+++ b/src/ProxyClient/Noop.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCache package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\PurgeInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
+use Http\Message\MessageFactory;
+
+/**
+ * Class Noop
+ *
+ * This is a no operation client, its only purpose is to provide an implementation for use in an enviroments that
+ * have no proxy to use.
+ *
+ * @author Gavin Staniforth <gavin@gsdev.me>
+ */
+class Noop implements ProxyClientInterface, BanInterface, PurgeInterface, RefreshInterface, TagsInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function ban(array $headers)
+    {
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function banPath($path, $contentType = null, $hosts = null)
+    {
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function invalidateTags(array $tags)
+    {
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTagsHeaderValue(array $tags)
+    {
+        return [];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTagsHeaderName()
+    {
+        return 'X-Noop-Cache-Tags';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function purge($url, array $headers = [])
+    {
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function refresh($url, array $headers = [])
+    {
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function flush()
+    {
+        return 0;
+    }
+}
--- a/src/ProxyClient/Noop.php
+++ b/src/ProxyClient/Noop.php
@@ -15,11 +15,8 @@ use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\PurgeInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
 use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
-use Http\Message\MessageFactory;
 
 /**
- * Class Noop
- *
  * This is a no operation client, its only purpose is to provide an implementation for use in an enviroments that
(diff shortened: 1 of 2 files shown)


src/ProxyClient/NoopClient.php
```diff
@@ -0,0 +1,90 @@
...
+use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
+use Http\Message\MessageFactory;
+
+/**
+ * Class Noop
+ *
+ * @author Gavin Staniforth <gavin@gsdev.me>
+ */
+class Noop implements ProxyClientInterface, BanInterface, PurgeInterface, RefreshInterface, TagsInterface
```
dbu: for PSR-2 class loading, class name and file name must match. to be consistent, please rename the file to Noop.php

src/ProxyClient/NoopClient.php
```diff
@@ -0,0 +1,90 @@
...
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\PurgeInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
+use Http\Message\MessageFactory;
+
+/**
+ * Class Noop
```
dbu: please explain in this doc block that this client is to avoid errors when some environments don't have a caching proxy.

src/ProxyClient/NoopClient.php
gsdevme: Oh shoot! Copy/Paste. Good spot.

dbu:
thanks a lot!

regarding phpunit tests: we use the phpunit that is installed on travisci. locally you can either use a globally installed phpunit or download the phpunit.phar into your project. i know its an ongoing debate whether phpunit should be in require-dev or not...
in this case, i feel we don't have the foundations to add a good test. the only meaningful test i can think of would be a base proxy client test that checks that all methods that are supposed to be chainable indeed return the client object, and run that test with each client implementation. if you want to write that, please do, but i can live without one in this case.

gsdevme:
I have a second PR to add phpunit to `require-dev` shall I hold off with that? My test was just going to ensure the docblocks are enforced in terms of types.

dbu:
given how much trouble BC breaks in phpunit major version changes are (and that travisci just updates phpunit without our control), i would be ok to have it in require-dev, actually. and yes, such a test makes sense - most sense is a base conformity test that is extended for each client to ensure that all of them act according to the interface doc. but if you don't want to go there, a PR for just the test on Noop is good to, i might do the generalization later.

tests/Unit/ProxyClient/NoopTest.php
```diff
@@ -0,0 +1,68 @@
...
+    private $noop;
+
+    protected function setUp()
+    {
+        $this->noop = new Noop();
+    }
+
+    public function testBan()
+    {
+        $this->assertInstanceOf(BanInterface::class, $this->noop->ban(['header-123']));
```
dbu: i think we should `assertSame($this->noop, $this->noop->ban...)` here. the contract specifically says we return the client itself, not any ban capable client.

gsdevme: Totally agree here, my IDE picked up `$this` as that and as I like a train monkey I just did it. Will sort later today.

gsdevme:
@dbu Rebased with the changes you have merged into master.

wickedOne:
a bit late to the party. probably nitpicking and possibly not relevant as this in the library rather than the bundle, but wouldn't it make more sense to call it a NullClient, like the [NullOutput](https://github.com/symfony/console/blob/master/Output/NullOutput.php), [NullContext](https://github.com/symfony/asset/blob/2.7/Context/NullContext.php) etc.?

gsdevme:
Not really, I don't feel like NULL has anything to do with this, its a no operation client.

wickedOne:
fair enough... 

as said, just a suggestion: using a Null[something] is a common practice in Symfony components and as this is a new feature i thought i'd thow it in here

gsdevme:
Yeah I've noticed Symfony does prefer to use NULL which is a little odd, noop as a concept has existed since forever.. Not that Im really against someone changing this.

`In computer science, a NOP or NOOP (short for No Operation) is an assembly language instruction, programming language statement, or computer protocol command that does nothing.`

wickedOne:
personally i _think_ it's a matter of taste. ``noop`` makes perfect sense. my remark was regarding the symfony way of dealing with this and as said, this is the lib so not sure whether we should take that into account.

at the moment especially the ``config_test.yml`` files of the applications i'm working on contain quite a bit of ``NullHandler`` and ``NullOutput`` settings.
so for me personally a ``NullClient`` would make more sense, but once more _probably_ a matter of taste...

gsdevme:
Calling it `NullClient` would mean we would have to rename `Varnish` to `VarnishClient` and `Symfony` to `SymfonyClient` also though? it would have to be named `Null` to keep it consistent and naming something `Null` isn't possible.

calumbrodie:
I'm 👍 for NullClient FWIW

src/ProxyClient/Noop.php
```diff
@@ -0,0 +1,93 @@
...
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\PurgeInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\TagsInterface;
+use Http\Message\MessageFactory;
+
+/**
+ * Class Noop
```
dbu: this line does not add information. i'd rather just have the second part, which explains very well what this class does.

dbu:
seems indeed a matter of taste to me. Noop says well what this does, lets stick with that.

@gsdevme i added a comment about the docblock, and please check styleci, there are some unused `use` statements. when you cleaned those up, i will merge.

gsdevme:
Updated

dbu:
thanks! 
i noticed that documentation was missing, added a bit in 4a96c2cbdeabae1176f2f055ab24e593eb039182

do you now continue with a PR on the bundle? if you do, please also add documentation in the appropriate places to that people realize that this feature exists.

gsdevme:
Thanks for accepting! Will do!

PR #290, opened 2016-06-29
chore(composer): adds phpunit to --dev

gsdevme:
PHPUnit 5.4 has a BC for this project, so I've added `<5.4` until that is resolved. I would _guess_ we need at least PHP 4.5+ but left it unrestricted for now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,8 @@
         "php-http/mock-client": "~0.1",
         "symfony/process": "^2.3||^3.0",
         "symfony/http-kernel": "^2.3||^3.0",
-        "puli/composer-plugin": "^1.0.0-beta9@beta"
+        "puli/composer-plugin": "^1.0.0-beta9@beta",
+        "phpunit/phpunit": "^4.5.0 || ^5.0.0 <5.4"
     },
     "suggest": {
         "php-http/discovery": "When not explicitly instantiating the php-http client for the proxy client",
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
   - if [ "$TRAVIS_PHP_VERSION" = 'hhvm' ]; then ./tests/ci/install-apache-hhvm.sh; fi
 
 script:
-  - phpunit --coverage-clover=coverage.clover
+  - vendor/bin/phpunit --coverage-clover=coverage.clover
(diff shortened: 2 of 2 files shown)


dbu:
please also adjust [travis.yml](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/.travis.yml#L39) to run ./bin/phpunit instead of the globally installed phpunit so that we see if we installed a workable version.

composer.json
```diff
@@ -37,7 +37,8 @@
         "php-http/mock-client": "~0.1",
         "symfony/process": "^2.3||^3.0",
         "symfony/http-kernel": "^2.3||^3.0",
-        "puli/composer-plugin": "^1.0.0-beta9@beta"
+        "puli/composer-plugin": "^1.0.0-beta9@beta",
+        "phpunit/phpunit": "<5.4"
```
dbu: we have one build with composer --prefer-lowest to check that too old versions are not ok. this should probably be `^5.0.0 <5.4` to avoid too old versions of phpunit.

gsdevme: Sure, updated :)

gsdevme:
@dbu Seems we can't have PHPUnit 5 as the project is still supporting PHP 5.5... although given 5.5 is now EOL should it not be increased to PHP 5.6 in the composer and dropped from the travis?

dbu:
is there a reason we need phpunit 5 and can't use phpunit 4? symfony itself, even symfony 3.1, still supports php 5.5 so i would love to keep supporting that for a while still.

gsdevme:
Like I said, I think PHP 4.5 would be ok but I would need to test. Symfony 3.1 might support PHP 5.5 but Doctrine doesn't to my understanding? I will try with 4.x versions later today and find out which one works.

dbu:
maybe `^4.5.0 || ^5.0.0 <5.4`?

dbu:
does the version constraint i proposed to allow both phpunit 4 and 5 work out?

gsdevme:
Sorry I haven't had chance to try, will pick it up today.

gsdevme:
@dbu Looks like thats working nicely, StyleCI is failing but nothing to do with this. Once merged I will rebase into my other PR to ensure thats working.

dbu:
thanks, looks good. indeed the styleci complaints should go away once this is merged to master.

PR #300, opened 2016-07-18
Updating dependencies, using discovery 1.0

### LINKED ISSUE #268: should we depend on discovery or not?
see https://github.com/php-http/discovery/issues/33 and https://github.com/php-http/discovery/issues/41 for some discussion. in the future, depending on discovery which needs puli might even trigger downloading puli.phar... 

as the user needs to specify a HTTPlug adapter explicitly anyways, we could just link to discovery and stop using the discovery inside FOSHttpCache alltogether. this would also mean we have to avoid discovering the message factory.

Nyholm:
This will remove the dependency on Puli. The php-http/discovery library is very lightweight, we should consider moving it from require-dev to require. 

This will fix #268

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,12 +27,12 @@
         "symfony/event-dispatcher": "^2.3||^3.0",
         "symfony/options-resolver": "^2.7||^3.0",
         "php-http/client-implementation": "^1.0.0",
-        "php-http/message": "^1.0.0"
+        "php-http/message": "^1.0.0",
+        "php-http/discovery": "^1.0"
     },
     "require-dev": {
         "mockery/mockery": "~0.9.1",
         "monolog/monolog": "^1.0",
-        "php-http/discovery": "^1.0",
         "php-http/guzzle6-adapter": "^1.0.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -32,12 +32,11 @@
     "require-dev": {
         "mockery/mockery": "~0.9.1",
         "monolog/monolog": "^1.0",
-        "php-http/discovery": "~0.8",
+        "php-http/discovery": "^1.0",
```
dbu: yes, lets move it to require.

Nyholm:
Those style issues was not introduced by me. Should I fix them?

dbu:
i will see about the style issue. i think its an update from the symfony code standard.

i was wondering if we need to remove mention of puli in the doc, but the docs actually are still about version 1 with explicit guzzle binding: http://foshttpcache.readthedocs.io/en/latest/proxy-clients.html

i will open an issue for that so we don't forget to update the docs before releasing.

Nyholm:
Thank you for merging.

PR #303, opened 2016-08-05
fix typo in sprintf usage

EvaLok:
copied the example and was wondering why it wasn't working.. then noticed ;)

### DIFF
--- a/doc/invalidation-handlers.rst
+++ b/doc/invalidation-handlers.rst
@@ -45,10 +45,10 @@ correct tags on all responses. You can add tags to the handler using::
 
 Before any content is sent out, you need to send the tag header_::
 
-    header(sprintf('%s: %s'),
+    header(sprintf('%s: %s',
         $tagHandler->getTagsHeaderName(),
         $tagHandler->getTagsHeaderValue()
-    );
+    ));
 
 .. tip::
 


dbu:
ups, thanks for the fix!

ddeboer:
And then we had this error again on master, cherry-picked: d126f162562867951d188b860e3d745cb05435b6. :wink:

dbu:
we should be able to merge whole 1.4 into master, no?

stof:
would be better indeed

ddeboer:
True, but this was the [only change](https://github.com/FriendsOfSymfony/FOSHttpCache/compare/1.4.2...master).

stof:
@ddeboer a merge is still better, as it means that the same commit is in both branches.
With a cherry pick, we have to manually look in the master history to discover that the patch was also applied, instead of using a comparison view again and seeing that no commits are in 1.4 without being in master.
This is why the Symfony maintenance workflow (used by this bundle as well, until your cherry-pick) is based on merging maintained branches rather than cherry picking.

ddeboer:
I know, merging would have been better – a slip-up due to simultaneously working on another project where we (necessarily) do cherry-picking instead because of diverging branches. 

So, should I merge anyway (and have 2 identical commit messages on master)?

stof:
I would say yes, as it allows to go back to the merge-based process for this repo (which is much easier to maintain)

ddeboer:
Okay, just did! Sorry for the confusion.

dbu:
thanks david. no worries, i think we can live with the occasional oddity in
the git history like two commits for the same change. merge is good, then
we are sure we dont diverge

PR #308, opened 2016-08-22
Add multiplexer client

thePanz:
Define a MultiplexerClient to handle multiple Proxy clients

### DIFF
--- a/doc/proxy-clients.rst
+++ b/doc/proxy-clients.rst
@@ -5,6 +5,8 @@ This library ships with clients for the Varnish and NGINX caching servers and
 the Symfony built-in HTTP cache. A Noop client that implements the interfaces
 but does nothing at all is provided for local development and testing purposes.
 
+A Multiplexer client is also available to forward calls to multiple proxy clients.
+
 The recommended usage is to have your application interact with the
 :doc:`cache invalidator <cache-invalidator>` which you set up with the proxy
 client suitable for the proxy server you use.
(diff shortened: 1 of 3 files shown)


src/ProxyClient/MultiplexerClient.php
```diff
@@ -0,0 +1,141 @@
...
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\Exception\ExceptionCollection;
+use FOS\HttpCache\Exception\InvalidArgumentException;
+use FOS\HttpCache\ProxyClient\Invalidation\BanInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\PurgeInterface;
+use FOS\HttpCache\ProxyClient\Invalidation\RefreshInterface;
+
+/**
+ * This class forwards invalidation to all attached clients.
```
dbu: can you add an `@author` line with your name here?

doc/proxy-clients.rst
```diff
@@ -136,6 +138,29 @@ cache invalidator optional in your code, you can (based on the environment)
...
+
+    use FOS\HttpCache\ProxyClient\MultiplexerClient;
+    use FOS\HttpCache\ProxyClient\Nginx;
+    use FOS\HttpCache\ProxyClient\Symfony;
+
+    $nginxClient = new Nginx($servers);
+    $symfonyClient = new Symfony([...]);
+    $client = new MultiplexerClient([$nginxClient, $symfonyClient]);
+
+.. note::
```
dbu: instead of making this a note, lets make this a code comment above the `new MultiplexerClient` line.

and say: "// Expects an array of ProxyClientInterface in the constructor"

dbu:
thanks! i wonder how we can say it, but in general having multiple layers of HTTP caching is not a good idea. maybe mention that its useful during transition? what other use cases do we have for it?

i guess that can be a `note` at the start of that chapter.

tests/Unit/ProxyClient/MultiplexerClientTest.php
```diff
@@ -0,0 +1,127 @@
...
+use FOS\HttpCache\ProxyClient\MultiplexerClient;
+use FOS\HttpCache\ProxyClient\ProxyClientInterface;
+
+class MultiplexerClientTest extends \PHPUnit_Framework_TestCase
+{
+    public function testBan()
+    {
+        $headers = ['Header1' => 'Header1-Value'];
+
+        $mockClient1 = $this->createMock(BanInterface::class);
```
dbu: we still support php 5.5 and that means phpunit 4.* which means no createMock :-(

can you please use the old getMock instead?

doc/proxy-clients.rst
```diff
@@ -5,6 +5,13 @@ This library ships with clients for the Varnish and NGINX caching servers and
 the Symfony built-in HTTP cache. A Noop client that implements the interfaces
 but does nothing at all is provided for local development and testing purposes.
 
+A Multiplexer client is also available to forward calls to multiple proxy clients.
+
+.. note::
+
+    Having multiple layers of HTTP caches in place is usually not a good idea, the
+    Multiplexer client is useful during transition phases.
```
dbu: please move the note into the section on the multiplexer client

thePanz: Done

dbu:
thanks a lot emanuele!

PR #311, opened 2016-09-29
feat(client): adds a noop client which implements all interfaces but …

gsdevme:
This was merged into master, but Im not sure what the plan for master is currently. Is it possible to have this added into 1.4? 

Edit: Ugh.. seems its not so easy.

### DIFF
--- a/src/ProxyClient/Noop.php
+++ b/src/ProxyClient/Noop.php
@@ -1,0 +1,90 @@
+<?php
+
+/*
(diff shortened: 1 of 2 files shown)


dbu:
i really hope i can wrap up the 2.0 release in the coming weeks. there needs to be one more iteration for the httplug integration, but i have something cooking for that, and once that is in its mainly about testing and validating we have no regressions.

PR #312, opened 2016-10-07
refactor http abstraction

### LINKED ISSUE #301: [doc] client setup with httplug
the doc is still talking about guzzle, update for httplug
http://foshttpcache.readthedocs.io/en/latest/proxy-clients.html

### LINKED ISSUE #305: how to HttpAdapter?
we currently have [HttpAdapter](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/Http/HttpAdapter.php) that creates requests and sends them through a Httplug client. could a Httplug plugin duplicate requests for each caching proxy or is that not something that not feasable? any other solutions or architecture ideas?

probably we should at least not instantiate the adapter in the AbstractProxyClient https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/AbstractProxyClient.php#L85

dbu:
fix #305, fix #301

the proxy clients know what they want to invalidate. the logic how to send the requests to each of the caching reverse proxy servers is now isolated in HttpAdapter, along with queuing the requests and sending them out in parallel on flush.

* [x] Clean up tests
* [x] Update documentation
* [x] Convert HTTP errors to exceptions in client


tests/Unit/ProxyClient/VarnishTest.php
```diff
@@ -11,22 +11,40 @@
...
+        $varnish = new Varnish($this->client);
+        $this->client->expects($this->once())
+            ->method('invalidate')
+        ;
+        $this->client->expects($this->once())
+            ->method('flush')
+        ;
         $varnish->ban([])->flush();
 
+        /*
```
dbu: clean up this class

dbu:
thanks for the input. you are right, i will refactor to use the RequestFactory instead.

but the overall architecture seems right to you?

src/ProxyClient/Http/HttpAdapter.php
```diff
@@ -254,6 +270,13 @@ private function setBaseUri($uriString = null)
      */
     private function filterUri($uriString, array $allowedParts = [])
     {
+        if (!is_string($uriString)) {
+            throw new \InvalidArgumentException(sprintf(
+                'Uri parameter must be a string, %s given',
```
sagikazarmark: Just for consistency: use URI or Uri

src/ProxyClient/Http/HttpAdapter.php
```diff
@@ -254,6 +270,13 @@ private function setBaseUri($uriString = null)
...
+                gettype($uriString)
```
sagikazarmark: Maybe output object instance and/or resource types too? I saw a oneliner for that somewhere

sagikazarmark:
Found a few more things.

I like the architecture. There is one more things which is not critical, but this is the time to think about it once more: I would try to find a better name for HttpAdapter, because it's not really an adapter. Maybe HttpBridge?

dbu:
naming things... what about calling it HttpInvalidator?

sagikazarmark:
Does it do any invalidation? As far as I saw it's a bridge between HTTP Client and ProxyClient...hence the Bridge idea. :smile: 

If it has any more behaviour, it's fine too.

dbu:
it accepts PSR 7 requests, queues them and then sends them to a list of servers. while sending them out, it also completes them with host header and sends them to the right ips.

the requests are meant to be invalidation requests, but could be other things. so i am not sure if bridge is the right word. the fact that this is HTTP is also known in the proxy client implementations.

src/ProxyClient/Http/HttpAdapter.php
dbu: don't think it will help much, when it has to be a string and nothing else

sagikazarmark:
Hm.... HttpDispatcher, HttpProxynator, HttpInvalidationBroker, HttpConnector?

HttpInvalidator is not actually true then as it doesn't do invalidation, it's the server that does it.

dbu:
dispatcher sounds good! thanks for the help mark.

sagikazarmark:
Anytime, although Proxynator was my personal favourite. :smile:

dbu:
> Proxynator 

maybe we should rename the library to FOSHttpProxynator :-)

PR #320, opened 2016-10-25
extract shared assertions into traits

### LINKED ISSUE #213: Reduce code duplication in tests

dbu:
fix #213

i opted to not put these traits into the src/Test folder so that we don't need to respect BC with them. i did not call the methods "testX" to avoid that they would be executed automatically.


ddeboer:
Nice cleanup, thanks!

PR #322, opened 2016-10-26
Add Varnish 5 compatibility

### LINKED ISSUE #319: Varnish 5 compatibility
[Varnish 5](https://www.varnish-cache.org/docs/trunk/whats-new/index.html) was released last month. “All users on 4.1 should be able to upgrade to this version without trouble”: this sounds hopeful. :wink:
- [ ] test against Varnish 5 on Travis
- [ ] make necessary changes to the VCLs that we ship (if necessary)
- [ ] update docs

ddeboer:
Fix #319.
- Make VCL compatible with Varnish 5.
- Move Varnish 4/5 files to varnish.
- Keep varnish-3 for now.
- Upgrade Travis config to Trusty/edge to get rid of
    /usr/lib/varnish/libvarnish.so: undefined symbol: pcre_free_study
- Update documentation.


dbu:
it looks like we have some missing dependency for varnish 5. could we use the same code as for varnish 4 with adding the repo instead of wget the .deb file?

ddeboer:
> it looks like we have some missing dependency for varnish 5. could we use the same code as for varnish 4 with adding the repo instead of wget the .deb file?

Unfortunately not: Varnish 5 is not (yet?) available [in a repo](https://www.varnish-cache.org/releases/install_debian.html#install-debian).

.travis.yml
```diff
@@ -25,10 +27,18 @@ install:
...
+      wget https://repo.varnish-cache.org/pkg/5.0.0/varnish_${VARNISH_VERSION}.0-1_amd64.deb
+      sudo dpkg -i varnish_${VARNISH_VERSION}.0-1_amd64.deb
+      sudo apt-get install -f libpcre3
+    else
+      curl http://repo.varnish-cache.org/debian/GPG-key.txt | sudo apt-key add -
+      echo "deb http://repo.varnish-cache.org/ubuntu/ precise varnish-${VARNISH_VERSION}" | sudo tee -a /etc/apt/sources.list
+      sudo apt-get update -qq
+      sudo apt-get install -qq varnish
+    fi
+  - sudo apt-get update -qq && apt-get install -qq python-sphinx enchant
```
dbu: you lost the other sudo here. maybe just make that two lines again, no need to chain the two things with &&.

or move the update into the varnish 5 case

dbu:
ah so they build the .deb but don't provide it in a repository. hm, whatever.

and ignore my last comment, the `apt-get install` without package name is a good idea!

dbu:
looks like their .deb does not properly specify all dependencies: `/usr/sbin/varnishd: symbol lookup error: /usr/lib/varnish/libvarnish.so: undefined symbol: pcre_free_study` :-(

dbu:
seems to not have helped. not sure if that is a build error or if the method is part of something else. or if we end up with a too old version. maybe the varnish people have an idea?

ddeboer:
Yeah, this is getting messy. 😛

dbu:
do we have any way to get some help from the varnish people? as in checking their .deb file and fixing dependencies (or figuring out whether pcre is just wrongly linked?)

ddeboer:
Made VCL Varnish 5-compatible.

I was able to install Varnish 5 into a local Ubuntu 12.04 Docker container, so I’ll give it one more try to get it running on Travis.

ddeboer:
For future reference: `/usr/lib/varnish/libvarnish.so: undefined symbol: pcre_free_study` is caused by a libpcre3 version older than 8.13; Ubuntu 12.04 has [8.12](http://packages.ubuntu.com/precise/libpcre3). 

I fixed this by upgrading our Travis setup to Trusty (14.04), which has a newer libpcre3.

doc/conf.py
```diff
@@ -263,5 +263,5 @@ def setup(app):
 
 config_block = {
     'varnish3': 'Varnish 3',
-    'varnish4': 'Varnish 4'
+    'varnish4': 'Varnish 4 & 5'
```
ddeboer: This renames the tab to `Varnish 4 & 5`. When Varnish 6 comes along, we can hopefully drop Varnish 3 support to keep the tabs manageable. :wink:

.travis.yml
```diff
@@ -25,10 +29,19 @@ install:
...
+      wget https://repo.varnish-cache.org/pkg/${VARNISH_VERSION}/varnish_${VARNISH_VERSION}-1_amd64.deb
+      sudo dpkg -i varnish_${VARNISH_VERSION}-1_amd64.deb
+      sudo apt-get -f install
+    else
+      curl http://repo.varnish-cache.org/debian/GPG-key.txt | sudo apt-key add -
+      echo "deb http://repo.varnish-cache.org/ubuntu/ precise varnish-${VARNISH_VERSION}" | sudo tee -a /etc/apt/sources.list
+      sudo apt-get update -qq
+      sudo apt-get install -qq varnish
+    fi
   - sudo apt-get update -qq
```
dbu: can we move this to the if case before the `apt-get -f install` line, to not do it twice when having an older varnish?

doc/conf.py
```diff
@@ -52,7 +52,7 @@
 
 # General information about the project.
 project = 'FOSHttpCache'
-copyright = '2014, FriendsOfSymfony'
+copyright = u'2014–2016, FriendsOfSymfony'
```
dbu: why the "u"? could we just remove the year as its not covered+updated by styleci?

ddeboer: > why the "u"? 

This makes it a Unicode string (in Python/Sphinx). We need that for the dash, as it’s an [EN dash](http://www.thepunctuationguide.com/en-dash.html), not a hyphen, between a range of numbers. Sorry about that, I’m a typography nerd. :wink:

>  could we just remove the year as its not covered+updated by styleci?

Does StyleCI update the years somewhere?

dbu: yes, in the php files we have a file header configured, and then we can just change the year in styleci to have it updated everywhere

ddeboer: Ah, does that work for .rst, too? Our [current docs](http://foshttpcache.readthedocs.io/en/stable/) still have 2014.

dbu:
we should add the folder rename to the changelog i think. apart from that looks ready to merge!

dbu:
styleci does not know about years, its just a static text field for the
file headers. we probably better don't have years in other places at
all, to avoid that they get outdated.

on the source code, it seems to be some legal mumbo-jumbo, but for the
doc i don't expect it to be that relevant.

anyways, we can merge this PR regardless, only thing left to do imho is
adjusting the changelog

ddeboer:
Updated change log.

dbu:
great, thanks a lot!

.travis.yml
```diff
@@ -25,10 +29,19 @@ install:
...
-  - sudo apt-get install -qq varnish python-sphinx enchant
+  - |
+    if [ "$VARNISH_VERSION" = "5.0.0" ]; then
+      wget https://repo.varnish-cache.org/pkg/${VARNISH_VERSION}/varnish_${VARNISH_VERSION}-1_amd64.deb
+      sudo dpkg -i varnish_${VARNISH_VERSION}-1_amd64.deb
+      sudo apt-get update -qq
+      sudo apt-get -f install
+    else
+      curl http://repo.varnish-cache.org/debian/GPG-key.txt | sudo apt-key add -
+      echo "deb http://repo.varnish-cache.org/ubuntu/ precise varnish-${VARNISH_VERSION}" | sudo tee -a /etc/apt/sources.list
```
stof: should not be ``precise`` here but ``trusty``, as you switched to trusty.

.travis.yml
```diff
@@ -25,10 +29,19 @@ install:
...
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq varnish python-sphinx enchant
+  - |
+    if [ "$VARNISH_VERSION" = "5.0.0" ]; then
+      wget https://repo.varnish-cache.org/pkg/${VARNISH_VERSION}/varnish_${VARNISH_VERSION}-1_amd64.deb
+      sudo dpkg -i varnish_${VARNISH_VERSION}-1_amd64.deb
+      sudo apt-get update -qq
+      sudo apt-get -f install
+    else
+      curl http://repo.varnish-cache.org/debian/GPG-key.txt | sudo apt-key add -
```
stof: using HTTPS would be better

.travis.yml
```diff
@@ -25,10 +29,19 @@ install:
...
+      echo "deb http://repo.varnish-cache.org/ubuntu/ precise varnish-${VARNISH_VERSION}" | sudo tee -a /etc/apt/sources.list
+      sudo apt-get update -qq
+      sudo apt-get install -qq varnish
```
stof: the way this is written, the build will not properly fail in case one of these commands fail, as they are not commands executed by Travis itself, and bash does not run in ``-e`` mode by default.

doc/conf.py
stof: should the key be renamed to ``varnish`` too (changing the doc too) as it is not only about varnish 4 anymore ?

dbu:
thanks a lot stof, agree on all your comments. created #323 to adjust these things.

PR #329, opened 2016-11-04
Prepare release 2.0.0

ddeboer:
Merge this just before tagging 2.0.0-alpha1.

- [ ] check [diff](https://github.com/FriendsOfSymfony/FOSHttpCache/compare/1.4.2...master) for anything we didn’t yet add to the changeling.

PR #330, opened 2016-11-05
Anonymous user context lookup

dbu:
always send a hash lookup request, even for anonymous users.

the symfony HttpCache implementation allows to configure a hardcoded hash for anonymous users if that is desired, but defaults to do a hash lookup.

alternative to #318 

cc @lolautruche @andrerom @bdunogier

### DIFF
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -180,10 +180,10 @@ based on session cookies or authorization headers. If the default settings are
 right for you, you don't need to do anything more. You can customize a number of
 options through the constructor:
 
-* **anonymous_hash**: Hash used for anonymous user. This is a performance
-  optimization to not do a backend request for users that are not logged in.
-  By default, the header is skipped. If you specify a header, that header is
-  used.
+* **anonymous_hash**: Hard-coded hash to use for anonymous users. This is a
+  performance optimization to not do a backend request for users that are not
+  logged in. If you specify a non-empty value for this field, that is used as
+  context hash header instead of doing a hash lookup for anonymous users.
 
 * **user_hash_accept_header**: Accept header value to be used to request the
   user hash to the backend application. Must match the setup of the backend
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -182,6 +182,8 @@ options through the constructor:
 
 * **anonymous_hash**: Hash used for anonymous user. This is a performance
   optimization to not do a backend request for users that are not logged in.
+  By default, the header is skipped. If you specify a header, that header is
+  used.
 
 * **user_hash_accept_header**: Accept header value to be used to request the
   user hash to the backend application. Must match the setup of the backend
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Moved Varnish 4 and 5 configuration files from `resources/config/varnish-4/`
   to `resources/config/varnish/`.
 * Changed default Varnish version to 5.
+* Removed special case for anonymous users in user context behaviour. Varnish
+  now does a hash lookup for anonymous users as well.
 
 ### NGINX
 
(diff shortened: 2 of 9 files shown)


andrerom:
WDYT @lolautruche @bdunogier ?

andrerom:
Ok, seems we are +1

But does this make sense from your perspective as well @dbu ?

doc/symfony-cache-configuration.rst
```diff
@@ -182,6 +182,8 @@ options through the constructor:
 
 * **anonymous_hash**: Hash used for anonymous user. This is a performance
   optimization to not do a backend request for users that are not logged in.
+  By default, the header is skipped. If you specify a header, that header is
+  used.
```
ddeboer: This is a bit unclear. What’s the header you’re talking about being skipped or used?

ddeboer:
> Ok, seems we are +1

Does that mean you have spoken with @lolautruche and @bdunogier? :wink:

> But does this make sense from your perspective as well @dbu ?

I have one argument against this change (which does a hash lookup for requests even if they have no Cookie/Authorization header) and in favour of #318 (which doesn’t): performance. Usually HTTP caching is optimised for the largest group of users. Arguably, that’s usually the anonymous users. Changing defaults to do a hash lookup (and hit Symfony) for anonymous users will degrade performance.

A couple of solutions:
1. Follow #318 instead.
2. In the docs, offer both VCL variants.
3. In the docs, emphasise the importance of caching the anonymous hash request.

dbu:
> 3. In the docs, emphasise the importance of caching the anonymous hash
> request.

i will work on 3. if the hash is cached, it should be one additional
request per cache lifetime, which is a very small price to pay for
consistency and avoiding errors.

lolautruche:
Hi

Sorry but I think I'm missing something here. Why do you want to remove the anonymous hash? I don't see the point... Anonymous hash is here for performance if you don't have any session. Difference will be quite big, even if the hash is cached.

However I understand interrogations people may have about this *anonymous-no-session* hash, so there's more a need of documentation.
And AFAICT, there was an anonymous hash both for SymfonyCache and Varnish.

doc/symfony-cache-configuration.rst
dbu: thats the old doc, sorry. it should say "By default, a hash lookup is done. If you specify a hash, no hash lookup for anonymous users occurs and the specified hash is used."

dbu:
one important reason is https://github.com/FriendsOfSymfony/FOSHttpCache/pull/318#issuecomment-258450706 - if you change the permission set of anonymous, a hardcoded hash will not change and thus the cache could be wrong.

its in general confusing when the hash lookup sometimes does not happen. if we cache the lookup, i don't see the cost. the hash lookup response is shared between all anonymous users, so its done only once.

lolautruche:
Right, but what will be the cache key? AFAIR for session based requests, it was the cookie header. Is it still the case?
Sorry I didn't look into all that for quite a long time 😊

dbu:
we use whatever the backend tells that the hash depends on. which by default are the headers "Authentication" and "Cookies". you need to use vcl to clean up the cookie and only keep the cookies relevant to the backend when using the context hash (otherwise it also won't work for logged in users).

lolautruche:
Right, so there should not be any problem.

+1 for me

dbu:
great, thanks @lolautruche - there was a lot of back and forth with this, glad we have a solution that looks ok for everyone.

ddeboer:
So we just need some small adjustments to the doc, @dbu, and we can merge this! 🌟

dbu:
i just updated the doc. can you check again if that makes sense? then lets get this merge, merge #329 and tag 2.0.0-alpha1 !

dbu:
regarding making it clear that the hash lookup needs to be cached: http://foshttpcache.readthedocs.io/en/latest/user-context.html#caching-the-hash-response is pretty clear to me. anything we could add? or any other place we should point this out?

ddeboer:
Hm tests are now failing... 😠

ddeboer:
Amended the commit to fix the build, hang on...

PR #332, opened 2016-11-11
Add support for Varnish xkey

andrerom:
xkey promises to have vastly better performance than invalidation by BAN for two reasons:
- It keeps a hash table of the tags around so it can efficiently do invalidation directly instead of filling up a list of regex bans on request data slowing things down
- It supports soft purge making it able to take into account grace time

It can be installed on any supported platform by Varnish 4.1+ as long as people have src files for Varnish, and  is also available on some Linux distros as package.  And besides that, it can be used with [Varnish Plus](http://book.varnish-software.com/4.0/chapters/Cache_Invalidation.html#hashtwo-xkey-varnish-software-implementation-of-surrogate-keys).

Further reading:
- Doc: https://github.com/varnish/varnish-modules/blob/master/docs/vmod_xkey.rst
- Wikipedia: https://wikitech.wikimedia.org/wiki/XKey

# Doc draft

Configuring xkey:
```php
[
    'tags_header' => 'xkey-softpurge',// Or 'xkey-purge' if you can not use grace, see VCL
    'tag_mode' => 'purgekeys',
]
```

# Todo:
- [x] Compile xkey on travis
  - [x] ~~Figure out why the _(added)_ Varnish 5.2 testing job fails~~ _(move 5.2 support to other pr)_
- [x] Add unit test coverage
- [x] Add functional test coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
     - php: 5.6
     - php: 7.0
     - php: 7.1
-      env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1 DEPENDENCIES="symfony/lts:^3"
+      env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1
     - php: 7.2
       env: DOCCHECK=true COVERAGE=true PHPUNIT_FLAGS="-v --coverage-text"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: required
 language: php
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: required
 language: php
 
 cache:
@@ -19,6 +20,7 @@ matrix:
     - php: 5.6
     - php: 7.0
     - php: 7.1
+      env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1 DEPENDENCIES="symfony/lts:^3"
     - php: 7.2
       env: DOCCHECK=true COVERAGE=true PHPUNIT_FLAGS="-v --coverage-text"
 
@@ -28,7 +30,7 @@ matrix:
 
       # Test Symfony LTS versions
     - php: 7.2
-      env: DEPENDENCIES="symfony/lts:^2"
+      env: VARNISH_VERSION=5.1 VARNISH_MODULES_VERSION=0.12.1 DEPENDENCIES="symfony/lts:^2"
     - php: 7.2
       env: DEPENDENCIES="symfony/lts:^3 toflar/psr6-symfony-http-cache-store:^1.0"
 
@@ -64,6 +66,7 @@ before_script:
     sudo apt-get install -qq varnish
   - if [ "$DOCCHECK" = true ]; then sudo apt-get install -qq python-sphinx enchant; fi
   - if [ "$DOCCHECK" = true ]; then sudo pip install -r doc/requirements.txt; fi
+  - if [ "$VARNISH_MODULES_VERSION" != "" ]; then sh ./tests/install-varnish-modules.sh; fi
   # Install NGINX
   - sh ./tests/install-nginx.sh
 
--- a/doc/proxy-clients.rst
+++ b/doc/proxy-clients.rst
@@ -101,11 +101,23 @@ You can also pass some options to the Varnish client:
   which tags to invalidate when sending invalidation requests to the caching
   proxy. Make sure that your :ref:`Varnish configuration <varnish_tagging>`
   corresponds to the header used here;
+* ``tag_mode`` (default: ban): Makes it possible to invalidate tags
+  using `PURGEKEYS` with xkey instead of normal `BAN`. Supported
+  modes: ``ban`` and ``purgekeys``.
 * ``header_length`` (default: 7500): Control the maximum header length when
   invalidating tags. If there are more tags to invalidate than fit into the
   header, the invalidation request is split into several requests;
 * ``default_ban_headers`` (default: []): Map of headers that are set on each
   ban request, merged with the built-in headers.
+
+Example for configuring xkey::
+
+    $options = [
+        'tags_header' => 'xkey-softpurge',// Or 'xkey-purge' if you can not use grace, see VCL
+        'tag_mode' => 'purgekeys'
+    ];
+
+    $varnish = new Varnish($httpDispatcher, $options);
 
 Additionally, you can specify the request factory used to build the
 invalidation HTTP requests. If not specified, auto discovery is used – which
--- a/doc/proxy-clients.rst
+++ b/doc/proxy-clients.rst
@@ -110,15 +110,6 @@ You can also pass some options to the Varnish client:
 * ``default_ban_headers`` (default: []): Map of headers that are set on each
   ban request, merged with the built-in headers.
 
-Example for configuring xkey::
-
-    $options = [
-        'tags_header' => 'xkey-softpurge',// Or 'xkey-purge' if you can not use grace, see VCL
-        'tag_mode' => 'purgekeys'
-    ];
-
-    $varnish = new Varnish($httpDispatcher, $options);
-
 Additionally, you can specify the request factory used to build the
 invalidation HTTP requests. If not specified, auto discovery is used – which
 usually is fine.
--- a/doc/varnish-configuration.rst
+++ b/doc/varnish-configuration.rst
@@ -33,29 +33,6 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
     trigger invalidation are whitelisted here. Otherwise, lost cache invalidation
     requests will lead to lots of confusion.
 
-Better tag invalidation: xkey
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. _varnish_xkey_intro:
-
-Since Varnish 4.1 you can use the official [xkey Varnish module] for better
-performance with cache tags.
-
-As explained by [Varnish Software](http://book.varnish-software.com/4.0/chapters/Cache_Invalidation.html#hashtwo-xkey-varnish-software-implementation-of-surrogate-keys):
-> (..), hashtwo/xkey is much more efficient than bans because of two reasons:
-> 1) looking up hash keys is much more efficient than traversing ban-lists, and
-> 2) every time you test a ban expression, it checks every object in the cache that is
-> older than the ban itself.
-
-It also has more predicable performance thanks to softpurge support on tags for grace.
-
-Minimum version of `varnish-modules` is v0.10.2*. It's part of Ubuntu 17.10 ("Artful")
-and higher. But can also be [installed on any other linux platform](https://github.com/varnish/varnish-modules#installation).
-
-* _v0.10.2 is the first version introducing support for purging several tags at once.
-You can technically use earlier versions like v0.9.1, however then you will not be
-able to invalidate more then one tag at a time._
-
 Provided VCL Subroutines
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -108,7 +85,6 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 
 Read more on `handling PURGE requests`_ in the Varnish documentation (for
 Varnish 3, see `purging for Varnish 3`_).
-
 
 Refresh
 ~~~~~~~
@@ -198,14 +174,29 @@ Tagging
 
 Feature: :ref:`cache tagging <tags>`
 
-For this feature you'll need to choose between either BAN _(default)_, or
-:ref:`xkey based tags invalidation <varnish_xkey_intro>`.
+For this feature you'll need to choose between either BAN _(default)_, or the more
+performant xkey based tag system.
 
 Using xkey
 ^^^^^^^^^^
 
-For an advanced ``xkey`` based setup you can use ``fos_tags_xkey.vcl`` and
-:ref:`configure Varnish Client for xkey <varnish_custom_tags_header>`.
+Since Varnish 4.1 you can use the official `xkey Varnish module`_ for better
+performance with cache tags.
+
+As explained in the `Cache Invalidation chapter of the Varnish documentation`_::
+> (..), hashtwo/xkey is much more efficient than bans because of two reasons:
+> 1) looking up hash keys is much more efficient than traversing ban-lists, and
+> 2) every time you test a ban expression, it checks every object in the cache that is
+> older than the ban itself.
+
+With xkey, you can also soft purge tagged responses to allow for a grace period where
+invalidated (stale) content is still delivered to the client.
+
+Minimum version of `varnish-modules` is v0.10.2*. It's part of Ubuntu 17.10 ("Artful")
+and higher, otherwise see their documentation for `installing xkey on other platforms`_.
+
+For ``xkey`` setup you can use ``fos_tags_xkey.vcl`` and
+:ref:`configure Varnish Client for xkey <varnish_custom_tags_header>` as shown below.
 
 Subroutines are provided in ``resources/config/varnish-[version]/fos_tags_xkey.vcl``.
 To enable this feature, add the following to ``your_varnish.vcl``:
@@ -231,6 +222,10 @@ Secondly we'll also need to configure Varnish Proxy client for xkey::
     ];
 
     $varnish = new Varnish($httpDispatcher, $options);
+
+\* _v0.10.2 is the first version introducing support for purging several tags at once.
+You can technically use earlier versions down to 0.9.x releases, however then you will not be
+able to invalidate more then one tag at a time._
 
 Using BAN
 ^^^^^^^^^
@@ -467,5 +462,7 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 .. _explained in the Varnish documentation: https://www.varnish-cache.org/trac/wiki/VCLExampleRemovingSomeCookies#RemovingallBUTsomecookies
 .. _curl Varnish plugin: https://github.com/varnish/libvmod-curl
 .. _xkey Varnish module: https://github.com/varnish/varnish-modules/blob/master/docs/vmod_xkey.rst
+.. _Cache Invalidation chapter of the Varnish documentation: http://book.varnish-software.com/4.0/chapters/Cache_Invalidation.html#hashtwo-xkey-varnish-software-implementation-of-surrogate-keys
+.. _installing xkey on other platforms: https://github.com/varnish/varnish-modules#installation
 .. _`builtin VCL`: https://github.com/varnishcache/varnish-cache/blob/5.0/bin/varnishd/builtin.vcl
 .. _`default VCL`: https://github.com/varnishcache/varnish-cache/blob/3.0/bin/varnishd/default.vcl
--- a/doc/varnish-configuration.rst
+++ b/doc/varnish-configuration.rst
@@ -43,7 +43,7 @@ performance with cache tags.
 
 As explained by [Varnish Software](http://book.varnish-software.com/4.0/chapters/Cache_Invalidation.html#hashtwo-xkey-varnish-software-implementation-of-surrogate-keys):
 > (..), hashtwo/xkey is much more efficient than bans because of two reasons:
-> 1)looking up hash keys is much more efficient than traversing ban-lists, and
+> 1) looking up hash keys is much more efficient than traversing ban-lists, and
 > 2) every time you test a ban expression, it checks every object in the cache that is
 > older than the ban itself.
 
@@ -53,8 +53,8 @@ Minimum version of `varnish-modules` is v0.10.2*. It's part of Ubuntu 17.10 ("Ar
 and higher. But can also be [installed on any other linux platform](https://github.com/varnish/varnish-modules#installation).
 
 * _v0.10.2 is the first version introducing support for purging several tags at once.
- You can technically use earlier versions like v0.9.1, however then you will not be
- able to invalidate more then one tag at a time._
+You can technically use earlier versions like v0.9.1, however then you will not be
+able to invalidate more then one tag at a time._
 
 Provided VCL Subroutines
 ~~~~~~~~~~~~~~~~~~~~~~~~
--- a/doc/varnish-configuration.rst
+++ b/doc/varnish-configuration.rst
@@ -33,6 +33,27 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
     trigger invalidation are whitelisted here. Otherwise, lost cache invalidation
     requests will lead to lots of confusion.
 
+Better tag invalidation: xkey
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Since Varnish 4.1 you can use the official [xkey Varnish module] for better
+performance with cache tags.
+
+As explained by [Varnish Software](http://book.varnish-software.com/4.0/chapters/Cache_Invalidation.html#hashtwo-xkey-varnish-software-implementation-of-surrogate-keys):
+> (..), hashtwo/xkey is much more efficient than bans because of two reasons:
+> 1)looking up hash keys is much more efficient than traversing ban-lists, and
+> 2) every time you test a ban expression, it checks every object in the cache that is
+> older than the ban itself.
+
+It also has more predicable performance thanks to softpurge support on tags for grace.
+
+Minimum version of `varnish-modules` is v0.10.2*. It's part of Ubuntu 17.10 ("Artful")
+and higher. But can also be [installed on any other linux platform](https://github.com/varnish/varnish-modules#installation).
+
+* _v0.10.2 is the first version introducing support for purging several tags at once.
+ You can technically use earlier versions like v0.9.1, however then you will not be
+ able to invalidate more then one tag at a time._
+
 Provided VCL Subroutines
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
@@ -85,6 +106,7 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 
 Read more on `handling PURGE requests`_ in the Varnish documentation (for
 Varnish 3, see `purging for Varnish 3`_).
+
 
 Refresh
 ~~~~~~~
@@ -173,6 +195,30 @@ Tagging
 ~~~~~~~
 
 Feature: :ref:`cache tagging <tags>`
+
+For this feature you'll need to choose between either xkey, or BAN based tags invalidation.
+
+Using xkey
+^^^^^^^^^^
+
+For an advanced ``xkey`` based setup you can use ``fos_tags_xkey.vcl`` and
+:ref:`configure Varnish Client for xkey <varnish_custom_tags_header>`.
+
+Subroutines are provided in ``resources/config/varnish-[version]/fos_tags_xkey.vcl``.
+To enable this feature, add the following to ``your_varnish.vcl``:
+
+.. configuration-block::
+
+    .. code-block:: varnish
+
+        include "path-to-config/varnish/fos_tags_xkey.vcl";
+
+        sub vcl_recv {
+            call fos_tags_xkey_recv;
+        }
+
+Using BAN
+^^^^^^^^^
 
 If you have included ``fos_ban.vcl``, tagging will be automatically enabled
 with the ``X-Cache-Tags`` header for both marking the tags on the response and
@@ -405,5 +451,6 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 .. _ban lurker: https://www.varnish-software.com/blog/ban-lurker
 .. _explained in the Varnish documentation: https://www.varnish-cache.org/trac/wiki/VCLExampleRemovingSomeCookies#RemovingallBUTsomecookies
 .. _curl Varnish plugin: https://github.com/varnish/libvmod-curl
+.. _xkey Varnish module: https://github.com/varnish/varnish-modules/blob/master/docs/vmod_xkey.rst
 .. _`builtin VCL`: https://github.com/varnishcache/varnish-cache/blob/5.0/bin/varnishd/builtin.vcl
 .. _`default VCL`: https://github.com/varnishcache/varnish-cache/blob/3.0/bin/varnishd/default.vcl
--- a/doc/varnish-configuration.rst
+++ b/doc/varnish-configuration.rst
@@ -220,6 +220,18 @@ To enable this feature, add the following to ``your_varnish.vcl``:
             call fos_tags_xkey_recv;
         }
 
+
+Secondly we'll also need to configure Varnish Proxy client for xkey::
+
+    use FOS\HttpCache\ProxyClient\Varnish;
+
+    $options = [
+        'tags_header' => 'xkey-softpurge',// Or 'xkey-purge' if you can not use grace, see VCL
+        'tag_mode' => 'purgekeys'
+    ];
+
+    $varnish = new Varnish($httpDispatcher, $options);
+
 Using BAN
 ^^^^^^^^^
 
--- a/doc/varnish-configuration.rst
+++ b/doc/varnish-configuration.rst
@@ -36,6 +36,8 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
 Better tag invalidation: xkey
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+.. _varnish_xkey_intro:
+
 Since Varnish 4.1 you can use the official [xkey Varnish module] for better
 performance with cache tags.
 
@@ -196,7 +198,8 @@ Tagging
 
 Feature: :ref:`cache tagging <tags>`
 
-For this feature you'll need to choose between either xkey, or BAN based tags invalidation.
+For this feature you'll need to choose between either BAN _(default)_, or
+:ref:`xkey based tags invalidation <varnish_xkey_intro>`.
 
 Using xkey
 ^^^^^^^^^^
--- a/resources/config/varnish/fos_tags_xkey.vcl
+++ b/resources/config/varnish/fos_tags_xkey.vcl
@@ -15,13 +15,13 @@ sub fos_tags_xkey_recv {
             return (synth(405, "Not allowed"));
         }
 
-        // If tag invalidation api allows caller to specify expire/purge in the future,
-        // then adjust this to be able to handle both headers in same request.
+        # Based on provided header invalidate (purge) or expire (softpurge) the tagged content
         if (req.http.xkey-purge) {
             set req.http.n-gone = xkey.purge(req.http.xkey-purge);
         } elseif (req.http.xkey-softpurge) {
             set req.http.n-gone = xkey.softpurge(req.http.xkey-softpurge);
         } else {
+            # If neither of the headers are provided we return 400 so you can detect wrong configuration
             return (synth(400));
         }
 
--- a/resources/config/varnish/fos_tags_xkey.vcl
+++ b/resources/config/varnish/fos_tags_xkey.vcl
@@ -1,0 +1,30 @@
+/*
+ * This file is part of the FOSHttpCache package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+import xkey;
+
+sub fos_tags_xkey_recv {
+    if (req.method == "PURGEKEYS") {
+        if (!client.ip ~ invalidators) {
+            return (synth(405, "Not allowed"));
+        }
+
+        // If tag invalidation api allows caller to specify expire/purge in the future,
+        // then adjust this to be able to handle both headers in same request.
+        if (req.http.xkey-purge) {
+            set req.http.n-gone = xkey.purge(req.http.xkey-purge);
+        } elseif (req.http.xkey-softpurge) {
+            set req.http.n-gone = xkey.softpurge(req.http.xkey-softpurge);
+        } else {
+            return (synth(400));
+        }
+
+        return (synth(200, "Invalidated "+req.http.n-gone+" objects"));
+    }
+}
--- a/resources/config/varnish/fos_tags_xkey.vcl
+++ b/resources/config/varnish/fos_tags_xkey.vcl
@@ -15,16 +15,22 @@ sub fos_tags_xkey_recv {
             return (synth(405, "Not allowed"));
         }
 
-        # Based on provided header invalidate (purge) or expire (softpurge) the tagged content
-        if (req.http.xkey-purge) {
-            set req.http.n-gone = xkey.purge(req.http.xkey-purge);
-        } elseif (req.http.xkey-softpurge) {
-            set req.http.n-gone = xkey.softpurge(req.http.xkey-softpurge);
-        } else {
-            # If neither of the headers are provided we return 400 so you can detect wrong configuration
+        # If neither of the headers are provided we return 400 to simplify detecting wrong configuration
+        if (!req.http.xkey-purge && !req.http.xkey-softpurge) {
             return (synth(400));
         }
 
-        return (synth(200, "Invalidated "+req.http.n-gone+" objects"));
+        # Based on provided header invalidate (purge) and/or expire (softpurge) the tagged content
+        set req.http.n-gone = 0;
+        set req.http.n-softgone = 0;
+        if (req.http.xkey-purge) {
+            set req.http.n-gone = xkey.purge(req.http.xkey-purge);
+        }
+
+        if (req.http.xkey-softpurge) {
+            set req.http.n-softgone = xkey.softpurge(req.http.xkey-softpurge);
+        }
+
+        return (synth(200, "Purged "+req.http.n-gone+" objects, expired "+req.http.n-softgone+" objects"));
     }
 }
--- a/src/ProxyClient/Varnish.php
+++ b/src/ProxyClient/Varnish.php
@@ -64,25 +64,32 @@ class Varnish extends HttpProxyClient implements BanCapable, PurgeCapable, Refre
     public function invalidateTags(array $tags)
     {
         $tagMode = $this->options['tag_mode'];
+        $invalidationMethod = "{$tagMode}TagChunk";
         $escapedTags = array_map('preg_quote', $this->escapeTags($tags));
 
         $chunkSize = $this->determineTagsPerHeader($escapedTags, 'ban' === $tagMode ? '|' : ' ');
 
         foreach (array_chunk($escapedTags, $chunkSize) as $tagchunk) {
-            if ('ban' === $tagMode) {
-                $tagExpression = sprintf('(^|,)(%s)(,|$)', implode('|', $tagchunk));
-                $this->ban([$this->options['tags_header'] => $tagExpression]);
-            } else {// purgekeys
-                $this->queueRequest(
-                    self::HTTP_METHOD_PURGEKEYS,
-                    '/',
-                    [$this->options['tags_header'] => implode(' ', $tagchunk)],
-                    false
-                );
-            }
+            $this->$invalidationMethod($tagchunk);
         }
 
         return $this;
+    }
+
+    private function banTagChunk(array $tagchunk)
+    {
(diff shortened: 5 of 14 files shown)


ddeboer:
Very interesting! 👍 This is what was previously known as Varnish Plus’s Hash Ninja, right?

We should document that xkey requires Varnish 4.1+.

andrerom:
> This is what was previously known as Varnish Plus’s Hash Ninja, right?

Yes, [Surrogate keys (Fastly) => Hash Ninja (Varnish Plus) => Secondary Hash/Hash-Two (Varnish Plus) => xkey (open source vmod)](https://old.varnish-cache.org/utility/xkey-vmod-surrogate-keys-hash-ninja-secondary-hashhash-two) afaik.

resources/config/varnish/fos_purge.vcl
```diff
@@ -12,6 +12,17 @@ sub fos_purge_recv {
         if (!client.ip ~ invalidators) {
             return (synth(405, "Not allowed"));
         }
+
+        # If http header "xkey" is set, we assume purge is on key and you have Varnish xkey installed
+        if (req.http.xkey) {
```
andrerom: note: I'm using `xkey` here, however xkey doc suggests using `key`, or in older doc `xkey-purge`. 

As it only supports 1 key here, it can make sense to use a different header, however then we need a separate config option in `src/ProxyClient/Varnish.php`.

ddeboer: Definitely make this configurable in on the PHP side. That’s what we did for the tags header, too.

ddeboer:
We could add [this piece](http://book.varnish-software.com/4.0/chapters/Cache_Invalidation.html#hashtwo-xkey-varnish-software-implementation-of-surrogate-keys) to our docs:

> In addition, hashtwo/xkey is much more efficient than bans because of two reasons: 1) looking up hash keys is much more efficient than traversing ban-lists, and 2) every time you test a ban expression, it checks every object in the cache that is older than the ban itself.

dbu:
very interesting. i would like to postpone this to release 2.1 as i really want to get 2.0 out asap

but its very good if you can check if you see anything that would prevent introducing this in 2.1 - do we need any BC breaks to make it happen?

ddeboer:
Yeah, my thoughts exactly: this is post 2.0 stuff.

I don’t foresee any BC breaks; just *adding* configuration options. @andrerom Do you see any possible BC breaks?

andrerom:
I assumed 2.1 for this so on purpose made it not break anything, but opened it now to check if this should rather be done some other way which might break things.

So, if we go with this approach, then no breaks.

src/ProxyClient/Varnish.php
```diff
@@ -151,6 +160,8 @@ protected function configureOptions()
         $resolver->setDefaults([
             'tags_header' => self::DEFAULT_HTTP_HEADER_CACHE_TAGS,
             'header_length' => 7500,
+            'tags_invalidation_by_purge' => false,
+            'tags_separator' => ',',
```
dbu: is making this configurable related to the xkey change? what is the reason you want to do this, handle tags with `,` in them? i thought we remove the separator at some point from tags, to be sure.

if there is a strong argument for it, we probably should fix that bit before 2.0.

dbu:
great, good point to look at this now! i do like the general approach, yes.

src/ProxyClient/Varnish.php
andrerom: Xkey tag header format is list of tags separated by space, while the one used for ban now is comma separated, right?

That said, the other approach I had in mind was to change defaults to use xkey, and adapt BAN VCL and Symfony proxy store to understand that natively _(to simplify learning use of Tagging by making it consistent across approaches). But _that_ is a bc break, so a bit late for 2.0.

dbu: ah, okay. we used comma separated until now, yes. as ban is based on regex, it could be anything as long as escaping happens correctly. its all on php side, varnish just stores the header and runs a regex. (actually, if escaping does not happen, we only risk to ban too much, which is only a performance loss but should not be too likely to really happen and thus no big deal)

imho we can't change to xkey as default until we drop support for all varnish versions that come without xkey. but the doc can clearly state that if you are lucky enough to have a varnish recent enough that it supports xkey, you should switch to that.

i wonder however if we really want to make the separator configurable. i would prefer to check if xkey is enabled and if so pick the right separator. or simply switch to a whitespace as separator instead of the comma. actually i could see switching to whitespace even with ban, if you can do that for 2.0.0. afaik its not even specified in the doc how we separate so its really up to us.

resources/config/varnish/fos_purge.vcl
```diff
@@ -12,6 +12,17 @@ sub fos_purge_recv {
...
+            # By default we recommend using soft purge to respect grace time, if you need to hard purge use:
+            # set req.http.n-gone = xkey.purge(req.http.xkey);
+            set req.http.n-gone = xkey.softpurge(req.http.xkey);
```
dbu: don't you need a `use xkey;` at the top of this file? and will this compile in varnish 4.0?

maybe instead of the if statement, we should offer two separate vcl files and tell which one to include?

ddeboer: Heh this is C, not PHP, so `import xkey;` :wink:

src/ProxyClient/Varnish.php
```diff
@@ -151,6 +160,8 @@ protected function configureOptions()
         $resolver->setDefaults([
             'tags_header' => self::DEFAULT_HTTP_HEADER_CACHE_TAGS,
             'header_length' => 7500,
+            'tags_invalidation_by_purge' => false,
```
andrerom: You mean ban and xkey, and default to ban?

resources/config/varnish/fos_purge.vcl
andrerom: I can add it here at the top, but 1: doesn't it need to be at the top of the file? And 2: adding it by default will make it break if you don't have it afaik.

src/ProxyClient/Varnish.php
andrerom: @dbu ?

andrerom:
Updated, made it more generic, which fits well with work on xkey atm to add support for purging several tags at once, ref: https://github.com/varnish/varnish-modules/pull/47

src/ProxyClient/Varnish.php
```diff
@@ -150,7 +158,10 @@ protected function configureOptions()
         $resolver = parent::configureOptions();
         $resolver->setDefaults([
             'tags_header' => self::DEFAULT_HTTP_HEADER_CACHE_TAGS,
+            'tags_purge_header' => self::DEFAULT_HTTP_PURGE_HEADER_CACHE_TAGS,
             'header_length' => 7500,
+            'tags_invalidation_by_purge' => false, // false, true or "single" for purge invalidation per tag
+            'tags_separator' => ',',
```
ddeboer: Can we get rid of this option? For ban, our library can just use commas. It seems the combined surrogate keys will be [space-delimited](https://github.com/varnish/varnish-modules/pull/47), so no choice there either.

andrerom: ok

andrerom: done

doc/proxy-clients.rst
```diff
@@ -97,12 +97,30 @@ You can also pass some options to the Varnish client:
 * ``tags_header`` (X-Cache-Tags): Allows you to change the HTTP header used for
   tagging. If you change this, make sure to use the correct header name in your
   :doc:`proxy server configuration <proxy-configuration>`;
+* ``tags_purge_header`` (key): Allows you to change the HTTP header used for
+  purging tags when ``tag_mode`` is set to one of the ``purge*`` options. If
+  you change this, make sure to use the correct header name in your
+  :doc:`proxy server configuration <proxy-configuration>`;
+* ``tag_mode`` (ban): Makes it possible to invalidate tags
+  using purge when for instance using ``xkey`` instead of ban. Supports three
+  mods: ``ban``, ``purge`` and ``purge_single`` for purge per tag.
```
stof: typo: ``modes``

doc/varnish-configuration.rst
```diff
@@ -174,8 +205,12 @@ Tagging
 
 Feature: :ref:`cache tagging <tags>`
 
-If you have included ``fos_ban.vcl``, tagging will be automatically enabled
-using an ``X-Cache-Tags`` header.
+For an advance ``xkey`` setup you'll have to use ``fos_purge_xkey.vcl``
```
stof: ``advanced``

src/ProxyClient/Varnish.php
```diff
@@ -150,6 +158,8 @@ protected function configureOptions()
         $resolver = parent::configureOptions();
         $resolver->setDefaults([
             'tags_header' => self::DEFAULT_HTTP_HEADER_CACHE_TAGS,
+            'tags_purge_header' => self::DEFAULT_HTTP_PURGE_HEADER_CACHE_TAGS,
+            'tag_mode' => 'ban', // 'ban', 'purge' or 'purge_single' for purge invalidation per tag
```
stof: you should also configure the allowed values to have stricter validation:

```php
$resolver->setAllowedValues('tag_mode', array('ban', 'purge', 'purge_single');
```

doc/proxy-clients.rst
```diff
@@ -97,12 +97,30 @@ You can also pass some options to the Varnish client:
...
   tags. If there are more tags to invalidate than fit into the header, the
   invalidation request is split into several requests;
 * ``default_ban_headers`` Map of headers that are set on each ban request,
   merged with the built-in headers.
 
+Example for configuring xkey v0.9:
+
+    $options = [
+        'tags_header' => 'xkey',
+        'tags_invalidation_by_purge' => 'single',
```
stof: this does not match the code

andrerom:
Thanks @stof, ready for new round /cc @ddeboer @dbu

doc/varnish-configuration.rst
```diff
@@ -86,6 +98,25 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 Read more on `handling PURGE requests`_ in the Varnish documentation (for
 Varnish 3, see `purging for Varnish 3`_).
 
+Purge with xkey
+^^^^^^^^^^^^^^^
+
+Alternatively for use with xkey this allows purges by url as well as by ``key``
```
ddeboer: URL

doc/varnish-configuration.rst
```diff
@@ -174,8 +205,12 @@ Tagging
 
 Feature: :ref:`cache tagging <tags>`
 
-If you have included ``fos_ban.vcl``, tagging will be automatically enabled
-using an ``X-Cache-Tags`` header.
+For an advanced ``xkey`` setup you'll have to use ``fos_purge_xkey.vcl``
+and configure Varnish Client
+:ref:`configured in the cache invalidator <varnish_custom_tags_header>`.
```
ddeboer: This sentence is not quite right.

andrerom:
> Please rebase on master (where I fixed the build).

done, failures due to spellchecker, any pointers?

ddeboer:
> done, failures due to spellchecker, any pointers?

On the spellchecker you mean? Please add the words to https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/doc/spelling_word_list.txt.

andrerom:
added

doc/proxy-clients.rst
```diff
@@ -97,12 +97,28 @@ You can also pass some options to the Varnish client:
 * ``tags_header`` (X-Cache-Tags): Allows you to change the HTTP header used for
   tagging. If you change this, make sure to use the correct header name in your
   :doc:`proxy server configuration <proxy-configuration>`;
+* ``tags_purge_header`` (key): Allows you to change the HTTP header used for
+  purging tags when ``tag_mode`` is set to one of the ``purge*`` options. If
+  you change this, make sure to use the correct header name in your
+  :doc:`proxy server configuration <proxy-configuration>`;
+* ``tag_mode`` (ban): Makes it possible to invalidate tags
+  using purge when for instance using ``xkey`` instead of ban. Supports three
+  modes: ``ban``, ``purge`` and ``purge_single`` for purge per tag.
```
dbu: can we rephrase this? "``tag_mode`` (ban): Specify the method how tags are invalidated. ``ban`` uses the built-in BAN functionality of Varnish, ``purge`` uses the xkey Varnish module, and ``purge_single`` ???" (not sure what the difference between purge and purge_single is, we should say this here (no long explanation, but a short note))

doc/proxy-clients.rst
```diff
@@ -97,12 +97,28 @@ You can also pass some options to the Varnish client:
 * ``tags_header`` (X-Cache-Tags): Allows you to change the HTTP header used for
   tagging. If you change this, make sure to use the correct header name in your
   :doc:`proxy server configuration <proxy-configuration>`;
+* ``tags_purge_header`` (key): Allows you to change the HTTP header used for
```
dbu: what is this header? is it in addition to tags_header? if its instead of the other tags_header, then why not use the tags_header configuration for it?

doc/varnish-configuration.rst
```diff
@@ -33,6 +33,18 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
     trigger invalidation are whitelisted here. Otherwise, lost cache invalidation
     requests will lead to lots of confusion.
 
+Advance Varnish Configuration: xkey
```
dbu: s/Advance/Advanced/

doc/varnish-configuration.rst
```diff
@@ -33,6 +33,18 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With Varnish 4 and 5 you can optionally take advantage of official Varnish
+modules, specifically `xkey Varnish module`_. This module will provide better
+performance for tag handling as it adds secondary hashes to objects, and as
+it allows for grace handling by means of softpurge on tags.
+
+These modules are bundled with for instance Ubuntu 16.10 and Debian 9
+and higher using package name ``varnish-modules``.
```
dbu: s/higher/later/

s/using/under the/

doc/varnish-configuration.rst
```diff
@@ -33,6 +33,18 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
     trigger invalidation are whitelisted here. Otherwise, lost cache invalidation
     requests will lead to lots of confusion.
 
+Advance Varnish Configuration: xkey
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With Varnish 4 and 5 you can optionally take advantage of official Varnish
+modules, specifically `xkey Varnish module`_. This module will provide better
+performance for tag handling as it adds secondary hashes to objects, and as
+it allows for grace handling by means of softpurge on tags.
```
dbu: can we say what it does and then just (softpurge) in paranthesis? i am not sure if people know well enough what softpurge means.

resources/config/varnish/fos_purge_xkey.vcl
```diff
@@ -0,0 +1,30 @@
...
+
+import xkey;
+
+sub fos_purge_recv {
+    if (req.method == "PURGE") {
+        if (!client.ip ~ invalidators) {
+            return (synth(405, "Not allowed"));
+        }
+
+        # If http header "key" is set, we assume purge is on key and you have Varnish xkey installed
```
dbu: i assume compile would fail on `import xkey;` anyways, if xkey is not available. if that is the case, i think we can remove this comment above

src/ProxyClient/Varnish.php
```diff
@@ -150,9 +158,13 @@ protected function configureOptions()
         $resolver = parent::configureOptions();
         $resolver->setDefaults([
             'tags_header' => self::DEFAULT_HTTP_HEADER_CACHE_TAGS,
+            'tags_purge_header' => self::DEFAULT_HTTP_PURGE_HEADER_CACHE_TAGS,
+            'tag_mode' => 'ban',
             'header_length' => 7500,
             'default_ban_headers' => [],
         ]);
+        // tag_mode options: 'ban', 'purge' or 'purge_single' for purge invalidation per tag
```
dbu: i think this comment is not needed, it just says what the code below is doing.

doc/proxy-clients.rst
```diff
@@ -97,12 +97,28 @@ You can also pass some options to the Varnish client:
...
 * ``header_length`` (7500): Control the maximum header size when invalidating
   tags. If there are more tags to invalidate than fit into the header, the
   invalidation request is split into several requests;
 * ``default_ban_headers`` Map of headers that are set on each ban request,
   merged with the built-in headers.
 
+Example for configuring xkey *(varnish-modules v0.9)*::
+
+    $options = [
+        'tags_header' => 'xkey',
```
dbu: is this configurable on varnish side? or is it hardcoded into the xkey module?

doc/proxy-clients.rst
andrerom: That was what I did at first using diet for both tag header and purge header, but @ddeboer asked me to split it since xkey's doc uses 'xkey' for requests and 'key' for purge in online examples (the header is different in older versions of the doc other places).

andrerom: Diet => it

doc/proxy-clients.rst
andrerom: Afaik hard coded for tags but not for purge

ddeboer:
Great work so far, @andrerom! 

We should also test this on Travis, by adding a couple of integration tests and installing the xkey vmod on Travis.

doc/proxy-clients.rst
andrerom: The difference between purge and purge single is somewhat important since xkey only takes one key at a time on purge atm, and this is where `purge_single` comes in.  `purge` is doing it like ban, just limited by header length, but this won't work with xkey before a PR over there is merged and new release trickles down to distributions.

See: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/332#issuecomment-264200486

ref

dbu: hm. should we remove purge for now and only have purge_single? if the feature is not even merged yet, its unsure if it ever will or it might be different. and if we only add it after its released, we can specify the version of the vmod that you need

andrerom:
> We should also test this on Travis, by adding a couple of integration tests and installing the xkey vmod on Travis.

agreed, anyone that wants to tackle that part?

JonasHaouzi:
Why should we do to finish this topic?

If we define before the battery of tests we need, I can help writing them, no problems.

dbu:
installing the xkey vmod in travis-ci would be a first step to be able to integration test this. would be great to get that, plus some integration tests and maybe unit tests where it makes sense.

we are currently focussing on getting 2.0 released and this is a feature planned for 2.1, but no harm in continuing work on this. and we can release 2.1 any time this is ready.

andrerom:
One attempt at getting the vmod available for testing is being done here: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/344

@dbu maybe a question to be asked there, but how to we get _that PR_ moving to be able to unblock this one?

dbu:
right, that would be the best way to get things testable. then people can use docker to develop locally as well. @ddeboer started working on that, but i guess he just did not get around to wrap it up.

ddeboer:
That’s right; I want to release version 2.0 of the library and bundle first.

andrerom:
ok, so blockers is 2.0 => testing PR => xkey PR.
Then what is blocking 2.0 release? :P

dbu:
> ok, so blockers is 2.0 => testing PR => xkey PR.

agree

> Then what is blocking 2.0 release? :P

we have an RC already, but there is a couple of cleanups on
FOSHttpCacheBundle and working on that might reveal issues. i think we
should wait for the bundle to be ready before tagging the library, or we
might have awkward problems of having a 2.0.1 of the library in two
weeks, with deprecations. or have to work around limitations. or need a
3.0 in 2 weeks.

andrerom:
ok, clear :)

gaillafr:
Would be great to have this feature soon! Thx @andrerom 😄

ddeboer:
> Any eta of a alpha/beta tag of the bundle so we can test out on our side?

Just [released](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/releases/) friendsofsymfony/http-cache-bundle 2.0.0-alpha1

XWB:
Will this feature make it into 2.0 @dbu ?

dbu:
hi @XWB this is too big a feature to add at this phase, we just tagged the hopefully last beta version. however, if somebody works on getting this ready to merge, i have no problems releaesing 2.1 in 2 weeks or whenever its ready...

first step is getting the docker image for testing ready: #344

andrerom:
> i have no problems releasing 2.1 in 2 weeks or whenever its ready...

well, this and docker image has been blocked by the 2.0 release for quite some time. So unsure about the 2 weeks part :)

andrerom:
@dbu varnish-modules compiles now, only failure now is with _(added)_ Varnish 5.2

So, functional tests, anything relevant in other on going tagging work I should know about? Or should I just start from tests fro BAN and adapt?

.travis.yml
```diff
@@ -1,27 +1,26 @@
...
-        - DOCCHECK=true
+      env: VARNISH_VERSION=4.1 DOCCHECK=true
+    - php: 7.0
+      env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1
+    - php: 7.0
+      env: VARNISH_VERSION=5.0
+    - php: 7.1
+      env: VARNISH_VERSION=5.1 VARNISH_MODULES_VERSION=0.12.1
+    - php: 7.2
+      env: VARNISH_VERSION=5.2
```
dbu: lets keep this out of this PR and do it separately to keep the scope of the PR to x-key support

doc/proxy-clients.rst
```diff
@@ -101,12 +101,24 @@ You can also pass some options to the Varnish client:
   which tags to invalidate when sending invalidation requests to the caching
   proxy. Make sure that your :ref:`Varnish configuration <varnish_tagging>`
   corresponds to the header used here;
+* ``tag_mode`` (ban): Makes it possible to invalidate tags
+  using purge when for instance using ``xkey`` instead of ban. Supports three
+  modes: ``ban``, ``purge``, and ``purge_single`` for purge per tag.
```
dbu: we should list what the modes do. maybe do like this: 
```
You can use the [link:xkey] module for improved cache tag handling. Without xkey, the only
supported mode is ``ban``. With xkey the supported modes are ``purge`` and ``purge_single``.
``purge_single`` sends a separate purge request per tag, which is the only mode supported by
the varnish modules prior to v0.10.2. If you have a newer version of the modules, it is more
efficient to use ``purge``.
```

doc/proxy-clients.rst
```diff
@@ -101,12 +101,24 @@ You can also pass some options to the Varnish client:
   which tags to invalidate when sending invalidation requests to the caching
   proxy. Make sure that your :ref:`Varnish configuration <varnish_tagging>`
   corresponds to the header used here;
+* ``tag_mode`` (ban): Makes it possible to invalidate tags
```
dbu: the format is inconsistent between the first point and the rest. lets use "(default: ``ban``)"

doc/proxy-clients.rst
```diff
@@ -101,12 +101,24 @@ You can also pass some options to the Varnish client:
...
+  using purge when for instance using ``xkey`` instead of ban. Supports three
+  modes: ``ban``, ``purge``, and ``purge_single`` for purge per tag.
 * ``header_length`` (default: 7500): Control the maximum header length when
   invalidating tags. If there are more tags to invalidate than fit into the
   header, the invalidation request is split into several requests;
 * ``default_ban_headers`` (default: []): Map of headers that are set on each
   ban request, merged with the built-in headers.
 
+Example for configuring xkey::
```
dbu: can we make this into a subsection "Configuring for xkey"?

if i understand correctly, you also need to change the cache tagging listener to use the header name `key` as that header is hardcoded in xkey. we should explain this here. (and maybe have a very short mention in the doc on the cache header option of the tag listener that references this subsection we create here: "See 'Configuring for xkey' if you want to use the Varnish xkey module")

doc/proxy-clients.rst
```diff
@@ -101,12 +101,24 @@ You can also pass some options to the Varnish client:
...
+
+    $options = [
+        'tags_header' => 'key',
```
dbu: if we follow my suggestion to keep the default X-Cache-Tags for xkey tag invalidation, this line can be removed

doc/varnish-configuration.rst
```diff
@@ -33,6 +33,18 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
     trigger invalidation are whitelisted here. Otherwise, lost cache invalidation
     requests will lead to lots of confusion.
 
+Advanced Varnish Configuration: xkey
```
dbu: maybe "Better tag invalidation: xkey" ?

doc/varnish-configuration.rst
```diff
@@ -33,6 +33,18 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With Varnish 4.1 and higher you can optionally take advantage of official Varnish
+modules, specifically `xkey Varnish module`_. This module will provide better
+performance for tag handling as it adds secondary hashes to objects, and as
+it allows for grace handling by means of softpurge on tags.
```
dbu: Since Varinsh 4.1 you can use the official [xkey Varnish module] for better performance with cache tags.
The module adds secondary hashes to objects, allowing to purge them specifically instead of using 
generic ``BAN`` instructions. This further allows grace handling by means of softpurge on tags.

doc/varnish-configuration.rst
```diff
@@ -33,6 +33,18 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
...
+
+These modules are bundled with for instance Ubuntu 16.10 and Debian 9
+using package name ``varnish-modules``.
```
dbu: i'd rather link to the official varnish doc on installation of the modules than give hints for specific distributions (although ubuntu and debian is the real thing, of course ;-) )

what we could mention here is that xkey is not installed alone but all the official modules have to be installed together. (do they also need to be activated, or are they automatically available once installed?)

doc/proxy-clients.rst
dbu: thinking about this some more, i think we should have a separate flag `tag_purge_single: false` that can be set to true and is about legacy modules support. then we can use ban/purge/softpurge|expire as options here. what is the better name, softpurge or expire?

doc/varnish-configuration.rst
```diff
@@ -86,6 +98,25 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 Read more on `handling PURGE requests`_ in the Varnish documentation (for
 Varnish 3, see `purging for Varnish 3`_).
 
+Purge with xkey
+^^^^^^^^^^^^^^^
+
+Alternatively for use with xkey this allows purges by URL as well as by ``key``
+header
+
+Subroutines are provided in ``resources/config/varnish-[version]/fos_purge_xkey.vcl``.
```
dbu: i think we did a bad choice with including the ban tags in the normal ban vcl. i'd prefer to have a separate file for xkey, maybe fos_tags_xkey.vcl ?

we would then use a special method PURGEKEYS resp. SOFTPURGEKEYS|EXPIREKEYS to avoid collisions with regular PURGE.

doc/varnish-configuration.rst
```diff
@@ -86,6 +98,25 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 Read more on `handling PURGE requests`_ in the Varnish documentation (for
 Varnish 3, see `purging for Varnish 3`_).
 
+Purge with xkey
```
dbu: to me this is essentially a tags topic, the fact that we "purge" the tag is the detail, not the main thing. so i would move this subsection below to the tags section. then we only have one place where we talk about xkey which should be less confusing.

doc/varnish-configuration.rst
```diff
@@ -174,6 +205,17 @@ Tagging
 
 Feature: :ref:`cache tagging <tags>`
 
+For this feature you'll need to choose between either xkey, or BAN based tags invalidation.
```
dbu: we should repeat the core difference here and link to the section above that introduces xkey.

resources/config/varnish/fos_purge_xkey.vcl
```diff
@@ -0,0 +1,29 @@
...
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+import xkey;
+
+sub fos_purge_recv {
+    if (req.method == "PURGE") {
```
dbu: i would prefer to use a specific method for key purge. that makes it more modular

resources/config/varnish/fos_purge_xkey.vcl
```diff
@@ -0,0 +1,29 @@
...
+        if (!client.ip ~ invalidators) {
+            return (synth(405, "Not allowed"));
+        }
+
+        if (req.http.key) {
+            # If possible use soft purge (expiry) to respect grace time instead of normal purge.
+            # set req.http.n-gone = xkey.softpurge(req.http.key);
```
dbu: our aim is that these snippets can be deployed to the server and used as is. i would prefer to use different http methods to distinguish. that would also allow the application to use both modes depending on what is happening.

resources/config/varnish/fos_purge_xkey.vcl
```diff
@@ -0,0 +1,29 @@
...
+
+import xkey;
+
+sub fos_purge_recv {
+    if (req.method == "PURGE") {
+        if (!client.ip ~ invalidators) {
+            return (synth(405, "Not allowed"));
+        }
+
+        if (req.http.key) {
```
dbu: should we call the header `X-Cache-Keys` instead of `key`? that would distinguish between the key on the response and the information which keys to invalidate, and we would avoid to require changing the tag header configuration depending on the tag invalidation mode.

resources/config/varnish/fos_purge_xkey.vcl
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+import xkey;
+
+sub fos_purge_recv {
```
dbu: this file should mention either here or next to the xkey function that the tags MUST be set in the header called `key` (if i understood correctly how xkey works)

doc/proxy-clients.rst
dbu: and we should say "See below for how to configure your application to support xkey".

resources/config/varnish/fos_purge_xkey.vcl
andrerom: you mean `X-Cache-Tags`, yes we can. On responses the header is `xkey`, that's afaik hardcoded in the module, but on purge we have full controll.

doc/proxy-clients.rst
andrerom: Ok, but note that #385 was not about letting people configure soft_purge over purge here, it was about the code using FOSHttpCache being able to tell the system if soft purge is ok or not, so I don't see a need for a soft_purge option here.

E.g. On deletion of an entity, soft purge is probably not desirable as it can lead to broken links during stale periods, while a creation or update of a entity can safely just expire whatever needs to be refreshed.

dbu:
ping @andrerom

doc/proxy-clients.rst
dbu: but would it not make sense to have vcl handle both cases, and have the application decide whether soft or hard purge should be done?

andrerom:
@dbu A bit pr occupied on other things atm, might find time tonight or tomorrow. But I'm stuck on the failing test, unsure why it fails.

dbu:
> Is the test failure related to the response cache tag we use (X-Cache-Tags vs xkey?)

i think it is, in CacheInvalidatorTest.php. are you sure the modules are installed and activated properly? can we better validate that the request actually goes to the right place in vcl? could it be that our detection of cache hit/miss is incorrect in this case? we do not look at the actual response afaik, just assume its a miss. or could there be some delay with this that leads to race conditions with the test?

so in short: i don't know why it fails, but here are some ideas what to look at ;-)

andrerom:
yep, similar thinking.

doc/proxy-clients.rst
andrerom: Yes, but I don't think that should be config, rather something application specifies on purge _(purge vs expiry, where the latter allows stalnes by grace or by stale headers depending on proxy)_-

resources/config/varnish/fos_purge_xkey.vcl
andrerom: That said, it's format is different _(space separated and not comma)_ and xkey doc suggest using `xkey-purge` header. Could use `xkey-purge` and `xkey-softpurge` headers to separate hard and soft purging for instance?

andrerom:
@dbu  passing, fixed in f690918... :)

resources/config/varnish/fos_purge_xkey.vcl
dbu: good point, lets use xkey-purge then, as the doc suggests. and xkey-softpurge for soft purge, so that the user can include one snippet and the php application can chose which mode to do.

dbu:
woot! that is awesome. so do we still need the docker setup? seems at least its not a blocker anymore...

can your please rebase on master to get the ci fixes in, and address my comments (mostly on doc and varnish configuration?)

andrerom:
> woot! that is awesome. so do we still need the docker setup? seems at least its not a blocker anymore...

no, we can do the docker stuff later and not have to depend on it

dbu:
i added this to the 2.1 milestone, but we want to release 2.1 next monday the latest, to get symfony 4 support official. if we can't wrap this up until then, i have no problem making a 2.2 in january.

andrerom:
Btw, Varnish has accepted doc update on tag format.
So they officially support same as Faslty now, space separated format both for response and purge header.

Example:
```
xkey: tag1 tag2 tag3
```

Would there be an idea to align up the code across backends around this?

dbu:
i don't think we can change it for our custom varnish header, as that would be a BC break. but for the xkey implementation sure, lets use that. will also possibly open the road to easily integrate fastly.

dbu:
i just released 2.1.0 of the library now. but as said, i am happy to release a 2.2 asap, whenever this is ready. just did not want to withhold symfony 4 support any longer.

doc/proxy-clients.rst
andrerom: Format is different, so would rather go for `xkey-purge`

doc/proxy-clients.rst
andrerom: Could you check this again and expand on what you wanted here? (as in, there is already an example here, should I change it so both have own sub header here? )

andrerom:
@dbu Updated and ready to move forward now, anything besides somehow trying to setup tests for softpurge you had in mind?

doc/proxy-clients.rst
dbu: my suggestion was that we group everything xkey related to its own section so we have a place to explain how the whole thing works together. i think we have most or all of the documentation somewhere, but its spread in between other things.

resources/config/varnish/fos_tags_xkey.vcl
```diff
@@ -0,0 +1,30 @@
...
+import xkey;
+
+sub fos_tags_xkey_recv {
+    if (req.method == "PURGEKEYS") {
+        if (!client.ip ~ invalidators) {
+            return (synth(405, "Not allowed"));
+        }
+
+        // If tag invalidation api allows caller to specify expire/purge in the future,
+        // then adjust this to be able to handle both headers in same request.
```
andrerom: @dbu NOTE: I was thinking of proposing to do a `ExpireTagCapable` for instance to have a method that clearly means expire and not purge, but then would need to create new `VarnishXkey` proxy client and so on as Ban would not support that. So just left this here for a possible 3.0 cleanup where things are aligned across newer proxies.

dbu: i think its good to have both handled in this same vcl snippet. are both purge and softpurge available with all versions? as in: does this vcl always compile when the xkey extension is available?

is there a use case to mix expire and purge? if so, we could change the code below to use just `if` instead of `elseif` to handle both headers.

andrerom: > are both purge and softpurge available with all versions? as in: does this vcl always compile when the xkey extension is available?

yes and yes

> is there a use case to mix expire and purge?

right now it  is not possible, both in terms of design of tag invalidation api and as long as FOS emits queued url request for each call to invalidate.

So I'll remove the comment, it was anyway more for your eyes to be aware of this being possible if we allow for it.

src/ProxyClient/Varnish.php
```diff
@@ -61,13 +63,23 @@ class Varnish extends HttpProxyClient implements BanCapable, PurgeCapable, Refre
...
+        $tagMode = $this->options['tag_mode'];
         $escapedTags = array_map('preg_quote', $this->escapeTags($tags));
 
-        $chunkSize = $this->determineTagsPerHeader($escapedTags, '|');
+        $chunkSize = $this->determineTagsPerHeader($escapedTags, 'ban' === $tagMode ? '|' : ' ');
 
         foreach (array_chunk($escapedTags, $chunkSize) as $tagchunk) {
-            $tagExpression = sprintf('(^|,)(%s)(,|$)', implode('|', $tagchunk));
-            $this->ban([$this->options['tags_header'] => $tagExpression]);
+            if ('ban' === $tagMode) {
```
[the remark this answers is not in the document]
andrerom: I ended up doing private methods here as I did not entirely see how inline anonymous functions would help cleanup flow here. If you want something else please share some pseudo example for me to apply form.

src/ProxyClient/Varnish.php
```diff
@@ -61,18 +63,35 @@ class Varnish extends HttpProxyClient implements BanCapable, PurgeCapable, Refre
...
+        $invalidationMethod = "{$tagMode}TagChunk";
         $escapedTags = array_map('preg_quote', $this->escapeTags($tags));
 
-        $chunkSize = $this->determineTagsPerHeader($escapedTags, '|');
+        $chunkSize = $this->determineTagsPerHeader($escapedTags, 'ban' === $tagMode ? '|' : ' ');
 
         foreach (array_chunk($escapedTags, $chunkSize) as $tagchunk) {
-            $tagExpression = sprintf('(^|,)(%s)(,|$)', implode('|', $tagchunk));
-            $this->ban([$this->options['tags_header'] => $tagExpression]);
+            $this->$invalidationMethod($tagchunk);
```
dbu: i would prefer to be explicit here and say `if ('ban' === $tagMode) { $this->banTags(); } else {$this->purgekeys();}`  or something like that.

andrerom: Ok, tought you said you did not want condition.

resources/config/varnish/fos_tags_xkey.vcl
```diff
@@ -0,0 +1,30 @@
...
+sub fos_tags_xkey_recv {
+    if (req.method == "PURGEKEYS") {
+        if (!client.ip ~ invalidators) {
+            return (synth(405, "Not allowed"));
+        }
+
+        # Based on provided header invalidate (purge) or expire (softpurge) the tagged content
+        if (req.http.xkey-purge) {
+            set req.http.n-gone = xkey.purge(req.http.xkey-purge);
+        } elseif (req.http.xkey-softpurge) {
```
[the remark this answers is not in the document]
andrerom: If you can tell me how to add up int’s in vcl to append an initial value of 0, then yes.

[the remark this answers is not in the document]
andrerom: As in if we support both the return response should reflect all the affected items.

andrerom:
> would be awesome to wrap this up this week, as i should release for symfony 4 support.


I’m all up for that, but off for rest of week in Scotland, so the changes you suggest are ok from my side so you can optionally go ahead and add them if you want to move towards 2.2rc

src/ProxyClient/Varnish.php
dbu: sorry, not sure what i meant anymore. but looking at the code i think a condition is more readable.

resources/config/varnish/fos_tags_xkey.vcl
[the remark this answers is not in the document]
dbu: oh well... maybe lets cheat and do 2 numbers and reply `"Purged "+req.http.n-gone+" objects and soft purged "+req.http.n-softgone+" objects"`?

doc/varnish-configuration.rst
```diff
@@ -33,6 +33,29 @@ simple ACL, valid for all Varnish versions from 3 onwards, looks as follows:
...
+> 2) every time you test a ban expression, it checks every object in the cache that is
+> older than the ban itself.
+
+It also has more predicable performance thanks to softpurge support on tags for grace.
+
+Minimum version of `varnish-modules` is v0.10.2*. It's part of Ubuntu 17.10 ("Artful")
+and higher. But can also be [installed on any other linux platform](https://github.com/varnish/varnish-modules#installation).
+
+* _v0.10.2 is the first version introducing support for purging several tags at once.
+You can technically use earlier versions like v0.9.1, however then you will not be
```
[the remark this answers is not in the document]
dbu: i'd just love to be precise in the documentation to avoid confusion.

resources/config/varnish/fos_tags_xkey.vcl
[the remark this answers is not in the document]
andrerom: why not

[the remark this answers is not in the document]
dbu: its debug information only, so i guess it should not be an issue. and summing it up seems to not be a varnish thing. we could go to inline C but seems not worth it. also inline C does only run when its specifically allowed.

doc/varnish-configuration.rst
```diff
@@ -174,6 +198,43 @@ Tagging
...
+:ref:`xkey based tags invalidation <varnish_xkey_intro>`.
+
+Using xkey
+^^^^^^^^^^
+
+For an advanced ``xkey`` based setup you can use ``fos_tags_xkey.vcl`` and
+:ref:`configure Varnish Client for xkey <varnish_custom_tags_header>`.
+
+Subroutines are provided in ``resources/config/varnish-[version]/fos_tags_xkey.vcl``.
+To enable this feature, add the following to ``your_varnish.vcl``:
```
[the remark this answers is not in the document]
andrerom: Somewhat in-precise here again on personal preferences which I have no idea how to second guess, here on setup instructions adaption for the changes you suggest, so I guess I'm in for 2-3 more rounds of reviews 🎉 on top of the 20 previous once. </>

But might be the incomming changes hits the 🥇, who knows.

.travis.yml
```diff
@@ -19,6 +19,7 @@ matrix:
     - php: 5.6
     - php: 7.0
     - php: 7.1
+      env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1 DEPENDENCIES="symfony/lts:^3"
```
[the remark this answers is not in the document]
andrerom: removed

src/ProxyClient/Varnish.php
andrerom: I agree, just way beyond caring to argue anymore with the input.

andrerom:
@dbu updated and passing

dbu:
thanks, i'll try to have a look tonight. 
sorry for being pedantic - i want the doc to be as understandable and clear as possible, but thats a) difficult and b) often also a matter of taste.

andrerom:
I get it, and I also realize this is somewhat generic problem when attempting to contribute features and other larger changes, where it is hard to know for both sides up front how to approach things.

dbu:
continued in #409

dbu:
merged in #409

dbu:
now that this is merged, any chance that you can add configuration support in FOSHttpCacheBundle for this feature as well?

dbu:
created https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/431 so we don't forget this

PR #333, opened 2016-11-12
Bump php requirements to PHP 5.6 for 2.0

andrerom:
Like Twig 2.0 [just did this week](https://github.com/twigphp/Twig/commit/af3b0a550c77b16fb6e9679ce6cc147d35888d04), it makes sense to drop PHP 5.5 support as it is EOL. 5.6 on the other hand is supported until 2018, and is the newest PHP version currently available on Debian/RHEL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ group: edge
 language: php
 
 php:
-  - 5.5
   - 5.6
 # broken with puli discovery  - hhvm
   - 7.0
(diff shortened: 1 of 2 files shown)


stof:
does it bring any benefit to the project ?

Twig 2.0 did it because made the code of the new optimization much simpler. And they were able to do that without impacting the ecosystem relying on PHP 5.5 support (Symfony 3 and Drupal for instance) because Twig 1.x and 2.0 have the same feature set (Twig 2 only cleans deprecated APIs). This is not the case for FOSHttpCache, so it would have a bigger impact for the community (potentially preventing some projects to migrate to the v2 of the library if their policy makes them keep PHP 5.5 support for now)

andrerom:
Besides reducing test matrix, urging people to use newer php versions, no real benefit, so you have a point.

dbu:
thanks for asking the question andré. but i agree with stof.

PR #335, opened 2016-12-03
Remove Interface suffix

ddeboer:
Following a convention of many PHP libraries, including PHP-HTTP.
See also https://github.com/php-http/HttplugBundle/issues/28

PR #336, opened 2016-12-03
Added interface for the hash generator

Nyholm:
I have the scenario that I do not want to cache responses for a specific group of users. Say that the administrators should always receive fresh content. 
To achieve this I created my `CustomUserContextProvider` to add some special parameters to the `UserContext`. Since the default `HashGenerator` makes a sha1 hash over all parameters and values I have no good way in writing a custom rule in my VCL to make sure the special "no-cache hash" gets a fresh response. 

By introducing an interface for the `HashGenerator` I may inject my own `HashGeneratorThatSupportsNoCacheHashes` to the `UserContextListener`. This will allow me to write an exception VCL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).q
+See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
 2.0.0
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
+See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).q
 
 2.0.0
 -----
@@ -10,6 +10,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 
 * Raised minimum PHP version to 5.5.
 * **BC break:** Removed the `Interface` suffix from all interfaces.
+* **BC break:** Renamed ``HashGenerator`` to ``DefaultHashGenerator``.
+* Added interface ``HashGenerator``
 
 ### HTTP
 
--- a/src/UserContext/DefaultHashGenerator.php
+++ b/src/UserContext/DefaultHashGenerator.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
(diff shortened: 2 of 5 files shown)


dbu:
this makes sense to me, yes. lets get rid of the interface suffix as per #335. i suggest we call the default implementation DefaultHashGenerator

Nyholm:
Thank you for the feedback. Ive update the bundle

dbu:
sorry, with merging the renaming change this is now in conflict. can you please rebase and fix the conflicts? i assume that we changed something on the HashGenerator that would now need to be moved to DefaultHashGenerator...

Nyholm:
I've updated the PR

dbu:
sorry, one last thing i missed until now: the documentation also needs to be adjusted, notably this section: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/5f39c447f652c68e6a17c3ea62a69fa1538e8f8f/doc/user-context.rst#calculating-the-user-context-hash - i would say to use the DefaultHashGenerator and then mention that completely custom is possible by implementing the interface.

dbu:
the bundle actually needs to be adjusted in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/cf1b6696ec06889ec3504fb3a9e7af94dd68a0c3/src/Resources/config/user_context.xml#L8

Nyholm:
Thank you for all the feedback and comments. Ive updated the change log.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 Changelog
 =========
 
-See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
+See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).q
```
dbu: extra `q`

Nyholm: I do not understand. So you do not want the q there?

=)

dbu:
thank you!

Nyholm:
Thank you for merging

src/UserContext/DefaultHashGenerator.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+namespace FOS\HttpCache\UserContext;
+
+use FOS\HttpCache\Exception\InvalidArgumentException;
+
+/**
+ * Generate a hash for a UserContext by getting all the parameters needed across all registered services.
+ */
+class DefaultHashGenerator
```
dbu: argh, missed something: this needs a `implements HashGenerator`. will add that

dbu: c2e78a4

Nyholm: Gee, this is a horrible PR for me. Full with typos.

PR #342, opened 2016-12-23
Make MultiplexerClient implement TagCapable

### LINKED ISSUE #338: multiplex client should implement TagInterface
the multiplexer client should also support multiplexing the tag invalidation methods provided by TagInterface

ddeboer:
Fix #338.

### DIFF
--- a/src/ProxyClient/MultiplexerClient.php
+++ b/src/ProxyClient/MultiplexerClient.php
@@ -13,16 +13,18 @@ namespace FOS\HttpCache\ProxyClient;
 
 use FOS\HttpCache\Exception\ExceptionCollection;
 use FOS\HttpCache\Exception\InvalidArgumentException;
+use FOS\HttpCache\Exception\UnsupportedProxyOperationException;
 use FOS\HttpCache\ProxyClient\Invalidation\BanCapable;
 use FOS\HttpCache\ProxyClient\Invalidation\PurgeCapable;
 use FOS\HttpCache\ProxyClient\Invalidation\RefreshCapable;
+use FOS\HttpCache\ProxyClient\Invalidation\TagCapable;
 
 /**
  * This class forwards invalidation to all attached clients.
  *
  * @author Emanuele Panzeri <thepanz@gmail.com>
  */
-class MultiplexerClient implements BanCapable, PurgeCapable, RefreshCapable
+class MultiplexerClient implements BanCapable, PurgeCapable, RefreshCapable, TagCapable
 {
     /**
      * @var ProxyClient[]
@@ -38,7 +40,8 @@ class MultiplexerClient implements BanCapable, PurgeCapable, RefreshCapable
     {
         foreach ($proxyClients as $proxyClient) {
             if (!$proxyClient instanceof ProxyClient) {
-                throw new InvalidArgumentException('Expected ProxyClientInterface, got: '.
+                throw new InvalidArgumentException(
+                    'Expected ProxyClientInterface, got: '.
                     (is_object($proxyClient) ? get_class($proxyClient) : gettype($proxyClient))
                 );
             }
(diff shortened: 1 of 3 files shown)


src/ProxyClient/MultiplexerClient.php
```diff
@@ -131,13 +165,19 @@ public function refresh($url, array $headers = [])
...
+        $result = null;
+
         foreach ($this->proxyClients as $proxyClient) {
             if (is_subclass_of($proxyClient, $interface)) {
-                call_user_func_array([$proxyClient, $method], $arguments);
+                $result = call_user_func_array([$proxyClient, $method], $arguments);
             }
         }
+
+        return $result;
```
ddeboer: We have to return a string for header name and value (not an array), so I’m simply returning the value of the last TagCapable client. Does that make sense? Or should we just return null or some constant string `multiple`?

dbu: i would hope that usually the clients are the same implementation, so this solution seems fine to me. we should add some note in the documentation of the multipart client about this though.

however, i propose that for the getters we call the first client instead of using invoke. if we agree on that, the changes to the invoke method can be reverted.

src/ProxyClient/MultiplexerClient.php
```diff
@@ -96,6 +98,38 @@ public function flush()
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function getTagsHeaderValue(array $tags)
+    {
+        return $this->invoke(TagCapable::class, 'getTagsHeaderValue', [$tags]);
```
dbu: this is not supposed to change anything on the client, so i think we should instead just call the first client and return that, instead of calling all of them and then return the last.

src/ProxyClient/MultiplexerClient.php
```diff
@@ -96,6 +98,38 @@ public function flush()
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTagsHeaderName()
+    {
+        return $this->invoke(TagCapable::class, 'getTagsHeaderName', []);
```
dbu: same for this

ddeboer:
Updated. Decided to not add the implementation details to the docs as it would clutter them up too much in my opinion.

src/ProxyClient/MultiplexerClient.php
```diff
@@ -126,18 +160,51 @@ public function refresh($url, array $headers = [])
...
+     * @param string $method    The method to invoke
+     * @param array  $arguments The arguments to be passed to the method
+     *
+     * @return mixed Return value of ProxyClient method call
+     */
+    private function invokeFirst($interface, $method, array $arguments)
+    {
+        foreach ($this->getProxyClients($interface) as $proxyClient) {
+            return call_user_func_array([$proxyClient, $method], $arguments);
+        }
```
dbu: what should we do when none of the clients implemented the requested interface? a bit hard to say - doing tags will not actually do anything, so we could throw an exception. on the other hand, we might just return a header name and an empty header value in that case, to not break users that check on the interface to know if they can use tags.

or we could create a TagCapableMultiplexerClient extends MultiplexerClient that will fail if none of its wrapped clients is TagCapable

ddeboer: I’m not sure really. Throwing an exception would be the most explicit way of handling that situation.

dbu: i guess the multiplexer client has this problem for all operations. it hides what the actual clients support or don't support. so yes, lets throw an exception. i will check if the documentation is explaining the special situation with multiplexer client

dbu: maybe you can review and merge #346 and then add a bit about exceptions for tag operations in this PR?

src/ProxyClient/MultiplexerClient.php
```diff
@@ -126,18 +161,53 @@ public function refresh($url, array $headers = [])
...
+
+        throw UnsupportedProxyOperationException::cacheDoesNotImplement($interface);
```
ddeboer: Now throwing exceptions (for `invokeFirst` only). Just being lazy and passing the interface’s FQDN in there, but the exception message is clear enough.

dbu:
thanks, i would say this is definitely good enough for something that is a special solution in the first place. i have restarted the php 5.5 build, it looked more like we have some test hoping to not run over the boundary of a second...

can you merge #346 and then rebase on master to add to the note about the multiplexer client to say that while other capabilities can be silently ignored, the tag operations will throw an exception as they would have to return a value?

dbu:
are we ready for 2.0 with this? or is there anything left to do in the library? we tagged alpha1, so i would say after merging this lets tag beta1? we should then try to make the bundle ready-ish in case that reveals issues with the library.

ddeboer:
Updated docs.

dbu:
great, thanks! i just looked over the bundle - there is quite some stuff in the 2.0 milestone. maybe we should go over that again and see what are quick-wins that we can do easily, and what are must haves, and then kick out everything else to not delay too much. new features can also come in 2.1.

PR #344, opened 2017-01-02
[WIP] Add Dockerfile

ddeboer:
This PR adds a Docker image that adds Varnish (from source) with varnish-modules (which we need for #332) and Nginx (todo) to an Alpine-based PHP image. It would be better to run Varnish and Nginx in separate containers (managed with Docker compose). However, our tests start and stop Varnish, so the Varnish binary must be installed in the PHP container itself. 

This Docker image can be used to:
- replace our testing setup on Travis
- have a local development and testing setup identical to the one on Travis
- install specific Varnish versions (because we build from source)
- replace the Vagrant VM in workshops, demos etc.

**Todo**
- [ ] Nginx
- [ ] Wrap in a Makefile to make sed the PHP and Varnish versions (for running on Travis)
- [ ] Update travis.yml
- [x] Switch to varnish-modules master when https://github.com/varnish/varnish-modules/pull/38 is merged.

**Questions**
1. Should this be in this repository itself? I prefer to have it here (and deploy to Docker hub under friendsofsymfony?) rather than in a separate repository.

/cc @joelwurtz as I know he has Docker experience. :wink:


docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '2'
+services:
+  app:
+    image: friendsofsymfony/fos-http-cache
```
ddeboer: Is this a good namespace?

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM php:7.0-fpm-alpine
```
joelwurtz: I had big issues when using varnish with alpine don't know if it's still the case, but compilation of the vcl was hard on alpine due to the musl lib

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    && tar xf varnish-modules.tar.gz \
+    && cd varnish-modules-issue-7 \
+    && PKG_CONFIG_PATH=/usr/local/lib/pkgconfig ./bootstrap \
+    && PKG_CONFIG_PATH=/usr/local/lib/pkgconfig ./configure \
+    && make \
+    && make install \
+    && cd .. \
+    # TODO: remove hard-coded branch name when https://github.com/varnish/varnish-modules/pull/38 has been merged
+    && rm -r varnish-modules-issue-7 varnish-modules.tar.gz \
+    && apk del .build-deps
```
joelwurtz: IMO it's better to have a bash script for that as it's easier to debug but not really a big deal

docker-compose.yml
joelwurtz: If you own the organisation on the docker registry then yes :)

dbu:
> +    image: friendsofsymfony/fos-http-cache
> 
> Is this a good namespace?

i don't want to create the impression that this is an image that is
useful for something else than developping/testing fos-http-cache. if
this is publicly visible, maybe call it friendsofsymfony/http-cache-test
or something like that? (the fos- is too much, its in the organization
name already)

to what joel commented: i don't know the fine details of docker, but
hope we can keep this as simple as possible. imho the docker image
should be defined by the config file in this repo and built on the fly,
not built separately and uploaded to the registry.

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM php:7.0-fpm-alpine
+
+ENV VARNISH_VERSION 5.0.0
+ENV VARNISH_MODULES_VERSION "issue/7"
+
+# See http://git.alpinelinux.org/cgit/aports/tree/main/varnish/musl-mode_t.patch
+ADD mode_t.patch /
```
stof: where is this patch ?

docker/Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM php:7.0-fpm-alpine
+
+ENV VARNISH_VERSION 5.0.0
```
dbu: can this be automated in travis to test several versions?

stof: shouldn't this use a build arg instead, as it is an env var affecting the image building, not the runtime (it is useless at runtime AFAIK.

btw, if a build arg is used, it may be possible to overwrite it when building the image, so testing different versions in different Travis jobs. Using ``ENV``, it cannot be achieved (except by using ``sed`` to rewrite the Dockerfile)

docker-compose.yml
stof: docker-composer should rather reference the local Dockerfile instead of referencing an image on the hub. It would avoid having to build the image and publish it each time you change the Dockerfile (which is a pain when contributing)

joelwurtz: It's also can be a pain for contributing to have it local as if the way of installing varnish has changed (archive not being available, new dependancy to add for compiling, other stuff...) then people wanting to test on local will need to have the knowledge of installing varnish. 

Having it on the hub remove the necessity of this knowledge and ensure that the image you pull is working.

IMO the best would be to have it on local by default, but still publish it to the docker hub, so if someone struggle when building he can still use the last version published.

dbu: having both a local config and a uploaded image seems like it could lead to confusion. if we use this on travis-ci we will quickly notice when the build no longer works, so i would vote for having a local docker file in this repo.

stof: I agree with @dbu. If the Dockerfile becomes broken, contributors to this library should rather report it so that it is fixed instead of using an older image. Thus, the Travis build on their PR would be broken because of that anyway.

docker/Dockerfile
andrerom: @ddeboer you can change it to: `ARG VARNISH_VERSION=5.0.0`

docker/Dockerfile
andrerom: if this is still a problem, and if we only care about varnish 5 here for now we can use Debian scratch which is very soon final anyway, and contains apt packages for `varnish` and `varnish-modules`. See: https://github.com/ezsystems/ezplatform/blob/a5711195af5b7f2bfd78d9de21ca62ad304a5fe5/doc/docker-compose/Dockerfile-varnish

gsdevme:
I mean this image is for development testing only right (I assume)? So surely we wouldn't want a new docker image created each PR, surely wants a stable resource to test that is static and consistent?

However when contributions need to be made to the image it should be possible without impacting all PRs to this PHP project. My view would be to move the Dockerfile into its own repository so that contributions against docker can be made & built & tested independent of the PHP Library.

Anyone with experience of PHP and not Docker can continue to test locally or via Travis without any funkyness happening.

dbu:
> I mean this image is for development testing only right (I assume)? So
> surely we wouldn't want a new docker image created each PR, surely wants
> a stable resource to test that is static and consistent?

i would hope that docker does not randomly break. the problem with a
hosted image is that it would need to be "compiled" and uploaded
whenever the definition changes. is building the image such a big thing?
if that adds like 30 seconds to the build i don't care much, vs the
effort we would take to maintain a separately hosted docker image.

if building on the fly is really impractical, then indeed a separate
repo sounds the way to go.

docker-compose.yml
andrerom: If so _(no need to have image on the hub which can be confusing and become out of date)_ the `image` param can be omitted.

dbu:
i am closing this as it is quite old.

if somebody wants to pick things up again, i am not against having a docker image testbed. it might be best in a separate repository though, depending on the amount of files needed for it.

PR #349, opened 2017-02-02
enable hhvm builds again

dbu:
httplug does no longer strictly depend on puli, maybe things work again now?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: php
 
 php:
   - 5.6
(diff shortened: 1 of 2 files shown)


dbu:
@ddeboer ok with this? or did you want to do another attempt at this? i did not spend much time, probably we miss something to run hhvm as a webserver so varnish can talk to it.

ddeboer:
I agree; with PHP 7 HHVM has become much less relevant anyway.

PR #352, opened 2017-03-06
have extension point for dynamic caching servers in HttpDispatcher

dbu:
inspired by https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/361 i think there are valid use cases where you want a dispatcher that gets server ips from some sort of modifiable source rather than the static configuration file

### DIFF
--- a/src/ProxyClient/HttpDispatcher.php
+++ b/src/ProxyClient/HttpDispatcher.php
@@ -71,6 +71,10 @@ class HttpDispatcher
      * If you specify a custom HTTP client, make sure that it converts HTTP
      * errors to exceptions.
      *
+     * If your proxy server IPs can not be statically configured, extend this
+     * class and overwrite getServers. Be sure to have some caching in
+     * getServers.
+     *
      * @param string[]             $servers    Caching proxy server hostnames or IP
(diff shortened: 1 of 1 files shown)


crevillo:
i'd say +1 :). 
i can't find another way to solve it from the php side. other option could involve to execute something when those varnish ips changes or a new varnish come in place (talking about autoscaling here), but that will be more a thing for deploy scripts than for this php part.

dbu:
the fancy version would be to inject a server ip provider instead of a simple array. but that feels quite overengineered for the normal use case. with this change, its possible to write your own dispatcher and overwrite getServers to load ips from some runtime information source.

as said in the other issue, the real cloud approach would be message queues with a cache invalidation consumer subscribed to that queue on each caching proxy.

plozmun:
Can you merge this?

dbu:
@ddeboer any objections or is this ok?

ddeboer:
Thanks!

plozmun:
Can you merge this to beta-1? Thank you

ddeboer:
This will be part of 2.0.0-beta2 (or 2.0.0 final).

PR #353, opened 2017-04-27
Implemented TagHeaderFormatter interface

Toflar:
See https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/281 for more information

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Raised minimum PHP version to 5.6.
 * **BC break:** Removed the `Interface` suffix from all interfaces.
 * **BC break:** Renamed `HashGenerator` to `DefaultHashGenerator`.
+* Added `HashGenerator` interface.
 
 
 ### HTTP
@@ -84,7 +85,6 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 
 * **BC break:** The `ResponseTagger` now expects an instance of
   `TagHeaderFormatter` instead of `TagCapable` as first argument.
-* Added `HashGenerator` interface.
 
 1.4.2
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Raised minimum PHP version to 5.6.
 * **BC break:** Removed the `Interface` suffix from all interfaces.
 * **BC break:** Renamed `HashGenerator` to `DefaultHashGenerator`.
+* **BC break:** The `ResponseTagger` now expects an instance of
+  `TagHeaderFormatterInterface` instead of `TagCapable` as first argument.
 * Added `HashGenerator` interface.
 
 ### HTTP
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Raised minimum PHP version to 5.6.
 * **BC break:** Removed the `Interface` suffix from all interfaces.
 * **BC break:** Renamed `HashGenerator` to `DefaultHashGenerator`.
-* **BC break:** The `ResponseTagger` now expects an instance of
-  `TagHeaderFormatterInterface` instead of `TagCapable` as first argument.
-* Added `HashGenerator` interface.
+
 
 ### HTTP
 
@@ -42,6 +40,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Abstracting tags by adding new `TagCapable` for ProxyClients.
 * Added `strict` option to `ResponseTagger` that throws an exception when empty
   tags are added. By default, empty tags are ignored.
+* Added a `TagHeaderFormatter` interface that is used within the `ResponseTagger`
+  to provide the header name and the formatted tags header value.
 
 ### Varnish
 
@@ -78,6 +78,13 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * **BC break:** Refactored the proxy client test system into traits. Removed 
   `ProxyTestCase`; use the traits `CacheAssertions` and `HttpCaller` instead.
 * Added HTTP method parameter to `HttpCaller::getResponse()`.
+
+2.0.0-beta3
+-----------
+
+* **BC break:** The `ResponseTagger` now expects an instance of
+  `TagHeaderFormatterInterface` instead of `TagCapable` as first argument.
+* Added `HashGenerator` interface.
 
 1.4.2
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,7 +83,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 -----------
 
 * **BC break:** The `ResponseTagger` now expects an instance of
-  `TagHeaderFormatterInterface` instead of `TagCapable` as first argument.
+  `TagHeaderFormatter` instead of `TagCapable` as first argument.
 * Added `HashGenerator` interface.
 
 1.4.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * **BC break:** Removed the `Interface` suffix from all interfaces.
 * **BC break:** Renamed `HashGenerator` to `DefaultHashGenerator`.
 * Added `HashGenerator` interface.
-__
+
 ### HTTP
 
 * **BC break:** Replaced hard coupling on Guzzle HTTP client with HTTPlug. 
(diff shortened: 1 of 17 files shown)


src/TagHeaderFormatter/TagHeaderFormatterInterface.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace FOS\HttpCache\TagHeaderFormatter;
+
+/**
+ * Interface for tag header formatting.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+interface TagHeaderFormatterInterface
```
dbu: this project follows the convention of naming interfaces without the `Interface` postfix. please rename this to TagHeaderFormatter.

src/TagHeaderFormatter/TagHeaderFormatterInterface.php
```diff
@@ -0,0 +1,38 @@
...
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\TagHeaderFormatter;
+
+/**
+ * Interface for tag header formatting.
```
dbu: Lets say "The TagHeaderFormatter is used for cache tagging with HTTP headers."

doc/response-tagging.rst
```diff
@@ -14,13 +14,19 @@ Setup
 
     Make sure to :doc:`configure your proxy <proxy-configuration>` for tagging first.
 
-The response tagger is a decorator around a proxy client that implements
-the ``TagCapable`` interface, handling adding tags to responses::
+The response tagger takes an instance of ``TagHeaderFormatterInterface`` as first
+argument which is responsible for providing the header name as well as formatting
+the provided tags into a proper header value. ``HttpCache`` ships with a
```
dbu: lets say "This library ships with..." instead of "`HttpCache` ships with...", i feel thats easier to read.

doc/response-tagging.rst
```diff
@@ -14,13 +14,19 @@ Setup
...
+``CommaSeparatedTagHeaderFormatter`` which simply turns an array of tags into a
+comma-separated list but you can provide any tag header formatter you like.
```
dbu: comma-separated list. If you need a different behaviour, provide your own implementation of the `TagHeaderFormatter` interface.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Raised minimum PHP version to 5.6.
 * **BC break:** Removed the `Interface` suffix from all interfaces.
 * **BC break:** Renamed `HashGenerator` to `DefaultHashGenerator`.
+* **BC break:** The `ResponseTagger` now expects an instance of
+  `TagHeaderFormatterInterface` instead of `TagCapable` as first argument.
```
dbu: version 1 had no ResponseTagger nor the TagCapable interface, so this is not really a BC break. lets add a header 2.0.0-beta3 below, above 1.4.2 and say this there. and instead edit the Tagging section in the 2.0 change notes.

Toflar:
Most of your comments should be addressed.

> imo we should remove these methods from TagCapable and add getTagHeaderFormatter instead, and replace the tags_header option in the Varnish proxy client with a tag_header_formatter option that requires a TagHeaderFormatter.

I have removed the methods. So you think the regular expression header formatting in the Varnish client should be put into its own `VarnishTagHeaderFormatter` which is default but can be overridden using the option, did I get this right?

CHANGELOG.md
```diff
@@ -77,6 +79,13 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
   `ProxyTestCase`; use the traits `CacheAssertions` and `HttpCaller` instead.
 * Added HTTP method parameter to `HttpCaller::getResponse()`.
 
+2.0.0-beta3
+-----------
+
+* **BC break:** The `ResponseTagger` now expects an instance of
+  `TagHeaderFormatterInterface` instead of `TagCapable` as first argument.
```
dbu: TagHeaderFormatter (without the interface suffix)

doc/response-tagging.rst
```diff
@@ -14,13 +14,20 @@ Setup
 
     Make sure to :doc:`configure your proxy <proxy-configuration>` for tagging first.
 
-The response tagger is a decorator around a proxy client that implements
-the ``TagCapable`` interface, handling adding tags to responses::
+The response tagger takes an instance of ``TagHeaderFormatterInterface`` as first
```
dbu: TagHeaderFormatter without interface suffix

Toflar:
> i am a bit afraid of having two places where the tag header is configured.

I think this was the initial request, no? In the other issue the request was that one system is doing the invalidation and the other is adding the headers. The names might very well differ. It's just a matter of configuration.

> what if we add a getTagHeaderFormatter() method to TagCapable? right now when i setup, i would need to know the default header that the Varnish client is using... if the tag capable proxy client would act as "factory" for the formatter, we can keep the header name in sync. something like this:
> 
> by default varnish client creates the formatter with the default header name
> if a header name is configured, varnish client creates the formatter with the configured header name
> or you can pass a configuration option with a formatter. then varnish client takes the header name from the formatter
> if both formatter and header name are configured, we raise an error

At the beginning, I wanted to do something like this too but it feels very complicated when I look at it again. If I look at the source code (and I guess I do this pretty unbiased because I haven't been involved in v1 nor do I know all parts of this lib) I can see that the `CacheInvalidator` is the one a developer would ask to invalidate certain tags using the `invalidateTags()` method. Now I don't see why I would ever want to ask the concrete `ProxyClient` for a `TagHeaderFormatter`? I don't really care how it is invalidating the tags, it should just do it. And in addition, if you're telling me that there's no value in putting the style how `Varnish` formats it's invalidation header in its own class then I also don't see why I would ever need to pass a different `TagHeaderFormatter` if it only uses it for the header name anyway?

So the current state allows you to implement your own `TagHeaderFormatter` for response tagging. You need to configure your proxy accordingly. Custom invalidation header formatting, however, is not possible except for if you replace the whole `ProxyClient`. A custom header name is possible though. Moreover, it depends on the `ProxyClient` implementation of what is possible and what's not (which might be good because e.g. `Varnish` maybe knows there's never a different header format that's working).

So I'm not sure what I need to do here now to be honest.

CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Raised minimum PHP version to 5.6.
 * **BC break:** Removed the `Interface` suffix from all interfaces.
 * **BC break:** Renamed `HashGenerator` to `DefaultHashGenerator`.
-* Added `HashGenerator` interface.
+
```
dbu: lets keep this line to make it easier for people who update only when 2.0 stable is out

CHANGELOG.md
```diff
@@ -77,6 +79,13 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
   `ProxyTestCase`; use the traits `CacheAssertions` and `HttpCaller` instead.
 * Added HTTP method parameter to `HttpCaller::getResponse()`.
 
+2.0.0-beta3
+-----------
+
+* **BC break:** The `ResponseTagger` now expects an instance of
+  `TagHeaderFormatter` instead of `TagCapable` as first argument.
+* Added `HashGenerator` interface.
```
dbu: lets not have this line here. (i agree with the beta3 title and the bc break note on ResponseTagger, lets keep that one)

Toflar:
Ping @dbu :)

src/TagHeaderFormatter/CommaSeparatedTagHeaderFormatter.php
```diff
@@ -0,0 +1,51 @@
...
+     * @var string
+     */
+    private $headerName;
+
+    /**
+     * CommaSeparatedTagHeaderFormatter constructor.
+     *
+     * @param string $headerName
+     */
+    public function __construct($headerName)
```
dbu: the argument should default to the constant

CHANGELOG.md
```diff
@@ -77,6 +79,14 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
   `ProxyTestCase`; use the traits `CacheAssertions` and `HttpCaller` instead.
 * Added HTTP method parameter to `HttpCaller::getResponse()`.
 
+2.0.0-beta3
+-----------
+
+* **BC break:** The `ResponseTagger` now expects an instance of
+  `TagHeaderFormatter` instead of `TagCapable` as first argument.
+* **BC break:** The constant `Varnish::DEFAULT_HTTP_HEADER_CACHE_TAGS` has been
```
dbu: can we say that we moved the constant to the TagHeaderFormatter?

Toflar:
Thanks for the feedback @dbu, PR updated accordingly.

dbu:
great, thanks a lot! i finally understood what you mean that the two headers do not need to be the same! i went over more parts of the documentation and realized that the header formatter should just be an optional configuration, not another argument. i created #354 with your changes plus my documentation updates and the refactoring. can you please look over #354 if i missed anything? if you agree i will merge that one.

PR #354, opened 2017-05-23
Rework response tagger

dbu:
continuation of #353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Abstracting tags by adding new `TagCapable` for ProxyClients.
 * Added `strict` option to `ResponseTagger` that throws an exception when empty
   tags are added. By default, empty tags are ignored.
+* Added a `TagHeaderFormatter` interface that is used within the `ResponseTagger`
+  to provide the header name and the formatted tags header value.
 
 ### Varnish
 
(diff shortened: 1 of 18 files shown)


dbu:
@Toflar can you check out the additional changes i did?

CHANGELOG.md
```diff
@@ -77,6 +79,13 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
   `ProxyTestCase`; use the traits `CacheAssertions` and `HttpCaller` instead.
 * Added HTTP method parameter to `HttpCaller::getResponse()`.
 
+2.0.0-beta3
+-----------
+
+* **BC break:** The `ResponseTagger` no longer expects an instance of
+  `TagCapable` as first argument. To adjust the tag header, use the new
```
Toflar: `tag header or the way its values are formatted`

src/ProxyClient/Varnish.php
```diff
@@ -36,6 +36,15 @@ class Varnish extends HttpProxyClient implements BanCapable, PurgeCapable, Refre
...
+
+    /**
+     * Default name of the header used to invalidate content with specific tags.
+     *
+     * This happens to be the same as TagHeaderFormatter::DEFAULT_HEADER_NAME
+     * but does not technically need to be the same.
+     *
+     * @var string
+     */
     const DEFAULT_HTTP_HEADER_CACHE_TAGS = 'X-Cache-Tags';
```
Toflar: Just a note here: X-headers are deprecated ([RFC 6648](https://tools.ietf.org/html/rfc6648)), not sure if you would like to adjust them while releasing `2.0`.

Toflar:
Otherwise: LGTM!

src/ProxyClient/Varnish.php
dbu: yeah, i know. i actually started working on that, but then decided its not worth it (mainly the hard to spot BC breaks with varnish configuration). see #279

Toflar: np for me :)

src/ResponseTagger.php
```diff
@@ -32,37 +35,41 @@ class ResponseTagger
...
-    public function __construct(TagCapable $proxyClient, array $options = [])
+    public function __construct(array $options = [])
     {
-        $this->proxyClient = $proxyClient;
-
         $resolver = new OptionsResolver();
         $resolver->setDefaults([
+            // callback to avoid instantiating the formatter when its not needed
+            'header_formatter' => function (Options $options) {
+                return new CommaSeparatedTagHeaderFormatter();
```
dbu: @stof is this a good idea? i wanted to avoid the overhead of creating an object when none is needed (because the formatter is provided by the user). the documentation on options resolver mentions the callback with an Options argument, but does not mention this use case explicitly.

PR #357, opened 2017-06-10
fix hhvm build

dbu:


dbu:
ok, we seem to need trusty for varnish, and hhvm seems borked.

PR #358, opened 2017-06-10
travis builds started failing

dbu:
something with the travis setup must have changed. i restarted the previous build on master that was green some weeks ago, and now it fails.

trying to figure out what is going on


dbu:
looks like something changed on the travis boxes to put a newline into the pid file...

PR #364, opened 2017-08-09
Improve Travis build

Jean85:
With this, the coverage will be collected faster, and even under PHP 7.x; xDebug is then disabled to avoid issues.

I've also added PHP 7.2 to the matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 dist: trusty
 group: edge
 language: php
+cache:
+  directories:
+    - $HOME/.composer/cache/files
 
 php:
   - 5.6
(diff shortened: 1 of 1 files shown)


Jean85:
Also: why do we have `--prefer-source` in the composer update step? It's a lot slower and it doesn't use the cache!

If it's ok, I would like to improve that.

Jean85:
Ok I've scrapped everything, now I've:
 * added PHP 7.2 with allowed failure (but it's already green!)
 * switched from `--prefer-source` to `--prefer-dist`
 * the build got down from ~2:50 to ~1:50!!!
 * added cache, so it will be even faster!

dbu:
great, thanks a lot!

Issue #363, opened 2017-08-09
phpunit 6 support


dbu:
we provide test cases (in src/Test). those tests are not compatible with the new namespaced version of phpunit 6. we should make it compatible and provide a wrapper for phpunit 5.

Jean85:
Maybe using the forward compat layer of PHPUnit 5.4+ is enough? I will look at the code to see if it's feasible.

dbu:
sounds great. bumping to phpunit 5.4+ would be fine for me.

dbu:
and thanks for looking into it!

Jean85:
I'm sorry, but I have bad news...
Your test suite (/tests) is easy to be made cross-compatible between PHPUnit 5 and 6, see: https://github.com/Jean85/FOSHttpCache/commit/dd07fc7bb3f8df071c6c3deae58e12f90ee328c6

I cannot say the same for the tests in src/Test. Cross compatibility is not achievable due to the fact that you are extending a TestListener, which has a broken FC layer:
```php
interface TestListener extends PHPUnit_Framework_TestListener
{
}
```
This extension doesn't override or change the method's signatures, which are:
```php
public function addError(PHPUnit_Framework_Test $test, Exception $e, $time);
public function addFailure(PHPUnit_Framework_Test $test, PHPUnit_Framework_AssertionFailedError $e, $time);

// etc ...
```
As you can see, the namespaced interface still (wrongly) uses the non-namespaced classes in its signatures.

We will have to jump from PHPUnit 5 to 6 or duplicate the offending classes for disambiguation.

dbu:
could we provide our own FC layer instead?

Jean85:
Nope, it's completely not doable due to code restriction:
 * applying the namespaced classes to the methods' signature will break the interface (we would be restricting the accepted types, breaking Liskov substitution principle)
 * removing the types from the arguments will break the interface (PHP does not support contravariance, yet)

That's why I'm advocating class duplication as the only feasible solution; but I still don't know which naming convention should I use for the duplication...

lyrixx:
I strongly suggest you to use https://github.com/symfony/phpunit-bridge to ease everything related to PHPUnit.

Jean85:
As I can see, they fixed the same issue with some nice _black magic_! https://github.com/symfony/phpunit-bridge/blob/3c0efb8609a32890a767bbbd39198a0e92572694/SymfonyTestsListener.php#L19
```php
[...]

PR #365, opened 2017-08-09
Update to PHPUnit 5.4+/6.x

Jean85:
As discussed in #363, this PR aims to migrate to PHPUnit 5.4, so it will be also cross compatible with PHPUnit 6.

The hardest part was migrating the `AbstractCacheConstraint` and the `WebServerListener` classes, but @lyrixx pointed us toward a nice trick that was used in the symfony/phpunit-bridge to solve the same issue.

### DIFF
--- a/src/Test/Legacy/WebServerListener.php
+++ b/src/Test/Legacy/WebServerListener.php
@@ -16,6 +16,7 @@ use PHPUnit\Framework\TestListener;
 
 /**
  * A PHPUnit test listener that starts and stops the PHP built-in web server.
+ * This legacy version is for PHPUnit 5.x (min 5.4.4 required, due to FC layer) 
  *
  * This listener is configured with a couple of constants from the phpunit.xml
  * file. To define constants in the phpunit file, use this syntax:
--- a/src/Test/Legacy/WebServerListener.php
+++ b/src/Test/Legacy/WebServerListener.php
@@ -16,7 +16,8 @@ use PHPUnit\Framework\TestListener;
 
 /**
  * A PHPUnit test listener that starts and stops the PHP built-in web server.
- * This legacy version is for PHPUnit 5.x (min 5.4.4 required, due to FC layer) 
+ *
+ * This legacy version is for PHPUnit 5.x (min 5.4.4 required, due to FC layer).
  *
  * This listener is configured with a couple of constants from the phpunit.xml
  * file. To define constants in the phpunit file, use this syntax:
--- a/src/Test/Legacy/WebServerListener.php
+++ b/src/Test/Legacy/WebServerListener.php
@@ -1,0 +1,142 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCache package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\Test\Legacy;
+
+use FOS\HttpCache\Test\WebServerListenerTrait;
+
+/**
+ * A PHPUnit test listener that starts and stops the PHP built-in web server.
+ *
+ * This listener is configured with a couple of constants from the phpunit.xml
+ * file. To define constants in the phpunit file, use this syntax:
+ * <php>
+ *     <const name="WEB_SERVER_HOSTNAME" value="localhost" />
+ * </php>
+ *
+ * WEB_SERVER_HOSTNAME host name of the web server (required)
+ * WEB_SERVER_PORT     port to listen on (required)
+ * WEB_SERVER_DOCROOT  path to the document root for the server (required)
+ */
+class WebServerListener implements \PHPUnit_Framework_TestListener
+{
+    /** @var WebServerListenerTrait */
+    private $trait;
+
+    public function __construct()
+    {
+        $this->trait = new WebServerListenerTrait();
+    }
+
+    /**
+     * Make sure the PHP built-in web server is running for tests with group
+     * 'webserver'.
+     */
+    public function startTestSuite(\PHPUnit_Framework_TestSuite $suite)
+    {
+        $this->trait->startTestSuite($suite);
+    }
+
+    /**
+     *  We don't need these.
+     */
+    public function endTestSuite(\PHPUnit_Framework_TestSuite $suite)
+    {
+    }
+
+    public function addError(\PHPUnit_Framework_Test $test, \Exception $e, $time)
+    {
+    }
+
+    public function addFailure(\PHPUnit_Framework_Test $test, \PHPUnit_Framework_AssertionFailedError $e, $time)
+    {
+    }
+
+    public function addIncompleteTest(\PHPUnit_Framework_Test $test, \Exception $e, $time)
+    {
+    }
+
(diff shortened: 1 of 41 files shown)


lyrixx:
Why don't you add the phpunit brige ? It will help with deprecation & co ?

Jean85:
@lyrixx right now I'm focused into making the build green, I would like to avoid having issues due to an other additional change.

Later, if the maintainer are ok with that, I can add it!

Jean85:
Build is green under PHP 7.1 & 7.2, but not on 7.0 and lower... the strange thing is that between 7.0 and 7.1 the only different dependency seems to be doctrine/instantiator (1.0.5 vs 1.1.0)...

Any suggestions on how to fix this?

dbu:
great work so far, thanks! i will only have time to look at this when i am back from a long weekend next tuesday.

Ocramius:
Seems unrelated: https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/jobs/262684454#L1713

Jean85:
Just tested it, Ocramius is right, with the same deps it fails under PHP 7.0 but not under 7.1/7.2...

dbu:
now it looks like you are only left with a name clash for the class TestListener... can you try to rebase and see if you can avoid the name clash?

Jean85:
Nope, the clash is still there :disappointed:

src/Test/Legacy/WebServerListener.php
```diff
@@ -0,0 +1,143 @@
...
+ */
+
+namespace FOS\HttpCache\Test\Legacy;
+
+use FOS\HttpCache\Test\WebServerListenerTrait;
+use PHPUnit\Framework\TestListener;
+
+/**
+ * A PHPUnit test listener that starts and stops the PHP built-in web server.
+ *
```
dbu: please add a line here to say what versions of php this legacy listener is for.

src/Test/PHPUnit/AbstractCacheConstraint.php
```diff
@@ -11,71 +11,18 @@
...
-        }
-
-        return strpos((string) $other->getHeaderLine($this->header), $this->getValue()) !== false;
-    }
+use PHPUnit\Framework\Constraint\Constraint;
 
+if (class_exists('PHPUnit_Runner_Version') && version_compare(\PHPUnit_Runner_Version::id(), '6.0.0', '<')) {
+    class_alias('FOS\HttpCache\Test\Legacy\PHPUnit\AbstractCacheConstraint', 'FOS\HttpCache\Test\PHPUnit\AbstractCacheConstraint');
+// Using an early return instead of a else does not work when using the PHPUnit phar due to some weird PHP behavior (the class
+// gets defined without executing the code before it and so the definition is not properly conditional)
```
dbu: please indent this comment so that the ci is happy.

src/Test/WebServerListener.php
```diff
@@ -11,179 +11,146 @@
...
+use PHPUnit\Framework\AssertionFailedError;
+use PHPUnit\Framework\Test;
+use PHPUnit\Framework\TestListener;
+use PHPUnit\Framework\TestSuite;
+use PHPUnit\Framework\Warning;
+
+if (class_exists('PHPUnit_Runner_Version') && version_compare(\PHPUnit_Runner_Version::id(), '6.0.0', '<')) {
+    class_alias('FOS\HttpCache\Test\Legacy\WebServerListener', 'FOS\HttpCache\Test\WebServerListener');
+// Using an early return instead of a else does not work when using the PHPUnit phar due to some weird PHP behavior (the class
+// gets defined without executing the code before it and so the definition is not properly conditional)
```
dbu: please indent to make the ci happy

src/Test/Legacy/WebServerListener.php
```diff
@@ -0,0 +1,143 @@
...
+ * file. To define constants in the phpunit file, use this syntax:
+ * <php>
+ *     <const name="WEB_SERVER_HOSTNAME" value="localhost" />
+ * </php>
+ *
+ * WEB_SERVER_HOSTNAME host name of the web server (required)
+ * WEB_SERVER_PORT     port to listen on (required)
+ * WEB_SERVER_DOCROOT  path to the document root for the server (required)
+ */
+class WebServerListener implements TestListener
```
dbu: if i understand my random experiments correctly, the class_alias thing merges the `use` statements from both files. to work around, i suggest you alias the TestListener in this file to LegacyTestListener.

src/Test/WebServerListenerTrait.php
```diff
@@ -0,0 +1,144 @@
...
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\Test;
+
+/**
+ * This fake trait is used to have the same code and behavior between WebServerListener and its legacy version.
```
dbu: why fake?

Jean85: Just because it's a class, not a real `trait` :smile:

src/Test/Legacy/WebServerListener.php
```diff
@@ -0,0 +1,144 @@
...
+ */
+
+namespace FOS\HttpCache\Test\Legacy;
+
+use FOS\HttpCache\Test\WebServerListenerTrait;
+use PHPUnit\Framework\TestListener;
+
+/**
+ * A PHPUnit test listener that starts and stops the PHP built-in web server.
+ * This legacy version is for PHPUnit 5.x (min 5.4.4 required, due to FC layer)
```
dbu: styleci complains about missing `.` at the end. please also add a blank line before this line.

Jean85:
CS fixed. CI seems to be stuck due to unreachable server for the NGINX package...

dbu:
i will restart the build later today to see. guess you could squash the commits and then we are hopefully good to merge!

Jean85:
If I squash the commits now I'll trigger an other not-useful build...
I'll do it later, but if I forgot it you can do it while merging, directly from the GitHub interface!

dbu:
@hpatoio we get build errors because frickle.com times out. they are the ones providing the purge extension for nginx... do you know them, would you know if they discontinued delivering the nginx extension or if its just a technical problem and they will fix their servers?

Jean85:
I don't know thoroughly your CI process, but shouldn't be easier to migrate it to Docker containers? A cache purge could be done with a container restart...

dbu:
> docker

there is #344 which is about adding a docker file. i think that would be the way forward, also to be able to install the varnish xkey plugin in #332 so that we could test that... but it seems that #344 has stalled. if you want to pick it up, i'd be happy. i'd prefer having a slightly longer build over having to host a custom docker image somewhere that then needs to be rebuilt and re-uploaded manually when one of the vendors changes.

dbu:
ok, the frickle.com server is back online. now we get only an error for the lowest version build. can you bump the minimal phpunit version some more, @Jean85 ?

Jean85:
Done... Yay the build is finally green!
Bumping up to PHPUnit 5.7.0 was the only way, the FC layer for the TestListener was added only up there.

dbu:
thanks a lot for your patience and persistence!

PR #366, opened 2017-08-17
Set the default value of queue in proxy client to an empty array

emodric:
PHP 7.2 throws a warning when `count` is called with a `null` argument which breaks Symfony apps (and presumably others too), so this makes sure the `AbstractProxyClient::$queue` is never `null` when counting it in `AbstractProxyClient::flush` method.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,9 @@ php:
   - 5.4
   - 5.5
   - 5.6
-  - 7
+  - 7.0
+  - 7.1
+  - 7.2
 
 env:
   - VARNISH_VERSION=4.0
@@ -28,7 +30,7 @@ branches:
 before_script:
   # Install deps
   - composer update $COMPOSER_FLAGS --dev --prefer-source --no-interaction
-  - sh -c "if [ '$TRAVIS_PHP_VERSION' = '7' ]; then wget https://phar.phpunit.de/phpunit-5.7.phar; fi"
+  - bash -c "if [[ '$TRAVIS_PHP_VERSION' == '7.'* ]] ; then wget https://phar.phpunit.de/phpunit-5.7.phar; fi"
   # Install Varnish
   - curl http://repo.varnish-cache.org/debian/GPG-key.txt | sudo apt-key add -
   - echo "deb http://repo.varnish-cache.org/ubuntu/ precise varnish-${VARNISH_VERSION}" | sudo tee -a /etc/apt/sources.list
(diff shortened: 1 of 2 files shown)


dbu:
looks reasonable, yes. i am trying to fix the 1.4 build in #369 - once i merged that, i will ask you to rebase this branch.

emodric:
@dbu Sure, no problem!

dbu:
ok, here we go. just fixed the build on 1.4. 

can you please also add a test that will show this now works, that would fail without your fix? php warnings break the build so just flushing an empty queue and asserting that this operates normally and does not make any invalidation requests would be good.

emodric:
Actually, it seems that `FOS\HttpCache\Tests\Unit\ProxyClient\AbstractProxyClientTest::testFlushEmpty` already fails without the patch on PHP 7.2 (it passes on 7.1 since 7.1 does not emit the warning), and the patch fixes it. I've rebased on top of 1.4, but what do you want to do about testing 7.2?

dbu:
ah ok. lets add 7.1 and 7.2 to the test matrix then. can you do it in your pull request? you will need to adjust .travisci.yml as well to use the right phpunit on all php 7 versions.

emodric:
@dbu I've added PHP 7.1 and 7.2 to the test matrix, but I'm not very well versed in shell scripting so if you have a better suggestion on how to handle 7.0, 7.1 and 7.2 versions, do tell.

.travis.yml
```diff
@@ -28,7 +30,7 @@ branches:
 before_script:
   # Install deps
   - composer update $COMPOSER_FLAGS --dev --prefer-source --no-interaction
-  - sh -c "if [ '$TRAVIS_PHP_VERSION' = '7' ]; then wget https://phar.phpunit.de/phpunit-5.7.phar; fi"
+  - sh -c "if [ '$TRAVIS_PHP_VERSION' = '7.0' ] || [ '$TRAVIS_PHP_VERSION' = '7.1' ] || [ '$TRAVIS_PHP_VERSION' = '7.2' ]; then wget https://phar.phpunit.de/phpunit-5.7.phar; fi"
```
dbu: lets try with bash: ``bash -c "if [[ '$TRAVIS_PHP_VERSION' == '7.'* ]] ; then wget https://phar.phpunit.de/phpunit-5.7.phar; fi" ``

and yes, bash is a really messed up thing :-/. the `*` is outside the quotes...

emodric:
@dbu Now it looks like as if it's randomly failing. Can you restart the build?

dbu:
yay! thanks a lot. will merge and then try to merge upstream - guess there will be conflicts with .travis.yml

i think the build on 1.4 has been flaky for a while. in 2.0 its better, so i won't bother anymore.

emodric:
Thanks @dbu :)

dbu:
i tagged 1.4.3 with this and plan to release 2.0.2 shortly, waiting for the build to run through

PR #372, opened 2017-08-19
Update to PHPUnit 5.4+/6.x

dbu:
trying to fix the problem of #365


Jean85:
I've pulled those additional commits in the original PR, #365

dbu:
integrated into #365

Issue #286, opened 2016-04-08
Symfony HTTP Cache / Tag Invalidation


dantleech:
Currently this library contains several classes related to extending the Symfony HTTP Cache by way of an EventDispatchingKernel trait.

There is a general need to implement tag invalidation with the Symfony HTTP cache, and this has been done [here](https://github.com/dantleech/sf-http-cache-tagging/tree/master/src).

Most of the logic in that package is concerned with tags and @magnusnordlander is proposing to create a [new package](https://github.com/php-tag-store/php-tag-store) which is dedicated to tag storage. Therefore the amount code required for adding the tagging functionality to the HTTP cache is not very large.

We could therefore reasonably add the tagging functionality in this package, however I am not so sure that it belongs here -  
- I believe all of the functionality we add to the Symfony HTTP Cache can be used without the infrastructure of the FOSHttpCache.
- I think this bundle should _integrate_ with third-party caches and not provide solutions itself.

I therefore propose to create a new package containing to the Symfony HTTP cache and ultimately remove the `SymfonyCache` namespace from this package.

wdyt?

dbu:
i think this is the right direction to go. 

so we would create a package for the event handling http cache without implementations?
and one for HttpCache tags? there are a couple of other events besides tags, like purge and refresh. where would those fit?

andrerom:
Fine with me as long as it remains transparent for users:
- responses needs to be the same, standardized tagging header *(& user context header, but digress)*
- ProxyClient API needs to be same across Sf, Varnsih, Akamai .. *(purge/refresh/expire/expireByTag)*

dantleech:
> so we would create a package for the event handling http cache without implementations?

I was thinking more of a single HTTPCache package with all of the features that we provide here, including the forthcoming tagging. I would think that would be easier to maintain and document etc - otherwise we should consider a new organization and have separate packages for each feature, but I don't personally think that its worth it.

> responses needs to be the same, standardized tagging header

Yeah, see no reason why we would change that.

> ProxyClient API needs to be same across Sf, Varnsih, Akamai

I imagine the ProxyClient would staying here and the proposed package would be just another third-party that we integrate with.

dbu:
i don't think we should micro-split features i think. the only question for me is if we want to split the event emitting trait out into a separate library. but on the other hand we can still do that when people ask for it and keep things simple for now.

we could add this in the organization that @magnusnordlander created, maybe?

[...]

PR #373, opened 2017-08-31
Symfony taggable cache

Toflar:
Proposal for https://github.com/FriendsOfSymfony/FOSHttpCache/issues/286.
I actually did most of the work already a few months ago when trying to bring it to Symfony core so the `TaggableStore` was 80% ready and covered by all kinds of unit tests already 😄 
It has a lot of unit tests, especially for proper `Vary` handling which I implemented quite different from what the Symfony default `Store` does. I did not use it in a project yet, so I have no idea if everything works as expected, I'm just pretty sure because of the unit tests. So it would be great if someone that actually has a project where he/she could use cache invalidation with tags could test this integration and see if the docs are clear 😄 

I'll add a few comments/annotations to my own code in a few minutes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
         "symfony/process": "^2.3||^3.0",
         "symfony/http-kernel": "^2.3||^3.0",
         "symfony/lock": "^3.4@dev",
-        "symfony/cache": "^3.2",
+        "symfony/cache": "^3.4@dev",
         "phpunit/phpunit": "^5.7 || ^6.0"
     },
     "suggest": {
--- a/composer.json
+++ b/composer.json
@@ -36,6 +36,8 @@
         "php-http/mock-client": "^0.3.2",
         "symfony/process": "^2.3||^3.0",
         "symfony/http-kernel": "^2.3||^3.0",
+        "symfony/lock": "^3.4@dev",
+        "symfony/cache": "^3.2",
         "phpunit/phpunit": "^5.7 || ^6.0"
     },
     "suggest": {
--- a/doc/response-tagging.rst
+++ b/doc/response-tagging.rst
@@ -19,9 +19,20 @@ The response tagger uses an instance of ``TagHeaderFormatter`` to know the
 header name used to mark tags on the content and to format the tags into the
 correct header value. This library ships with a
 ``CommaSeparatedTagHeaderFormatter`` that formats an array of tags into a
-comma-separated list. This format is expected for invalidation with the
-Varnish reverse proxy. When using the default settings, everything is created
-automatically and the ``X-Cache-Tags`` header will be used::
+comma-separated list. What kind of format you need obviously depends on what
+proxy you use and how this proxy is configured (if it even can be configured).
+It is important you do understand that the ``ResponseTagger`` is not directly
+linked to the proxy configuration so you are the one that has to make sure, they
+match. Don't worry too much though, the default settings will work out of the box
+and if you really need to adjust formatting, header names etc. you likely have
+some special environment that needs special adjustments. So for now, just remember
+you can configure all the stuff but you don't have to.
+
+For example, the :doc:`default configuration of Varnish <varnish-configuration>` in
+this documentation requires the header to contain a comma-separated list of tags
+and the default examples use the header ``X-Cache-Tags``. If you don't change
+the ``TagHeaderFormatter`` nor the header name, everything will be created
+automatically and work out of the box for you::
 
     use FOS\HttpCache\ResponseTagger;
 
@@ -29,9 +40,9 @@ automatically and the ``X-Cache-Tags`` header will be used::
 
 .. _response_tagger_optional_parameters:
 
-If you need a different behavior, you can provide your own implementation of
-the ``TagHeaderFormatter`` interface. But be aware that your
-:ref:`Varnish configuration <varnish_tagging>` has to match with the tag on the response.
+As explained already, if you need a different behavior, you can provide your own
+implementation of the ``TagHeaderFormatter`` interface. But again, be aware that your
+:doc:`proxy configuration <proxy-configuration>` has to match with the response.
 For example, to use a different header name::
 
     use FOS\HttpCache\ResponseTagger;
--- a/doc/response-tagging.rst
+++ b/doc/response-tagging.rst
@@ -40,9 +40,9 @@ automatically and work out of the box for you::
 
 .. _response_tagger_optional_parameters:
 
-As explained already, if you need a different behavior, you can provide your own
-implementation of the ``TagHeaderFormatter`` interface. But again, be aware that your
-:doc:`proxy configuration <proxy-configuration>` has to match with the response.
+If you need a different behavior, you can provide your own
+``TagHeaderFormatter`` instance. Be aware that your
+:doc:`proxy configuration <proxy-configuration>` has to match the response.
 For example, to use a different header name::
 
     use FOS\HttpCache\ResponseTagger;
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -300,7 +300,7 @@ filling up your file system without ever cleaning up expired cache entries.
 The `TaggableStore` counts all the cache write operations (so fetching items
 from the cache is not slowed down) and after reaching a configurable
 threshold (default `500`), it prunes expired data. If you want to disable
-pruning, you can set the threshold to `0`.
+pruning, you can set the option `prune_threshold` to `0`.
 This means that after every `500` HTTP cache writes, your file system directory
 will be cleaned up and thus kept in good shape.
 You can configure the prune threshold by providing a different threshold as
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -275,40 +275,40 @@ The TaggableStore
 
 .. versionadded:: 2.1
 
-    The `TaggableStore` has been added in version 2.1.
+    The ``TaggableStore`` has been added in version 2.1.
 
 .. warning::
 
-    You need at least versions 3.4 of `symfony/cache` and `symfony/lock`
-    to use this feature! Add the following lines to your `composer.json` and run
-    `composer update`::
+    You need at least versions 3.4 of ``symfony/cache`` and ``symfony/lock``
+    to use this feature! Add the following lines to your ``composer.json`` and run
+    ``composer update``::
 
-    "symfony/lock": "^3.4",
-    "symfony/cache": "^3.4",
+        "symfony/lock": "^3.4",
+        "symfony/cache": "^3.4",
 
 
-Symfony's `HttpCache` does not support tags based cache invalidation by default.
-However, this library ships with a `TaggableStore` and a corresponding
-`PurgeTagsListener` which provide this functionality.
+Symfony's ``HttpCache`` does not support tags based cache invalidation by default.
+However, this library ships with a ``TaggableStore`` and a corresponding
+``PurgeTagsListener`` which provide this functionality.
 Even if you do not want to invalidate cache entries by tags, you might be
-interested in using the `TaggableStore` instead of the default `Store`
+interested in using the ``TaggableStore`` instead of the default ``Store``
 implementation Symfony ships with.
 
-That's because `TaggableStore` also prunes expired entries on a regular basis
-which is something the default `Store` does not. The default `Store` keeps
+That's because ``TaggableStore`` also prunes expired entries on a regular basis
+which is something the default ``Store`` does not. The default ``Store`` keeps
 filling up your file system without ever cleaning up expired cache entries.
-The `TaggableStore` counts all the cache write operations (so fetching items
+The ``TaggableStore`` counts all the cache write operations (so fetching items
 from the cache is not slowed down) and after reaching a configurable
-threshold (default `500`), it prunes expired data. If you want to disable
-pruning, you can set the option `prune_threshold` to `0`.
-This means that after every `500` HTTP cache writes, your file system directory
+threshold (default ``500``), it prunes expired data. If you want to disable
+pruning, you can set the option ``prune_threshold`` to ``0``.
+This means that after every ``500`` HTTP cache writes, your file system directory
 will be cleaned up and thus kept in good shape.
 You can configure the prune threshold by providing a different threshold as
-second argument to the constructor of `TaggableStore`.
+second argument to the constructor of ``TaggableStore``.
 
-For this to work, you have to use the `TaggableStore` in your kernel.
+For this to work, you have to use the ``TaggableStore`` in your kernel.
 If you want support for tag based cache invalidation, you also need to register
-the `PurgeListener` so your `AppCache` should end up looking like this::
+the ``PurgeListener`` so your ``AppCache`` should end up looking like this::
 
     use FOS\HttpCache\SymfonyCache\TaggableStore();
     use FOS\HttpCache\SymfonyCache\PurgeTagsListener();
@@ -330,7 +330,7 @@ the `PurgeListener` so your `AppCache` should end up looking like this::
         $this->addSubscriber(new PurgeTagsListener());
     }
 
-The `TaggableStore` can be configured by passing an array of `$options` as a
+The ``TaggableStore`` can be configured by passing an array of ``$options`` as a
 second argument:
 
 * **prune_threshold**: Configure the number of write actions until the
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -330,6 +330,32 @@ the `PurgeListener` so your `AppCache` should end up looking like this::
         $this->addSubscriber(new PurgeTagsListener());
     }
 
+The `TaggableStore` can be configured by passing an array of `$options` as a
+second argument:
+
+* **prune_threshold**: Configure the number of write actions until the
+  store will prune the expired cache entries. Pass 0 if you want to disable
+  automated pruning.
+  Type: int
+
+* **purge_tags_header**: The HTTP header name used to check for tags
+  Type: string
+
+* **cache**: The cache adapter.
+  Use this option if you want to use a different cache implementation than the
+  default one.
+  Note that there are very good reasons that the local adapters are used by
+  default. This is to protect you as a developer! Only override it if you're
+  really sure your cache implementation meets the needs of Symfony's HttpCache.
+  Type: TagAwareAdapterInterface
+
+* **lock_factory**: The lock factory.
+  Use this option if you want to use a different lock implementation than the
+  default one.
+  Note that there are very good reasons that the local adapters are used by
+  default. This is to protect you as a developer! Only override it if you're
+  really sure your lock implementation meets the needs of Symfony's HttpCache.
+  Type: Factory
 
 Cleaning the Cookie Header
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -30,6 +30,8 @@ Using the trait
 Your ``AppCache`` needs to implement ``CacheInvalidation`` and use the
 trait ``FOS\HttpCache\SymfonyCache\EventDispatchingHttpCache``::
 
+    // app/AppCache.php
+
     use FOS\HttpCache\SymfonyCache\CacheInvalidation;
     use FOS\HttpCache\SymfonyCache\EventDispatchingHttpCache;
     use Symfony\Component\HttpFoundation\Request;
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -180,6 +180,115 @@ You can configure the listener just the same as the `PurgeListener` plus the
 
   **default**: ``X-Cache-Tags``
 
+If you want support for tag based cache invalidation, you need to register both,
+the ``PurgeListener`` and the ``TaggableStore`` so your ``AppCache`` should end
+up looking like this::
+
+    use FOS\HttpCache\SymfonyCache\TaggableStore();
+    use FOS\HttpCache\SymfonyCache\PurgeTagsListener();
+
+    // ...
+
+    /**
+     * Overwrite constructor to register the TaggableStore.
+     */
+    public function __construct(
+        HttpKernelInterface $kernel,
+        SurrogateInterface $surrogate = null,
+        array $options = []
+    ) {
+        $store = new TaggableStore($kernel->getCacheDir());
+
+        parent::__construct($kernel, $store, $surrogate, $options);
+
+        $this->addSubscriber(new PurgeTagsListener());
+    }
+
+
+.. _taggablestore:
+
+The TaggableStore
+^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.1
+
+    The ``TaggableStore`` has been added in version 2.1.
+
+.. warning::
+
+    You need at least versions 3.4 of ``symfony/cache`` and ``symfony/lock``
+    to use this feature! Add the following lines to your ``composer.json`` and run
+    ``composer update``::
+
+        "symfony/lock": "^3.4",
+        "symfony/cache": "^3.4",
+
+
+Symfony's ``HttpCache`` does not support tags based cache invalidation by default.
+However, this library ships with a ``TaggableStore`` and a corresponding
+``PurgeTagsListener`` which provide this functionality.
+Even if you do not want to invalidate cache entries by tags, you might be
+interested in using the ``TaggableStore`` instead of the default ``Store``
+implementation Symfony ships with.
+
+That's because ``TaggableStore`` also prunes expired entries on a regular basis
+which is something the default ``Store`` does not. The default ``Store`` keeps
+filling up your file system without ever cleaning up expired cache entries.
+The ``TaggableStore`` counts all the cache write operations (so fetching items
+from the cache is not slowed down) and after reaching a configurable
+threshold (default ``500``), it prunes expired data. If you want to disable
+pruning, you can set the option ``prune_threshold`` to ``0``.
+This means that after every ``500`` HTTP cache writes, your file system directory
+will be cleaned up and thus kept in good shape.
+You can configure the prune threshold by providing a different threshold as
+second argument to the constructor of ``TaggableStore``.
+
+For this to work, you have to use the ``TaggableStore`` in your kernel::
+
+    use FOS\HttpCache\SymfonyCache\PurgeTagsListener();
+
+    // ...
+
+    /**
+     * Overwrite constructor to register the TaggableStore.
+     */
+    public function __construct(
+        HttpKernelInterface $kernel,
+        SurrogateInterface $surrogate = null,
+        array $options = []
+    ) {
+        $store = new TaggableStore($kernel->getCacheDir());
+
+        parent::__construct($kernel, $store, $surrogate, $options);
+    }
+
+The ``TaggableStore`` can be configured by passing an array of ``$options`` as a
+second argument:
+
+* **prune_threshold**: Configure the number of write actions until the
+  store will prune the expired cache entries. Pass 0 if you want to disable
+  automated pruning.
+  Type: int
+
+* **purge_tags_header**: The HTTP header name used to check for tags
+  Type: string
+
+* **cache**: The cache adapter.
+  Use this option if you want to use a different cache implementation than the
+  default one.
+  Note that there are very good reasons that the local adapters are used by
+  default. This is to protect you as a developer! Only override it if you're
+  really sure your cache implementation meets the needs of Symfony's HttpCache.
+  Type: Symfony\Component\Cache\Adapter\TagAwareAdapterInterface
+
+* **lock_factory**: The lock factory.
+  Use this option if you want to use a different lock implementation than the
+  default one.
+  Note that there are very good reasons that the local adapters are used by
+  default. This is to protect you as a developer! Only override it if you're
+  really sure your lock implementation meets the needs of Symfony's HttpCache.
+  Type: Symfony\Component\Lock\Factory
+
 Refresh
 ~~~~~~~
 
@@ -266,96 +375,6 @@ options through the constructor:
 
         RewriteEngine On
         RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
-
-
-.. _taggablestore:
-
-The TaggableStore
-^^^^^^^^^^^^^^^^^
-
-.. versionadded:: 2.1
-
-    The ``TaggableStore`` has been added in version 2.1.
-
-.. warning::
-
-    You need at least versions 3.4 of ``symfony/cache`` and ``symfony/lock``
-    to use this feature! Add the following lines to your ``composer.json`` and run
-    ``composer update``::
-
-        "symfony/lock": "^3.4",
-        "symfony/cache": "^3.4",
-
-
-Symfony's ``HttpCache`` does not support tags based cache invalidation by default.
-However, this library ships with a ``TaggableStore`` and a corresponding
-``PurgeTagsListener`` which provide this functionality.
-Even if you do not want to invalidate cache entries by tags, you might be
-interested in using the ``TaggableStore`` instead of the default ``Store``
-implementation Symfony ships with.
-
-That's because ``TaggableStore`` also prunes expired entries on a regular basis
-which is something the default ``Store`` does not. The default ``Store`` keeps
-filling up your file system without ever cleaning up expired cache entries.
-The ``TaggableStore`` counts all the cache write operations (so fetching items
-from the cache is not slowed down) and after reaching a configurable
-threshold (default ``500``), it prunes expired data. If you want to disable
-pruning, you can set the option ``prune_threshold`` to ``0``.
-This means that after every ``500`` HTTP cache writes, your file system directory
-will be cleaned up and thus kept in good shape.
-You can configure the prune threshold by providing a different threshold as
-second argument to the constructor of ``TaggableStore``.
-
-For this to work, you have to use the ``TaggableStore`` in your kernel.
-If you want support for tag based cache invalidation, you also need to register
-the ``PurgeListener`` so your ``AppCache`` should end up looking like this::
-
-    use FOS\HttpCache\SymfonyCache\TaggableStore();
-    use FOS\HttpCache\SymfonyCache\PurgeTagsListener();
-
-    // ...
-
-    /**
-     * Overwrite constructor to register the TaggableStore.
-     */
-    public function __construct(
-        HttpKernelInterface $kernel,
-        SurrogateInterface $surrogate = null,
-        array $options = []
-    ) {
-        $store = new TaggableStore($kernel->getCacheDir());
-
-        parent::__construct($kernel, $store, $surrogate, $options);
-
-        $this->addSubscriber(new PurgeTagsListener());
-    }
-
-The ``TaggableStore`` can be configured by passing an array of ``$options`` as a
-second argument:
-
-* **prune_threshold**: Configure the number of write actions until the
-  store will prune the expired cache entries. Pass 0 if you want to disable
-  automated pruning.
-  Type: int
-
-* **purge_tags_header**: The HTTP header name used to check for tags
-  Type: string
-
-* **cache**: The cache adapter.
-  Use this option if you want to use a different cache implementation than the
-  default one.
-  Note that there are very good reasons that the local adapters are used by
-  default. This is to protect you as a developer! Only override it if you're
-  really sure your cache implementation meets the needs of Symfony's HttpCache.
-  Type: TagAwareAdapterInterface
-
-* **lock_factory**: The lock factory.
-  Use this option if you want to use a different lock implementation than the
-  default one.
-  Note that there are very good reasons that the local adapters are used by
-  default. This is to protect you as a developer! Only override it if you're
-  really sure your lock implementation meets the needs of Symfony's HttpCache.
-  Type: Factory
 
 Cleaning the Cookie Header
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -299,7 +299,8 @@ which is something the default `Store` does not. The default `Store` keeps
 filling up your file system without ever cleaning up expired cache entries.
 The `TaggableStore` counts all the cache write operations (so fetching items
 from the cache is not slowed down) and after reaching a configurable
-threshold (default `500`), it prunes expired data.
+threshold (default `500`), it prunes expired data. If you want to disable
+pruning, you can set the threshold to `0`.
 This means that after every `500` HTTP cache writes, your file system directory
 will be cleaned up and thus kept in good shape.
 You can configure the prune threshold by providing a different threshold as
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -131,28 +131,6 @@ other hosts, provide the IPs of the machines allowed to purge, or provide a
 RequestMatcher that checks for an Authorization header or similar. *Only set
 one of ``client_ips`` or ``client_matcher``*.
 
-Symfony's `HttpCache` does not support cache invalidation by tags by default.
-However, this library ships with a `TaggableStore` that provides exactly that.
-If you want to use this, adjust your `AppCache` as follows::
-
-    use FOS\HttpCache\SymfonyCache\TaggableStore();
-
-    // ...
-
-    /**
-     * Overwrite constructor to register the TaggableStore.
-     */
-    public function __construct(
-        HttpKernelInterface $kernel,
-        StoreInterface $store,
-        SurrogateInterface $surrogate = null,
-        array $options = []
-    ) {
-        $store = new TaggableStore($kernel->getCacheDir());
-
-        parent::__construct($kernel, $store, $surrogate, $options);
-    }
-
 
 * **client_ips**: String with IP or array of IPs that are allowed to
   purge the cache.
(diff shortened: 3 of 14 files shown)


composer.json
```diff
@@ -36,6 +36,8 @@
         "php-http/mock-client": "^0.3.2",
         "symfony/process": "^2.3||^3.0",
         "symfony/http-kernel": "^2.3||^3.0",
+        "symfony/lock": "^3.4@dev",
```
Toflar: These are `require-dev`.

doc/symfony-cache-configuration.rst
```diff
@@ -129,6 +129,29 @@ other hosts, provide the IPs of the machines allowed to purge, or provide a
...
+    /**
+     * Overwrite constructor to register the TaggableStore.
+     */
+    public function __construct(
+        HttpKernelInterface $kernel,
+        StoreInterface $store,
+        SurrogateInterface $surrogate = null,
+        array $options = []
+    ) {
+        $store = new TaggableStore($kernel->getCacheDir());
```
Toflar: Not sure if that's the best way to document but basically you just have to provide `TaggableStore` instead of having Symfony calling `createStore()`.

src/ProxyClient/Symfony.php
```diff
@@ -52,10 +53,26 @@ public function refresh($url, array $headers = [])
     protected function configureOptions()
     {
         $resolver = parent::configureOptions();
-        $resolver->setDefaults([
```
Toflar: I've unified here so it matches the `PurgeListener` options.

src/ProxyClient/Symfony.php
```diff
@@ -52,10 +53,26 @@ public function refresh($url, array $headers = [])
...
+    /**
+     * Remove/Expire cache objects based on cache tags.
+     *
+     * @param array $tags Tags that should be removed/expired from the cache
+     *
+     * @return $this
+     */
+    public function invalidateTags(array $tags)
+    {
+        // TODO: how should we escape best here?
```
Toflar: I would like to escape `,` and maybe also `new line` but we need to unescape again in the `PurgeListener` so I'm not sure if using the same method as `Varnish` does is a good idea. Especially because `Varnish` just escapes by replacing both with a `_` which makes it impossible to unescape again. To be honest, I find tags containing a `,` strange anyway so I'd prohibit this in the `ResponseTagger` upfront so it cannot happen at all but I guess that would be a BC break. Wdyt?

src/SymfonyCache/PurgeListener.php
```diff
@@ -81,7 +93,20 @@ public function handlePurge(CacheEvent $event)
...
 
         $response = new Response();
-        if ($event->getKernel()->getStore()->purge($request->getUri())) {
+        $store = $event->getKernel()->getStore();
+
+        if ($request->headers->has($this->purgeTagsHeader)
+            && $store instanceof TaggableStore
+        ) {
+            // TODO: need to unescape again here
+            $tags = explode(',', $request->headers->get($this->purgeTagsHeader));
```
Toflar: See my comment about unescaping in `Symfony`.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+
+        $this->purgeTagsHeader = $purgeTagsHeader;
+        $this->cache = $cache = new TagAwareAdapter(new FilesystemAdapter('fos-http-cache', 0, $cacheDir));
+        $this->lockFactory = new Factory(new FlockStore($cacheDir));
+    }
+
+    /**
+     * @return TagAwareAdapter
+     */
+    public function getCache()
```
Toflar: Basically this is only needed for unit tests but it doesn't harm either.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+use Symfony\Component\Lock\Factory;
+use Symfony\Component\Lock\LockInterface;
+use Symfony\Component\Lock\Store\FlockStore;
+
+/**
+ * Implements a storage for Symfony's HttpCache that supports tagging.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class TaggableStore implements StoreInterface
```
Toflar: Do you like that name? I wanted it to remain simple. The one the Symfony core provides is called `Store` so I just simply went for `TaggableStore`. Note that there's quite some stuff going on here. Proof that it works is a pretty extensive unit test suite for that class 😄

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+     * @param Request $request A Request instance
+     *
+     * @return bool|string true if the lock is acquired, the path to the current lock otherwise
+     */
+    public function lock(Request $request)
+    {
+        $cacheKey = $this->getCacheKey($request);
+
+        if (isset($this->locks[$cacheKey])) {
+            return false;
```
jderusse: why not trying to acquire this lock ?

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    /**
+     * Cleanups storage.
+     */
+    public function cleanup()
+    {
+        /** @var LockInterface $lock */
+        foreach ($this->locks as $lock) {
+            $lock->release();
```
jderusse: you should wrap in a try/catch block

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+     */
+    public function __construct($cacheDir, $purgeTagsHeader = PurgeListener::DEFAULT_PURGE_TAGS_HEADER)
+    {
+        if (!class_exists(Factory::class)) {
+            throw new \RuntimeException('You need at least Symfony 3.4 to use the TaggableStore.');
+        }
+
+        $this->purgeTagsHeader = $purgeTagsHeader;
+        $this->cache = $cache = new TagAwareAdapter(new FilesystemAdapter('fos-http-cache', 0, $cacheDir));
+        $this->lockFactory = new Factory(new FlockStore($cacheDir));
```
jderusse: The issue with Flock, is that we need to create an empty file for each key, and don't delete theme (for race condition).

Given we will create a lot of keys here, it will consume a lot of inode. I suggest tu use semaphore here and fallback to flock if semaphore is not supported. See https://github.com/symfony/symfony/blob/3ff0ff02accafaeb6f35eca5323bcb6be0f3311e/src/Symfony/Component/Console/Command/LockableTrait.php#L46-L50

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    /**
+     * @return TagAwareAdapter
```
jderusse: what's about `TagAwareAdapterInterface`

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+     * @param string $cacheDir
+     */
+    public function __construct($cacheDir, $purgeTagsHeader = PurgeListener::DEFAULT_PURGE_TAGS_HEADER)
+    {
+        if (!class_exists(Factory::class)) {
+            throw new \RuntimeException('You need at least Symfony 3.4 to use the TaggableStore.');
+        }
+
+        $this->purgeTagsHeader = $purgeTagsHeader;
+        $this->cache = $cache = new TagAwareAdapter(new FilesystemAdapter('fos-http-cache', 0, $cacheDir));
```
jderusse: Adapter could be injected instead of built (if you inject it, you should inject the lock too, because the injected cache can be a shared cache like redis, and the Lock should be shared too)

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    /**
+     * Locks the cache for a given Request.
+     *
+     * @param Request $request A Request instance
+     *
+     * @return bool|string true if the lock is acquired, the path to the current lock otherwise
+     */
+    public function lock(Request $request)
```
jderusse: seems to  be not used

src/SymfonyCache/TaggableStore.php
Toflar: Both are not injectable on purpose. Otherwise it would be a fully PSR-6 compatible `Store` which was rejected in Symfony core as no other adapter except for file makes sense and might be dangerous using because of concurrency. See the discussion in https://github.com/FriendsOfSymfony/FOSHttpCache/issues/286 and the referenced PR in Symfony.

src/SymfonyCache/TaggableStore.php
Toflar: Yeah, that's a general problem that is not specific to this implementation (it's the same in the default `Store` of `HttpCache`). I'll add semaphore though, thanks for the hint!

src/SymfonyCache/TaggableStore.php
Toflar: See my comment in `__construct()`. We use a hardcoded implementation because we don't want other adapters to be used :)

src/SymfonyCache/TaggableStore.php
Toflar: What do you mean? It's part of the `StoreInterface`.

src/SymfonyCache/TaggableStore.php
Toflar: Basically, I checked the implementation of `Store` (see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/HttpCache/Store.php#L69) so that's why it's the way it is. How should that whole method look like in your opinion?

src/SymfonyCache/TaggableStore.php
Toflar: True, will add!

src/SymfonyCache/TaggableStore.php
jderusse: You can TypeHint with TagAwareInterface ? What if someone want to use redis tombe share cache between multiple front?

src/SymfonyCache/TaggableStore.php
jderusse: Oops.. nevermind

src/SymfonyCache/TaggableStore.php
jderusse: So if you lock then release. You'll never be able to lock again... Is it the expected behavior?

src/SymfonyCache/TaggableStore.php
Toflar: I guess you did not really understand: The users **must not** use any other adapter like redis etc. It **must** be forbidden. That's why it's hardcoded to the adapters that make sense.

src/SymfonyCache/TaggableStore.php
Toflar: I guess not :) You mean either we drop the `isset()` there or if we keep it as is, we need to `unset($this->locks[$cacheKey]);` in `unlock()`? :)

jderusse: Yep, either unset in the release method. Or try to acquire it here. As you want

src/SymfonyCache/TaggableStore.php
jderusse: The point of Fabien is to not use PSR6 and simply use `$cached = include(...)`. (His comment give 3 good reasons).
As soon as you include cache component and use PSR6, I don't understand why adapters MUST be the Filesystem?

Toflar: Here's the comment: https://github.com/symfony/symfony/pull/20061#issuecomment-313339092

Quote: 

> The cache contains some PHP (when using ESI for instance) and storing PHP in anything else than a filesystem would mean eval()ing strings coming from Redis/Memcache/...,

So why would you allow anything else than `Filesystem`?

src/SymfonyCache/TaggableStore.php
Toflar: Fixed in 486008e47250810ebd777de688f6cb7401e38033.

src/SymfonyCache/TaggableStore.php
Toflar: Fixed and covered by unit test in 5f7a1ea55b7592c3f8d0e3b3996a655ffbbb6edc.

src/SymfonyCache/TaggableStore.php
Toflar: Added in 2f4b53497692b953fced5505ae5a151789419962.

src/SymfonyCache/TaggableStore.php
jderusse: Because here you use a cache adapter on top of filesystem and you must deal with the same issue.

jderusse: As soon as you start using PSR6, you are dealing with the same interface (same output, same evaluate things...) The only difference between adapter are performance vs scalability

Toflar: Hmm, I still don't get your point, sorry 😄  So trying to sum up: There was a pull request for Symfony core allowing to use any PSR-6 cache back end. It was rejected because it would allow to use Redis, Memcache etc. as cache back ends. Ergo: everything except for a filesystem cache does not make sense. However, for me this does not mean we cannot use the cache component. We just have to make sure nobody can use anything else than the `Filesystem` adapter.

doc/symfony-cache-configuration.rst
```diff
@@ -129,6 +129,29 @@ other hosts, provide the IPs of the machines allowed to purge, or provide a
 RequestMatcher that checks for an Authorization header or similar. *Only set
 one of ``client_ips`` or ``client_matcher``*.
 
+Symfony's `HttpCache` does not support cache invalidation by tags by default.
+However, this library ships with a `TaggableStore` that provides exactly that.
+If you want to use this, adjust your `AppCache` as follows::
+
```
dbu: can you add a comment that specifies the filename? i guess `// app/AppCache.php`?

doc/symfony-cache-configuration.rst
```diff
@@ -129,6 +129,29 @@ other hosts, provide the IPs of the machines allowed to purge, or provide a
...
+    use FOS\HttpCache\SymfonyCache\TaggableStore();
+
+    // ...
```
dbu: please have the `class AppCache extends HttpCache` line here

doc/symfony-cache-configuration.rst
```diff
@@ -129,6 +129,29 @@ other hosts, provide the IPs of the machines allowed to purge, or provide a
...
+
+    /**
+     * Overwrite constructor to register the TaggableStore.
+     */
+    public function __construct(
+        HttpKernelInterface $kernel,
+        StoreInterface $store,
+        SurrogateInterface $surrogate = null,
+        array $options = []
```
dbu: where do these parameters come from?

doc/symfony-cache-configuration.rst
```diff
@@ -129,6 +129,29 @@ other hosts, provide the IPs of the machines allowed to purge, or provide a
...
+    use FOS\HttpCache\SymfonyCache\TaggableStore();
+
+    // ...
+
+    /**
+     * Overwrite constructor to register the TaggableStore.
+     */
+    public function __construct(
+        HttpKernelInterface $kernel,
+        StoreInterface $store,
```
dbu: can we remove this parameter? its never used as you then create the store manually.

src/ProxyClient/Symfony.php
dbu: as far as i remember the logic with escaping was that we don't care about unescaping. the worst case would be that foo,bar and foo_bar collide, which would mean we invalidate too many things. its an edge case and one we decided we can live with, as its semantically correct (invalidating more then necessary is ok, the opposite would be a problem)

this works as long as escaping is done both when setting the tags and when creating invalidation requests.

src/SymfonyCache/PurgeListener.php
```diff
@@ -81,7 +93,20 @@ public function handlePurge(CacheEvent $event)
         }
 
         $response = new Response();
-        if ($event->getKernel()->getStore()->purge($request->getUri())) {
+        $store = $event->getKernel()->getStore();
+
+        if ($request->headers->has($this->purgeTagsHeader)
+            && $store instanceof TaggableStore
```
dbu: should we not fail if there is a request to purge tags but we don't have a store supporting purge?

src/SymfonyCache/PurgeListener.php
```diff
@@ -81,7 +93,20 @@ public function handlePurge(CacheEvent $event)
         }
 
         $response = new Response();
-        if ($event->getKernel()->getStore()->purge($request->getUri())) {
+        $store = $event->getKernel()->getStore();
+
+        if ($request->headers->has($this->purgeTagsHeader)
```
dbu: why do we do tag invalidation in the same listener as purging? i feel like it should be a separate listener. in varnish, its glued into the ban handling because its a ban operation. for httpcache it seems to make sense to consider it not a BAN because its not a blacklist but a direct action onto the cache store. but purge sounds wrong to me too, because the path of the tag invalidation request does not matter. could we use a separate HTTP method like PURGETAGS? (this is an open question, i am not convinced it would be the best. but the mixup looks odd to me here)

src/SymfonyCache/TaggableStore.php
dbu: i would use reflection in the unit test rather than expose this. we don't want people to interact with the tag cache from outside.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,400 @@
...
+        $cacheKey = $this->getCacheKey($request);
+
+        if (!isset($this->locks[$cacheKey])) {
+            return false;
+        }
+
+        try {
+            $this->locks[$cacheKey]->release();
+            unset($this->locks[$cacheKey]);
+        } catch (LockReleasingException $e) {
```
dbu: should we still unset if we get an exception here?

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,400 @@
...
+     */
+    public function purge($url)
+    {
+        $cacheKey = $this->getCacheKey(Request::create($url));
+
+        return $this->cache->deleteItem($cacheKey);
+    }
+
+    /**
+     * Cleanups storage.
```
dbu: i am a bit confused by the name of the method, and this non-telling phpdoc. but just noticed that its the same on the interface. lets do `{@inheritDoc}` and then add a line `Release all locks.`

doc/symfony-cache-configuration.rst
dbu: or alternatively make $store optional and create it below if not specified.

src/ProxyClient/Symfony.php
Toflar: I see. So I'll just escape and don't care about unescaping, right? It's an edge case anyway, I really don't know why I would ever use a comma in a tag.
We could, however, also use the same header multiple times. There's nothing wrong in using

```
X-Cache-Tags: foobar
X-Cache-Tags: other tag
```

But I don't know if http-plug supports this?

src/SymfonyCache/TaggableStore.php
Toflar: I agree. Fixed in d56f6a43b573eb72bd7733bdb02785547a6e5df1.

doc/symfony-cache-configuration.rst
Toflar: There are 2-3 examples before that all talk about the `AppCache`, I think it's redundant information :)

doc/symfony-cache-configuration.rst
Toflar: Same here, I guess the people know about `app/AppCache.php` but I added a hint in 0e3d337e5e1e43cfc802d2ef74997c82432c5c38 anyway (for the first example to not mess up with all other examples).

doc/symfony-cache-configuration.rst
Toflar: Yeah well, these parameters depend on whether you use the `HttpCache` of the `FrameworkBundle` or `HttpKernel` directly. Not sure what to do here best but I mean, it's kind of obvious what you have to do there, isn't it? 😄

doc/symfony-cache-configuration.rst
Toflar: I took the very same example as already used in the `Cache event listeners` section.

src/ProxyClient/Symfony.php
dbu: i'd replace all `,` with `_` but then collapse the tags into one header with , separation. in the listener, you can then still split by `,`.

if i remember correctly, the varnish client even splits this into multiple requests if the list of tags is getting too long, to avoid problems with too long headers.

doc/symfony-cache-configuration.rst
dbu: ah ok, the section above does the same. i find it still useful, but am ok with consistency here for now. might go over that doc at some point again.

src/SymfonyCache/PurgeListener.php
Toflar: Well, I can just tell what my idea behind this was: `PURGE` is not an official HTTP method by specification. If I was to pick one that exists, I would've probably gone with `DELETE`. Anyway, the spec allows to invent custom ones so yeah, we could use `PURGETAGS`. But then again, to purge, means to clean and so I checked the docs of Varnish (which I personally consider to be somewhat a market leader here) and it says:

> A purge is what happens when you pick out an object from the cache and discard it along with its variants. Usually a purge is invoked through HTTP with the method PURGE.

And "tagging" to me is an object from the cache so imho it's perfectly fine. Also having it in the same listener is simple, doesn't change any behavior at all if you don't send that header, keeps things in place. I really like it where it is to be honest.

src/SymfonyCache/PurgeListener.php
Toflar: Matter of preference. I thought about that too. Either fail or ignore silently. You prefer failing?

doc/symfony-cache-configuration.rst
dbu: $store makes sense for normal constructor but not for this class, because its unused as you define $store below in the constructor code.

doc/symfony-cache-configuration.rst
dbu: if we would use the one from FrameworkBundle, we could just overwrite the createStore method. i guess the reason is that here we are in the library and not the symfony bundle, so we don't want any extra dependencies. so this makes sense (apart from the $store parameter)

Toflar:
> i guess the most difficult question right now is about allowing to inject the cache implementation or not. i can see scenarios where it would indeed be a security problem to use memcache or redis, but i think there are also valid scenarios where its just as save as the filesystem (at least with redis which does password protect the connection)

I agree. Which is also why I supported the original PR and actually did a PR to finish it. I think it's not the task of this implementation to dictate a developer what to do. But then again it's also pretty fine if it helps you not to make stupid mistakes 😄 
I mean, I can allow injection, it's no problem...would also make the unit tests simpler and I could make the class `final` 😄

src/SymfonyCache/PurgeListener.php
dbu: its a basic setup error, not a runtime error that could occasionally happen, so yes i would explode loudly.

src/SymfonyCache/PurgeListener.php
dbu: my problem is that just depending on the tags header, this either uses the url to purge a specific page, or it uses the header to purge all tagged things, and ignores the url. so basically the two things have no code in common. it shows in the nested if/elseif/else. i would prefer a separate listener for that reason.

for the naming i agree that the term purge is correct. but as the semantics of the purge payload is completely different, i'd rather use a separate http method name.

Toflar: Okay, it's now in it's own `PurgeTagsListener` and the docs are updated accordingly.

src/SymfonyCache/PurgeListener.php
Toflar: Implemented with the new listener now.

src/ProxyClient/Symfony.php
Toflar: Implemented in b45271066225312c3752f05c574fefa20855382b. Sorry, I'm too used to phpunit, I never used mockery so it was way faster for me to use the "native" mock stuff for consecutive argument assertions :)

Toflar: Ah, and I implemented it so it's a more general solution. I thought that might be needed by other proxies too in the future. I did not adjust Varnish though because there's some regex stuff going on there and I figured that's maybe too specific.

Toflar:
I guess all the comments are addressed.
I've implemented setter injection in 16e4f22e1e6a2eb4a52f8e83add06ac6dcc0584a for both, the cache and the lock factory and added phpdoc to explain why these are setter injections and not regular constructor injections. Personally, I feel like this should protect developers from blindly injecting their own implementations but read the method description first. I think this is probably the best way to protect the devs, provide sane defaults but still allow @jderusse and everybody else to adjust what they need. What do you think?

Toflar:
I noticed in Symfony 3.4, the filesystem type cache adapters received a `prune()` method (see https://github.com/symfony/symfony/pull/23603). Unfortunately the `TagAwareAdpater` does not yet but it will soon 😄  (see https://github.com/symfony/symfony/pull/24075).
So I figured I would try to find some wanna-be-intelligent way on how to keep our store cleaned up. I count the number of `write()` calls now and if it reaches a certain threshold, the store is pruned 😄  How do you like that?

Toflar:
Unit tests failing are not related to the lib. Seems like `repo.varnish-cache.org is no more` is the problem :( @dbu could you fix that in master so I can rebase? :)

dbu:
i will look at this on monday.

doc/response-tagging.rst
```diff
@@ -19,19 +19,30 @@ The response tagger uses an instance of ``TagHeaderFormatter`` to know the
...
     use FOS\HttpCache\ResponseTagger;
 
     $responseTagger = new ResponseTagger();
 
 .. _response_tagger_optional_parameters:
 
-If you need a different behavior, you can provide your own implementation of
-the ``TagHeaderFormatter`` interface. But be aware that your
-:ref:`Varnish configuration <varnish_tagging>` has to match with the tag on the response.
+As explained already, if you need a different behavior, you can provide your own
```
dbu: lets not say that we repeat here - it feels unfriendly ;-)

doc/response-tagging.rst
```diff
@@ -19,19 +19,30 @@ The response tagger uses an instance of ``TagHeaderFormatter`` to know the
...
+implementation of the ``TagHeaderFormatter`` interface. But again, be aware that your
```
dbu: lets not add "again".

maybe "provide your own ``TagHeaderFormatter`` instance"? does not need to be a custom implementation of the interface, like in the example below. (i know, that is just reformatting, but while we touch it...)

doc/symfony-cache-configuration.rst
```diff
@@ -143,6 +146,79 @@ one of ``client_ips`` or ``client_matcher``*.
...
 
+
+Purge tags (cache invalidation using tags)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. warning::
+
+    You need at least Symfony 3.4 to use this feature!
+
+
```
dbu: extra blank line, please remove

doc/symfony-cache-configuration.rst
```diff
@@ -143,6 +146,79 @@ one of ``client_ips`` or ``client_matcher``*.
 
   **default**: ``PURGE``
 
+
+Purge tags (cache invalidation using tags)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
```
dbu: please add a `.. versionadded:: 2.1.0` that says that cache tagging for AppCache has been added in FOSHttpCache 2.1.0

doc/symfony-cache-configuration.rst
```diff
@@ -143,6 +146,79 @@ one of ``client_ips`` or ``client_matcher``*.
...
+.. warning::
+
+    You need at least Symfony 3.4 to use this feature!
```
dbu: i think this is not strictly correct. you need symfony/lock and symfony/cache at 3.4, but afaik you can use symfony 3.3 together with those components, or am i wrong?

we should tell the composer require line for those two components, because they are optional dependencies.

doc/symfony-cache-configuration.rst
```diff
@@ -143,6 +146,79 @@ one of ``client_ips`` or ``client_matcher``*.
...
+
+  **default**: ``null``
+
+* **purge_tags_method**: HTTP Method used with purge tags requests.
+
+  **default**: ``PURGETAGS``
+
+* **purge_tags_header**: HTTP Header that contains the comma-separated tags to purge.
+
+  **default**: ``X-Cache-Tags``
```
dbu: lets move this list below the code example. we should provide an example where to set the options. 

for the formatting, do we have nothing similar in the doc yet? if not, we could use something like we do for configuration options in the bundle doc: https://raw.githubusercontent.com/FriendsOfSymfony/FOSHttpCacheBundle/4fc92af83dd63bb4a2b586336711c8cf9b5af7f9/Resources/doc/reference/configuration/match.rst

doc/symfony-cache-configuration.rst
```diff
@@ -143,6 +146,79 @@ one of ``client_ips`` or ``client_matcher``*.
...
+    Symfony ships with. That's because `TaggableStore` also prunes expired
+    entries on a regular basis which is something the default `Store` does not.
+    The default `Store` keeps filling up your file system without ever cleaning
+    up expired cache entries. The `TaggableStore` counts all the cache write
+    operations (so fetching items from the cache is not slowed down) and after
+    reaching a configurable threshold (default `500`), it prunes expired data.
+    This means that after every `500` HTTP cache writes, your file system directory
+    will be cleaned up and thus kept in good shape.
+    You can configure the prune threshold by providing a different threshold as
+    second argument to the constructor of `TaggableStore`.
```
dbu: this seems a bit hidden to me. i think the taggable store should have its own documentation section somewhere on this page and the note should just reference that section.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,493 @@
...
+     * @var LockInterface[]
+     */
+    private $locks = [];
+
+    /**
+     * Constructor.
+     *
+     * @param string $cacheDir        The cache directory
+     * @param int    $pruneThreshold  The number of write operations until orphan data gets pruned (default: 500)
+     * @param string $purgeTagsHeader The tags header name
```
dbu: lets use an options resolver for the prune threshold and the header

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,493 @@
...
+     * that the local adapters are used by default and you do not inject them in
+     * the constructor but via setter injection instead. This is to protect you
+     * as a developer! Only use this method if you're really sure your cache
+     * implementation meets the needs of Symfony's HttpCache.
+     *
+     * @param TagAwareAdapterInterface $cache
+     *
+     * @return TaggableStore
+     */
+    public function setCache(TagAwareAdapterInterface $cache)
```
dbu: lets move this into the options. with the current implementation, the file system adapter is created and then replaced, which is just odd. i agree that we put a warning there - we should mention the "very good reasons" however, so that people can make an informed decision ;-)

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,493 @@
...
+        return $this;
+    }
+
+    /**
+     * Use this method if you want to use a different lock implementation than
+     * the one created in the constructor. Note that there are very good reasons
+     * that the local adapters are used by default and you do not inject them in
+     * the constructor but via setter injection instead. This is to protect you
+     * as a developer! Only use this method if you're really sure your lock
+     * implementation meets the needs of Symfony's HttpCache.
```
dbu: and same with the lock factory. mention that lock factory and cache storage must have the same scope e.g. on multi-server installations.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,493 @@
...
+            $body,
+            $cacheData['status'],
+            $cacheData['headers']
+        );
+    }
+
+    /**
+     * @param string $cacheDir
+     *
+     * @return \Symfony\Component\Lock\StoreInterface
```
dbu: please use `use` statements for phpdoc as well.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,493 @@
...
+        }
+    }
+
+    /**
+     * Increases a counter every time an item is stored to the cache and then
+     * prunes expired cache entries if a configurable threshold is reached.
+     * This only happens during write operations so cache retrieval is not
+     * slowed down.
+     */
+    private function pruneExpiredEntries()
```
dbu: good idea! i think its ok to have this active by default. but it should be possible to disable this by setting the threshold to -1, in which case we would not even get the cache item with the counter.

we should think of some way to trigger cache pruning out of band, like in a symfony command or such. that would be the better implementation if people really care about performance.

dbu:
and i will look at the build error. aparently varnish changed where they provide their downloads :-(

ddeboer:
Yeah, their packages are now hosted at packagecloud.io.

doc/symfony-cache-configuration.rst
Toflar: Updated.

src/SymfonyCache/TaggableStore.php
Toflar: To be honest, I don't like the options resolver. It's way too verbose. Real arguments are self documenting and better to read.

src/SymfonyCache/TaggableStore.php
Toflar: I didn't feel like creating an alias for this one time usage. `StoreInterface` is already taken.

src/SymfonyCache/TaggableStore.php
Toflar: I chose `0` instead of `-1` but it's now possible :-)

doc/symfony-cache-configuration.rst
dbu: okay, fair enough. i will go over this after merging and try to find a better formatting then, that better links names and default values

src/SymfonyCache/TaggableStore.php
dbu: i disagree. the problem is that now if i want to change the header, i MUST specify some threshold. and if we add more options, the problem increases. imho optionsresolver is a very clean way to specify what is allowed, and includes validation like the threshold must be > 0 or the header must be a string without whitespace.

src/SymfonyCache/TaggableStore.php
dbu: i would actually alias this to LockStoreInterface then. the point is that the `use` statements better list what things this class interacts with.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,494 @@
...
+        }
+
+        return new Response(
+            $body,
+            $cacheData['status'],
+            $cacheData['headers']
+        );
+    }
+
+    /**
```
dbu: can you add some text here? actually, i would like to change this method to getDefaultLockFactory and let it return a lock factory instead. with a phpdoc:

```
Build and return a default lock factory for when no explicit factory is specified.

The default factory uses the best quality lock store that is available on this system.
```

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,494 @@
...
+     * @var LockInterface[]
+     */
+    private $locks = [];
+
+    /**
+     * Constructor.
+     *
+     * @param string $cacheDir        The cache directory
+     * @param int    $pruneThreshold  The number of write operations until orphan data gets pruned (default: 500)
```
dbu: the magic value 0 needs to be mentioned here

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,494 @@
...
+     */
+    private $lockFactory;
+
+    /**
+     * @var LockInterface[]
+     */
+    private $locks = [];
+
+    /**
+     * Constructor.
```
dbu: please remove this line, it does not add any information ;-)

Toflar:
@dbu addressed all your comments and switched to options resolver. Tests still failing due to wrong package source which I don't like to fix in that PR because it does not belong here :-)

doc/symfony-cache-configuration.rst
```diff
@@ -230,6 +267,96 @@ options through the constructor:
...
 
+
+.. _taggablestore:
+
+The TaggableStore
+^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 2.1
+
+    The `TaggableStore` has been added in version 2.1.
```
dbu: in rst, you need to use two backticks for code formatting. a single backtick does not work (it only works in markdown). can you check that for all of the doc you are adding please?

doc/symfony-cache-configuration.rst
```diff
@@ -230,6 +267,96 @@ options through the constructor:
...
+
+.. warning::
+
+    You need at least versions 3.4 of `symfony/cache` and `symfony/lock`
+    to use this feature! Add the following lines to your `composer.json` and run
+    `composer update`::
+
+    "symfony/lock": "^3.4",
```
dbu: please indent this by another 4 spaces

doc/symfony-cache-configuration.rst
```diff
@@ -230,6 +267,96 @@ options through the constructor:
...
+  really sure your cache implementation meets the needs of Symfony's HttpCache.
+  Type: TagAwareAdapterInterface
+
+* **lock_factory**: The lock factory.
+  Use this option if you want to use a different lock implementation than the
+  default one.
+  Note that there are very good reasons that the local adapters are used by
+  default. This is to protect you as a developer! Only override it if you're
+  really sure your lock implementation meets the needs of Symfony's HttpCache.
+  Type: Factory
```
dbu: i think we need a fully qualified name with namespace here

doc/symfony-cache-configuration.rst
```diff
@@ -230,6 +267,96 @@ options through the constructor:
...
+* **purge_tags_header**: The HTTP header name used to check for tags
+  Type: string
+
+* **cache**: The cache adapter.
+  Use this option if you want to use a different cache implementation than the
+  default one.
+  Note that there are very good reasons that the local adapters are used by
+  default. This is to protect you as a developer! Only override it if you're
+  really sure your cache implementation meets the needs of Symfony's HttpCache.
+  Type: TagAwareAdapterInterface
```
dbu: lets use a FQN here

doc/symfony-cache-configuration.rst
```diff
@@ -230,6 +267,96 @@ options through the constructor:
         RewriteEngine On
         RewriteRule .* - [E=HTTP_AUTHORIZATION:%{HTTP:Authorization}]
 
+
+.. _taggablestore:
+
+The TaggableStore
```
dbu: i think this location makes the taggable store a sub section of user context, which is not correct. i would make it a title on the same level as those, using ~. and lets place it right after purge tags.

doc/symfony-cache-configuration.rst
```diff
@@ -230,6 +267,96 @@ options through the constructor:
...
+    public function __construct(
+        HttpKernelInterface $kernel,
+        SurrogateInterface $surrogate = null,
+        array $options = []
+    ) {
+        $store = new TaggableStore($kernel->getCacheDir());
+
+        parent::__construct($kernel, $store, $surrogate, $options);
+
+        $this->addSubscriber(new PurgeTagsListener());
```
dbu: i would put the code example with the purge tags section, and mention the options of the PurgeTagsListener there as well (purge http method and such)

but keep the information on options for the taggable store here.

src/SymfonyCache/TaggableStore.php
```diff
@@ -0,0 +1,500 @@
...
+
+    /**
+     * Release all locks.
+     *
+     * {@inheritdoc}
+     */
+    public function cleanup()
+    {
+        try {
+            /** @var LockInterface $lock */
```
dbu: the docblock `@var LockInterface[]` on the $locks property should be enough. at least for phpstorm it is...

dbu:
quick question to coordinate the FOSHttpCacheBundle release: do we need additional configuration for the tag purge header in the symfony cache proxy configuration? if so, i will wait with tagging a new version of the bundle until we wrapped this up and i tagged a library release.

dbu:
ok, fixed the varnish install in master. can you please rebase?

Toflar:
Rebased. Updated docs. Added Changelog entry.

> quick question to coordinate the FOSHttpCacheBundle release: do we need additional configuration for the tag purge header in the symfony cache proxy configuration?

Yeah would be a great addition. But I don't feel like this is a must. The default works out of the box and if you want to adjust the header name (for whatever reason you would like to do that) then you have to submit a PR adding support for that in the bundle 😄

dbu:
thanks a lot. this looks good now! i want to do some cleanups and will then merge this later this week and ideally tag a release.

Toflar:
Cool! I think we're not in a hurry here as Symfony 3.4 is still 1.5 months away.
I really hope this alternative `HttpCache` `Store` implementation will help out many people in the Symfony community 👍

javiereguiluz:
@Toflar for those of us who can't review this code or follow this long discussion, could you please summarize as concisely as possible the advantages of this feature over the existing Symfony Http Cache and also, if there's any, the drawbacks or shortcomings compared to Symfony? Thanks!

Toflar:
Hey @javiereguiluz, sure I can give it a try :)

As concisely as possible: It supports tag based cache invalidation and auto-pruning of local cache entries and builds on top of the latest Symfony components that were not available by the time the `Store` shipped with Symfony was developed.

And here are some more details, if you have time:

First off: If you are not familiar with the `FOS/HttpCache` library, you can ignore everything in the PR except for the `TaggableStore` (and its tests of course).

* The `TaggableStore` does not replace the existing Symfony `HttpCache` but is just a different implementation of the `StoreInterface` the `HttpCache` uses than the one shipped with Symfony.
* Instead of re-implementing locking and caching, it uses the well tested Symfony `Cache` and `Lock` components, both with the local filesystem adapters by default.
* Thanks to the `TagAwareAdapter` of the `Cache` component, it supports tag based cache invalidation.
* Thanks to the `PrunableInterface` of the `Cache` component, added in Symfony 3.4, it supports auto-pruning of the expired entries on the filesystem trying to prevent flooding the filesystem.
* For the sake of completeness: Thanks to the `Cache` and `Lock` components, it also allows to use a different cache or lock adapter than the local filesystem. However, as stated many times all over the different projects, developers should be careful about choosing the right adapter as some of them are definitely not well suited for the purpose of `HttpCache` and you're better of switching to a "real" reverse proxy as Varnish or LiteSpeed Cache. That's why local adapters are default and usually the best choice.

javiereguiluz:
@Toflar fantastic explanation! Thanks!!

dbu:
@javiereguiluz toflar tried a pull request to make the default store of symfony use the cache + lock components, but that got refused. for the tag capabilities, we need something and i like this solution a lot more as its built on existing and tested components. 

but if you have any kind of feedback on it, i would be glad to hear it, now we can still easily change things.

javiereguiluz:
@dbu I don't care about tags because my HTTP Cache is super basic ... but *"auto-pruning of the expired entries"* is a super-killer feature to me. Otherwise, I must clean up the http_cache/ dir from time to time because nothing is deleted and it grows to GB of data every day :(  We've received in the Symfony repository countless reports about this behavior.

Toflar:
Thinking about it, it would maybe make sense to have the `TaggableStore` in its own library as I feel @javiereguiluz would use it but does not need anything else the `FOSHttpCache` library provides. Am I right about that?
And yes, I still think it would make sense to replace the default `Store` in Symfony with it but to be honest I have invested many, many hours in the PR against Symfony, then in this PR and the whole concept as such so I'm kind of happy I'm close to finally using it in my projects. So another PR for Symfony is out of scope as of today but a separate library would maybe make sense and should not be too big of a deal.

Toflar:
Ping @javiereguiluz: any comment on having a standalone store implementation?

javiereguiluz:
I can't make technical comments because I haven't reviewed this in detail ... but what I need (and lots of others, according to the community feedback) is a Symfony HttpCache store which doesn't grow in size without limits and which doesn't store expired entries forever.

Ideally, your changes would be merged in Symfony itself so we can benefit from it without having to do or install anything. Thanks!

dbu:
at this point, i think a separate repository would be really cool! then the store should just be called Toflar/PsrCacheStore imho. the fact that it supports tags is free from the fact that it uses PSR-6. its important for FOSHttpCache but the main feature of the store is using PSR-6 caches.

@javiereguiluz toflar first tried a PR against symfony core but that got refused by fabien with [this comment](https://github.com/symfony/symfony/pull/20061#issuecomment-313339092). i can see the concern, but think its a bit over-cautious and seeing how TaggableStore has less code thanks to the lock and cache components i feel it would be progress. lets see if this will change sometimes.

@Toflar i did a bit of refactoring on the store. i will push as soon as i got all tests fixed. if you agree, i suggest you then copy them from my branch into a repo, along with the documentation on the store. your composer.json can even properly specify that it requires symfony/cache and symfony/lock. and then i can adjust the FOSHttpCache doc to tell people to install toflar/symfony-cache-store (or whatever you want to call it) and reference the correct namespace. the tag listener for HttpCache would stay in this repository i think.

would that make sense?

dbu:
continued in #376.

PR #376, opened 2017-09-26
Symfony taggable cache

### LINKED ISSUE #286: Symfony HTTP Cache / Tag Invalidation

dbu:
replace #373, fix #286
squashed all commits and did some tweaks to the documentation.

TODO
* [x] Once Symfony 3.4 is out: remove the `@beta` in composer.json

=> Bundle support: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/403

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,21 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 ------------------
 
 * Support Symfony 4.
+
+### Testing
+
+* Upgraded phpunit to 5.7 / 6. If you use anything from the
+  `FOS\HttpCache\Test` namespace you need to update your project to use
+  PHPUnit 6 (or 5.7, if you are using PHP 5.6).
+
+### Symfony HttpCache
+
+* Added a `PurgeTagsListener` for tag based invalidation with the Symfony
+  `HttpCache` reverse caching proxy. This requires the newly created
+  [Toflar Psr6Store](https://github.com/Toflar/psr6-symfony-http-cache-store)
+  built on PSR-6 cache and supporting pruning expired cache entries.
 * Using Request::isMethodCacheable rather than Request::isMethodSafe to
   correctly handle OPTIONS and TRACE requests.
-* Upgraded phpunit to 5.7 / 6. If you use anything from the FOS\HttpCache\Test
-  namespace you need to update your project to use phpunit 6 (or 5.7, if you
-  are using PHP 5.6).
 
 2.0.2
 -----
--- a/composer.json
+++ b/composer.json
@@ -34,9 +34,9 @@
         "monolog/monolog": "^1.0",
         "php-http/guzzle6-adapter": "^1.0.0",
         "php-http/mock-client": "^0.3.2",
+        "phpunit/phpunit": "^5.7 || ^6.0",
         "symfony/process": "^2.3 || ^3.0 || ^4.0",
-        "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0",
-        "phpunit/phpunit": "^5.7 || ^6.0"
+        "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0"
     },
     "suggest": {
         "friendsofsymfony/http-cache-bundle": "For integration with the Symfony framework",
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -30,6 +30,8 @@ Using the trait
 Your ``AppCache`` needs to implement ``CacheInvalidation`` and use the
 trait ``FOS\HttpCache\SymfonyCache\EventDispatchingHttpCache``::
 
+    // app/AppCache.php
+
     use FOS\HttpCache\SymfonyCache\CacheInvalidation;
     use FOS\HttpCache\SymfonyCache\EventDispatchingHttpCache;
     use Symfony\Component\HttpFoundation\Request;
@@ -148,16 +150,18 @@ Refresh
 
 To support :ref:`cache refresh <cache refresh>`, register the
 ``RefreshListener``. You can pass the constructor an option to specify
-what clients are allowed to refresh cache entries. Refreshing is only allowed
-from the same machine by default. To refresh from other hosts, provide the
-IPs of the machines allowed to refresh, or provide a RequestMatcher that
-checks for an Authorization header or similar. *Only set one of
-``client_ips`` or ``client_matcher``*.
+what clients are allowed to refresh cache entries.
 
 The refresh listener needs to access the ``HttpCache::fetch`` method which
 is protected on the base HttpCache class. The ``EventDispatchingHttpCache``
 exposes the method as public, but if you implement your own kernel, you need
 to overwrite the method to make it public.
+
+Refreshing is only allowed from the same machine by default. To refresh from
+other hosts, provide the IPs of the machines allowed to refresh, or provide a
+RequestMatcher that checks for an Authorization header or similar. *Only set
+one of ``client_ips`` or ``client_matcher``*.
+
 
 * **client_ips**: String with IP or array of IPs that are allowed to
   refresh the cache.
@@ -168,6 +172,81 @@ to overwrite the method to make it public.
   allowed to refresh.
 
   **default**: ``null``
+
+Tagging
+~~~~~~~
+
+.. versionadded:: 2.1
+
+    Support for tag invalidation with Symfony HttpCache has been added in
+    version 2.1.
+
+To support :doc:`cache tags <response-tagging>`, require the additional package
+``toflar/psr6-symfony-http-cache-store:^1.0`` with composer and register the
+``PurgeTagsListener`` in your cache kernel. The purge listener needs your cache
+to use the special ``Toflar\Psr6HttpCacheStore\Psr6Store`` store, as the default
+store does not have tagging support.
+
+.. note::
+
+    Symfony's ``HttpCache`` store implementation does not support tags.
+    Therefore, you need the `Toflar Psr6Store`_ which implements the Symfony
+    Store interface but supports cache tagging. See the project README for more
+    information on the store.
+
+    To install the store, run
+    ``composer require toflar/psr6-symfony-http-cache-store``.
+
+Purging tags is only allowed from the same machine by default. To change this,
+you have the same configuration options as with the ``PurgeListener``. *Only
+set one of ``client_ips`` or ``client_matcher``*. Additionally, you can
+configure the HTTP method and header used for tag purging:
+
+* **client_ips**: String with IP or array of IPs that are allowed to
+  purge the cache.
+
+  **default**: ``127.0.0.1``
+
+* **client_matcher**: RequestMatcherInterface that only matches requests that are
+  allowed to purge.
+
+  **default**: ``null``
+
+* **tags_method**: HTTP Method used with purge tags requests.
+
+  **default**: ``PURGETAGS``
+
+* **tags_header**: HTTP Header that contains the comma-separated tags to purge.
+
+  **default**: ``X-Cache-Tags``
+
+To get cache tagging support, register the ``PurgeTagsListener`` and use the
+``Psr6Store`` in your ``AppCache``::
+
+    // app/AppCache.php
+
+    use Toflar\Psr6HttpCacheStore\Psr6Store;
+    use FOS\HttpCache\SymfonyCache\PurgeTagsListener;
+
+    // ...
+
+    /**
+     * Overwrite constructor to register the Psr6Store and PurgeTagsListener.
+     */
+    public function __construct(
+        HttpKernelInterface $kernel,
+        SurrogateInterface $surrogate = null,
+        array $options = []
+    ) {
+        $store = new Psr6Store([
+            'cache_directory' => $kernel->getCacheDir(),
+            'cache_tags_header' => 'X-Cache-Tags',
+        ]);
+
+        parent::__construct($kernel, $store, $surrogate, $options);
+
+        $this->addSubscriber(new PurgeTagsListener());
+    }
 
 .. _symfony-cache user context:
 
@@ -268,3 +347,4 @@ and at the HTML body of the response.
 
 .. _HttpCache: http://symfony.com/doc/current/book/http_cache.html#symfony-reverse-proxy
 .. _HttpKernel: http://symfony.com/doc/current/components/http_kernel.html
+.. _Toflar Psr6Store: https://github.com/Toflar/psr6-symfony-http-cache-store
--- a/src/ProxyClient/HttpProxyClient.php
+++ b/src/ProxyClient/HttpProxyClient.php
@@ -102,6 +102,12 @@ abstract class HttpProxyClient implements ProxyClient
      * Reusable function for proxy clients.
      * Escapes `,` and `\n` (newline) characters.
      *
+     * Note: This is not a safe escaping function, it can lead to collisions,
+     * e.g. between "foo,bar" and "foo_bar". But from the nature of the data,
+     * such collisions are unlikely, and from the function of cache tagging,
+     * collisions would in the worst case lead to unintended invalidations,
+     * which is not a bug.
+     *
      * @param array $tags The tags to escape
      *
      * @return array Sane tags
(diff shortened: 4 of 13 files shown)


Toflar:
Are you sure I should not extract an extra library? Maybe we should wait for Javier's response there?

dbu:
@Toflar okay, pushed my last changes. mainly i realized that the cache directory parameter only makes sense when not passing the cache and lock instance to the store, so i moved it into the options as well. i also refactored the OptionsResolver to use closures for the default cache and lock implementation to avoid instantiating them when not needed. this looks ready to me.

do you want to take the store and its test into its own repo and create a readme from the doc section on the store?

Toflar:
Perfect, thank you! Yes, I'll take care extracting it into its own library and then I'll update my original PR against FOSHttpCache and see if I can build some kind of bridge. Would that be fine with you?

dbu:
great! i did some tweaks in this branch. can you do the update from this branch? or i can just edit it myself, its mostly deleting the things that moved to your repo anyways :-)

i would keep the store with the tag handling support - its basically free with PSR-6. then there should be no need to bridge things, we can just tell in the doc to require your repo when you want cache tagging in symfony HttpCache, and everything should just work the same as it works now in this pull request (apart from adjusting the namespace of the store)

Toflar:
Yeah I'll let you know when it's ready :)

Toflar:
Here we go: https://github.com/Toflar/psr6-symfony-http-cache-store 😄 

Registered over at packagist already too, so you could update the PR requiring `dev-master` (I guess I'll leave it like that until Symfony 3.4.0 stable is published, wdyt?).

dbu:
thank you! i update the PR, but tests fail because the Psr6Store is final. this feels stupid, seems to be something with how we do the mocking of the store in the unit test.

lets gather some feedback on the store and then release it asap. to merge this, i'd want at least an 1.0.0-alpha1 - i think thats reasonable to do soon.

doc/symfony-cache-configuration.rst
```diff
@@ -169,6 +173,79 @@ to overwrite the method to make it public.
...
+     * Overwrite constructor to register the Psr6Store and PurgeTagsListener.
+     */
+    public function __construct(
+        HttpKernelInterface $kernel,
+        SurrogateInterface $surrogate = null,
+        array $options = []
+    ) {
+        $store = new Psr6Store([
+            'cache_directory' => $kernel->getCacheDir(),
+            'cache_tags_header' => 'X-Cache-Tags',
```
Toflar: You noticed. Well done!

Toflar:
> thank you! i update the PR, but tests fail because the Psr6Store is final.

Yeah I think it should not be extended but I can remove it. I just don't want to provide BC for child classes...

dbu:
having it final seems correct to me. but how to test? i wonder if it could make sense to define an interface for this that extends the symfony StoreInterface with the prune and invalidateTags methods. that would define clearly what is added, and we could mock the interface...

Toflar:
Makes sense to me, I also mocked the `Lock` which is `final` too and I could mock it thanks to the `LockInterface` so to me that seems like good practice. Can you work with https://github.com/Toflar/psr6-symfony-http-cache-store/commit/ebf22b737c8cf84743a607fa787985b52958f765?

dbu:
that looks great! i added some comments about wording and naming, but the structure is exactly what i had in mind.

composer.json
```diff
@@ -36,7 +36,10 @@
         "php-http/mock-client": "^0.3.2",
         "symfony/process": "^2.3||^3.0",
         "symfony/http-kernel": "^2.3||^3.0",
-        "phpunit/phpunit": "^5.7 || ^6.0"
+        "phpunit/phpunit": "^5.7 || ^6.0",
+        "toflar/psr6-symfony-http-cache-store": "dev-master",
+        "symfony/lock": "^3.4@dev",
+        "symfony/cache": "^3.4@dev"
```
dbu: we need to adjust this once the psr6-symfony-http-cache-store and symfony 3.4 are released.

CHANGELOG.md
```diff
@@ -6,9 +6,13 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 2.1.x (unreleased)
 ------------------
 
-* Upgraded phpunit to 5.7 / 6. If you use anything from the FOS\HttpCache\Test
-  namespace you need to update your project to use phpunit 6 (or 5.7, if you
-  are using PHP 5.6).
+* [Testing] Upgraded phpunit to 5.7 / 6. If you use anything from the
+  `FOS\HttpCache\Test` namespace you need to update your project to use
+  PHPUnit 6 (or 5.7, if you are using PHP 5.6).
+* [Symfony HttpCache] Added a `TaggableStore` for tag based invalidation with
```
Toflar: This needs to be updated.

doc/symfony-cache-configuration.rst
```diff
@@ -169,6 +173,79 @@ to overwrite the method to make it public.
...
+~~~~~~~
+
+.. versionadded:: 2.1
+
+    Support for tag invalidation with Symfony HttpCache has been added in
+    version 2.1.
+
+To support :doc:`cache tags <response-tagging>`, register the
+``PurgeTagsListener``. The purge listener needs the special
+``Toflar\Psr6HttpCacheStore\Psr6Store`` cache store with tag support.
```
Toflar: `tagging support`

doc/symfony-cache-configuration.rst
```diff
@@ -169,6 +173,79 @@ to overwrite the method to make it public.
...
+
+.. note::
+
+    Symfony's ``HttpCache`` store implementation does not support tags.
+    Therefore, you need the `Toflar Psr6Store`_ which implements the Symfony
+    Store interface but supports cache tagging. See the project README for more
+    information on the store.
+
+    To install the store, run
+    ``composer require toflar/psr6-symfony-http-cache-store dev-master``.
```
Toflar: I'd remove `dev-master` here. Once we release it it's obsolete and composer will choose the right version anyway :)

CHANGELOG.md
dbu: i deleted the wrong half of this :-( thanks for spotting, fixed

doc/symfony-cache-configuration.rst
dbu: indeed. i thought we might merge this before the store is tagged, but as we have to wait for stable symfony 3.4 anyways we can wait.

src/ProxyClient/Symfony.php
```diff
@@ -52,10 +54,37 @@ public function refresh($url, array $headers = [])
...
-            'purge_method' => PurgeListener::DEFAULT_PURGE_METHOD,
-        ]);
+        $resolver->setDefault('purge_method', PurgeListener::DEFAULT_PURGE_METHOD);
+        $resolver->setAllowedTypes('purge_method', 'string');
+        $resolver->setDefault('purge_tags_method', PurgeTagsListener::DEFAULT_PURGE_TAGS_METHOD);
+        $resolver->setAllowedTypes('purge_tags_method', 'string');
+        $resolver->setDefault('purge_tags_header', PurgeTagsListener::DEFAULT_PURGE_TAGS_HEADER);
+        $resolver->setAllowedTypes('purge_tags_header', 'string');
+        $resolver->setDefault('purge_tags_header_length', 7500);
+        $resolver->setAllowedTypes('purge_tags_header_length', 'int');
```
andrerom: we should perhaps align naming with Varnish one or the other way?

andrerom: Also code wise easier to read if `setDefaults()` is used imo: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/332/files#diff-006816ea733cc0564d6a5b345ed47929R160

andrerom:
with 3.4 now in beta2 I guess this can be updated to `@beta` flag and moved forward to merge.

Is there plans for functional tests here?

src/ProxyClient/Symfony.php
dbu: agreed with setDefaults. for the naming i agree as long as its really the same semantics. we can't change existing names easily, because backwards compatibility is important.

dbu:
i want to wait with merging until 3.4 is tagged as stable release.

functional tests would indeed be good. i don't have the time to build them - if toflar finds the time to do them would be awesome!

Toflar:
No plan for functional tests. Need to get too many things done before the stable release of 3.4 and we're already at beta3 now :-D

composer.json
```diff
@@ -36,7 +36,10 @@
         "php-http/mock-client": "^0.3.2",
         "symfony/process": "^2.3||^3.0",
         "symfony/http-kernel": "^2.3||^3.0",
-        "phpunit/phpunit": "^5.7 || ^6.0"
+        "phpunit/phpunit": "^5.7 || ^6.0",
+        "toflar/psr6-symfony-http-cache-store": "dev-master",
+        "symfony/lock": "^3.4@beta",
+        "symfony/cache": "^3.4@beta"
```
dbu: this needs to be ^3.4 and ^1.0 , then we can merge. lets hope symfony 3.4 comes out on time ;-)

andrerom:
Todo: Update bundle in some cases once this gets merged: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/DependencyInjection/FOSHttpCacheExtension.php#L403-L410

composer.json
NavyCoat: There is it: https://symfony.com/blog/symfony-3-4-0-released

Toflar: I'll be releasing `toflar/psr6-symfony-http-cache-store` very soon. I even set a reminder in my calendar when November 30th was announced as release date 😄

Toflar:
Ping @dbu: Tagged version 1.0.0 of https://packagist.org/packages/toflar/psr6-symfony-http-cache-store. Now also tested with `composer update --prefer-lowest`, fancy new Symfony 4.0 and PHP 7.2! 🎉

src/ProxyClient/HttpProxyClient.php
```diff
@@ -114,4 +114,47 @@ protected function escapeTags(array $tags)
...
+     * invalidation requests where you might need multiple requests if you
+     * want to invalidate too many cache tags (so the header would get too long).
+     *
+     * @param string $value
+     * @param int    $length
+     * @param string $delimiter
+     *
+     * @return array
+     */
+    protected function splitLongHeaderValue($value, $length = 7500, $delimiter = ',')
```
dbu: finally looking over this again and i notice that this method is quite different from what we do to solve the same problem in Varnish::invalidateTags. can we instead put the logic from Varnish::invalidateTags into a HttpProxyClient::generateTagInvalidationChunks and use that for both?

this method is inefficient because the caller first implodes all tags, then splits them again and builds chunks from it

dbu:
hm, the toflar store requires symfony kernel 3.0 or newer. is it time to drop symfony 2 support here? or would it be easy to allow symfony 2 in the store?

Toflar:
> @Toflar would you have time to address the unifcation of varnish vs symfony proxy client method to invalidate tags avoiding too long http headers?

I don't think they can be unified. The Varnish proxy requires a very special regex-style invalidation header which makes absolutely no sense for the SF proxy.

> hm, the toflar store requires symfony kernel 3.0 or newer. is it time to drop symfony 2 support here? or would it be easy to allow symfony 2 in the store?

It depends on the `Lock` component, so SF 3.4+ only.

Toflar:
Not sure but we wanted to add it as a `suggest`, didn't we? I mean, if you don't use the SF `HttpCache` but Varnish instead, why would you need to install it?

andrerom:
> It depends on the Lock component, so SF 3.4+ only.

but technically lock can be installed independently right? same as with cache afaik

andrerom:
>> @Toflar would you have time to address the unifcation of varnish vs symfony proxy client method to invalidate tags avoiding too long http headers?
> I don't think they can be unified. The Varnish proxy requires a very special regex-style invalidation header which makes absolutely no sense for the SF proxy.

@dbu Maybe we can try to do that as followup once/if xkey gets in?

Toflar:
> but technically lock can be installed independently right? same as with cache afaik

the question was about the store and this is based on top of both of these components so they are a requirement 😊 but the whole store lib is optional for foscache 😊

composer.json
```diff
@@ -34,9 +34,12 @@
         "monolog/monolog": "^1.0",
         "php-http/guzzle6-adapter": "^1.0.0",
         "php-http/mock-client": "^0.3.2",
+        "phpunit/phpunit": "^5.7 || ^6.0",
         "symfony/process": "^2.3 || ^3.0 || ^4.0",
         "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0",
-        "phpunit/phpunit": "^5.7 || ^6.0"
+        "symfony/lock": "^3.4 || ^4.0",
+        "symfony/cache": "^3.4 || ^4.0",
```
dbu: @Toflar do we need the lock and cache as direct dependency? guess not, we only had that for stability weakening? the store brings them and its the store that has the dependency, not directly us?

Toflar: They're not needed. They're `dev` requirements anyway and I guess it's a leftover from the times where the store implementation was part of this PR and not a separate lib yet :)

andrerom: Was also probably needed during dev/beta of lock :)

dbu:
@Toflar i refactored the tag invalidation header splitting. can you check if you have any feedback? i think i am now happy with this.

Toflar:
Nice work, I've added a comment about the delimiter but otherwise: 👍

src/ProxyClient/HttpProxyClient.php
```diff
@@ -109,54 +109,33 @@ protected function queueRequest($method, $url, array $headers, $validateHost = t
...
      *
-     * @return array
+     * @return int Number of tags per tag invalidation request
      */
-    protected function splitLongHeaderValue($value, $length = 7500, $delimiter = ',')
+    protected function determineTagsPerHeader($escapedTags)
     {
-        if (mb_strlen($value) <= $length) {
-            return [$value];
+        if (mb_strlen(implode(',', $escapedTags)) < $this->options['header_length']) {
```
Toflar: Either I'm blind or my comment is gone? Hope I do not repeat myself here now but just in case: I mentioned that `,` is hardcoded now but I might want to use `foobar` as a delimiter later which affects the length of the header so it should be the second argument of this method :)

dbu: you commented on the commit, instead of in the pull request ;-)

thanks for the input, i guess i should indeed pass in the delimiter, it costs us nothing and is future proof.

andrerom: both xkey and Surrogate-Key (Fastly) works with space separated tags afaik, so a bit outside of this but might be worth to either abstract as stated in this thread or even just align.

dbu:
@andrerom i added whitespace to the things to replace in tags. do you want to refactor this PR to use spaces instead of comma? i don't have time to do it and want to merge it this week...

andrerom:
I'm unsure if I have time for my PR's, we are doing two releases in parallel for end of year, one is major.

dbu:
> I'm unsure if I have time for my own PR's, so would not bet on that, we are doing two releases in parallel for end of year, one is major.

sounds like a busy week indeed. i think its not that important that we are consistent with the separator between different caches, as each cache has its own client anyways.

i will revert the whitespace escaping for now. thanks to the failing tests i realized that this would be a potential BC break if you don't wipe the cache after deploying this version - invalidations would then use the new escaping and not hit old tags.

dbu:
@Toflar want to do a last review of this before i merge?

dbu:
yay, done! thanks a ton!

i will only tag the library next week, hopefully getting the other big contribution with varnish xkey support in. if you spot something until monday, we can just get it into the head. otherwise we can still to a .1 patch release with a fix. its a new feature, so no drama if that happens.

pbowyer:
Thanks everyone!

Issue #377, opened 2017-09-29
UserContextListener set http method wrong


blankse:
I try to integrate the FOSCommentBundle in a eZ Platform installation.
In dev environment it works. But in prod environment (without varnish) the UserContextListener makes me problems.

It set with the $request->isMethodSafe() function the method of the request.
(https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/SymfonyCache/UserContextListener.php#L107)

But the FOSCommentBundle used PATCH and PUT methods and at this moment the self::$httpMethodParameterOverride variable is not set to true.
(https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpFoundation/Request.php#L1206)

This happens later in the FrameworkBundle.
(https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/FrameworkBundle.php#L65)

How can I handle this? Is this a bug of this repository?

dbu:
hi, the symfony cache listeners are executed by the symfony HttpCache. at that point, the symfony kernel and the configuration are not yet loaded, so symfony configuration can not be applied.

what is the problem you see? the listener checks if the request is safe (that is, if its cacheable at all - only GET and HEAD requests are cacheable). PATCH and PUT are not safe, so nothing should happen. or does the FOSCommentBundle send GET requests with some magic request attribute to tell they are to be treated as PATCH / PUT?

blankse:
The FOSCommentBundle send a POST request with an _method param set to PUT.

The listener set the request method to POST with the isMethodSafe() call.

Later in the RouterListener (the second method call) the method is still POST and no route is found.

Is it possible to set the method back to null after the isMethodSafe() call? Or clone the request before?

dbu:
ah, web browsers and the need to do horrible things :-/

i was totally not aware of this side effect. yeah i guess we should somehow reset this. is resetting the method to `null` always safe? cloning sounds like it could have bad side effects with large requests...

dbu:
or would it be better to not call that method and just check the "real" request method without triggering side effects?

blankse:
> or would it be better to not call that method and just check the "real" request method without triggering side effects?

I think this would be better.

The method attribute is protected and i think the setMethod function is not a good idea.

PR #378, opened 2017-10-09
Fix: UserContextListener set http method wrong

blankse:
Fix: https://github.com/FriendsOfSymfony/FOSHttpCache/issues/377

### DIFF
--- a/src/SymfonyCache/UserContextListener.php
+++ b/src/SymfonyCache/UserContextListener.php
@@ -104,7 +104,9 @@ class UserContextListener implements EventSubscriberInterface
                 return;
             }
 
-            if ($request->isMethodSafe() && $hash = $this->getUserHash($event->getKernel(), $request)) {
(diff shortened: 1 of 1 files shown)


blankse:
I add the comment.

dbu:
thanks a lot, merged. i will wait with the release until symfony 3.4 is released, so that we can also merge #376. a patch release is not really possible because the build process with varnish installation had to be changed, and those commits are after BC breaking changes... i hope thats not too much hassle for you? you can use `2.1@dev` until then if thats doable?

blankse:
eZ Platform still using FOSHttpCache 1.4... Can I also create a PR for the 1.4 branch?

dbu:
ups. yeah sure. we should have done it the other way round but the change is so small that there should be no merge conflicts when i later merge 1.4 upstream again. can you do the pull request against the 1.4 branch? there i can also immediately release a patch version with this.

blankse:
Yes, the other way would be better, but I see it too late.

I created the PR against 1.4. Thank you!

PR #379, opened 2017-10-09
Fix: UserContextListener set http method wrong (1.4)

blankse:
Backport of https://github.com/FriendsOfSymfony/FOSHttpCache/pull/378

### DIFF
--- a/src/SymfonyCache/UserContextSubscriber.php
+++ b/src/SymfonyCache/UserContextSubscriber.php
@@ -104,7 +104,9 @@ class UserContextSubscriber implements EventSubscriberInterface
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


dbu:
i tagged 1.4.4 with this (plus the fixes to the build process to get a green build on 1.4 again). thanks again for the help!

btw, if you want, you could add a unit test that tests this scenario and now works but would fail without your fix. there should exist a unit test for the UserContextListener. (if you do, only do it in master and don't bother with the 1.4 branch)

blankse:
Thanks again!
Here is the unit test: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/383

PR #390, opened 2017-11-21
Added package name

hpatoio:
Thought was a good idea to add the package name in the doc. Otherwise people need to go to packagist to find package name.


dbu:
thanks, this makes sense!

Issue #393, opened 2017-11-30
Varnish Cpu efficiency with CacheTags 


olivierHa:
Hello, 

I've noticed that this library is appending "(,.+)?$" when submitting BAN to Varnish : 

https://github.com/FriendsOfSymfony/FOSHttpCache/blob/ecc458282b8a6f75387c3cb4705615b0f5b6d6a8/src/ProxyClient/Varnish.php#L73

But on Varnish side the BAN is called using a regex : 

https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/resources/config/varnish/fos_ban.vcl#L22

Is there any reasons to add such a suffix ?

Regards

Olivier

dbu:
yes, the trailing comma or end of string is required to avoid accidental invalidation on partial matches. if there are the tags foo and foobar, invalidating with only the string "foo" would also invalidate the tag "foobar".

but looking at that expression, i think that we should instead match on `'(%s)(,|$)'`. then we don't need to match the whole tag header, in case those are long. do you want to try a pull request for that?

there is also work on supporting the varnish xkey module, which would be a much more efficient way to handle tags, as it would be purges instead of bans. see #332 for that topic (it needs some effort in figuring out how to test it, if you want to help make it happen)

olivierHa:
Thanks for this quick reply. I understand better the end of the x-cache-tag expression :+1: 

I will try to submit a PR

PR #394, opened 2017-11-30
Better regexp for Varnish ban

olivierHa:
As explained in #393 , regexp used for banning aren't efficient.

### DIFF
--- a/tests/Unit/ProxyClient/VarnishTest.php
+++ b/tests/Unit/ProxyClient/VarnishTest.php
@@ -109,7 +109,7 @@ class VarnishTest extends TestCase
             \Mockery::on(
                 function (RequestInterface $request) {
                     $this->assertEquals('BAN', $request->getMethod());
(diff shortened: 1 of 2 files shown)


tests/Unit/ProxyClient/VarnishTest.php
```diff
@@ -109,7 +109,7 @@ public function testTagsHeaders()
             \Mockery::on(
                 function (RequestInterface $request) {
                     $this->assertEquals('BAN', $request->getMethod());
-                    $this->assertEquals('(mytag|othertag)(,.+)?$', $request->getHeaderLine('X-Cache-Tags'));
+                    $this->assertEquals('(mytag|othertag)(,|$))?$', $request->getHeaderLine('X-Cache-Tags'));
```
dbu: there must be no more question mark. it must either be followed by , or |. the code is right but i think the assumption is wrong.

dbu:
@ddeboer would you mind reviewing this as well? i think its correct, but maybe i am missing something? messing it up would be nasty...

ddeboer:
Looks good to me!

PR #396, opened 2017-12-11
[Docker] Simplified testing container

andrerom:
To be able to run tests locally, to hopefully manage to unblock myself of #332 

Note that this is a somwhat simpler approach then what was done in #344. For instance this on purpose does not intend to be pushed to docker hub, you build locally and you test

Todo:
- [ ] Update Docker on Travis to get this to run _(effort: low)_
- [x] ~coverage not covered here yet _(effort: medium)_~ _(followup)_
- [ ] Not tested doc generation yet _(effort: low)_


.travis.yml
```diff
@@ -1,83 +1,52 @@
...
+  - docker-compose exec fos sh -c "composer validate --strict --no-check-lock"
+  - docker-compose exec fos sh -c "vendor/bin/phpunit $PHPUNIT_FLAGS"
+  #- if [[ "$DOCCHECK" = true ]]; then make -C doc SPHINXOPTS='-nW' html; fi
+  #- if [[ "$DOCCHECK" = true ]]; then make -C doc spelling; fi
 
-after_script:
-  - wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover
+#after_script:
+  # Todo fix to use curl in image, and either install xdebug OR change to use phpdbg, then re add $PHPUNIT_FLAGS usage
+  #- wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover
```
dbu: i can live with dropping scrutinizer from our test suite. maybe we could introduce phpstan instead (in a separate pull request, though)

.travis.yml
```diff
@@ -1,83 +1,52 @@
...
 
 script:
-  - composer validate --strict --no-check-lock
-  - vendor/bin/phpunit $PHPUNIT_FLAGS
-  - if [[ "$DOCCHECK" = true ]]; then make -C doc SPHINXOPTS='-nW' html; fi
-  - if [[ "$DOCCHECK" = true ]]; then make -C doc spelling; fi
+  - docker-compose exec fos sh -c "composer validate --strict --no-check-lock"
+  - docker-compose exec fos sh -c "vendor/bin/phpunit $PHPUNIT_FLAGS"
+  #- if [[ "$DOCCHECK" = true ]]; then make -C doc SPHINXOPTS='-nW' html; fi
+  #- if [[ "$DOCCHECK" = true ]]; then make -C doc spelling; fi
```
dbu: if you can run the sphinx install in the docker container, it would also be useful for locally building the doc.

tests/docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
...
+    NGINX_VERSION=1.9.15 \
+    NGINX_CACHE_PURGE_VERSION=2.3 \
+    VARNISH_MALLOC_SIZE="256M" \
+    COMPOSER_ALLOW_SUPERUSER=1 \
+    DEBIAN_FRONTEND=noninteractive
+
+# Use official packages from Varnish and build with varnish-modules mainly for xkey
+# see: https://github.com/varnish/varnish-modules/tree/master/docs
+RUN set -xe \
+        && buildDeps=" \
```
dbu: is it best practice to chain commands together like this? would it not be better to repeat `RUN`? i am worried how this is for debugging, should one of the setup commands fail.

ddeboer: Each `RUN` step creates a new Docker cache layer, so (e.g.) when building something from source and then discarding the source and any `-dev` package dependencies, you want to do that in a single `RUN` step.

dbu: thanks for the explanation.

andrerom: Basically if this is not done, the image will take longer to build and end up taking lot more space as earlier layers will still contain the files your delete in later layers. It's strange they haven't found a better way to deal with this, like for instance grouping blocks of commands as one layer, but thats a bit outside scope here :)

.travis.yml
andrerom: Either way we need to make sure it can be done inside the image. If this can be done in followup then I can try to wrap this up soonish, like end of week.

.travis.yml
andrerom: Yes, I have afaik all packages installed so once I find some time I'll test if it works and re add it against container.

tests/docker/Dockerfile
```diff
@@ -0,0 +1,76 @@
+# PHP Version number as set by offical php docker image tags: https://hub.docker.com/r/library/php/tags/
+# Apache is setup as http server, varnish + nginx is installed for use as reverse proxies
+ARG PHP_VERSION=7.1
+FROM php:${PHP_VERSION}-apache
```
ddeboer: This requires Docker >= 17.05.0, so you need to install that (Travis now runs 17.03).

andrerom: Travis is running on 17.09 since the 12th ;)

dbu: nice when problems solve themselves :-)

PR #402, opened 2018-01-12
Create AnonymousRequestMatcher with isAnonymous Logic to share it with FOSHttpCacheBundle

blankse:
Needed for https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/425

So it is possible to use the same isAnonymous Logic in UserContextSubscriber of FOSHttpCache and FOSHttpCacheBundle

### DIFF
--- a/src/SymfonyCache/UserContextSubscriber.php
+++ b/src/SymfonyCache/UserContextSubscriber.php
@@ -11,6 +11,7 @@
 
 namespace FOS\HttpCache\SymfonyCache;
 
+use FOS\HttpCache\UserContext\AnonymousRequestMatcher;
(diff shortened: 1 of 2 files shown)


dbu:
thanks, this makes a lot of sense! (and sorry for the slow response)

i should tag a version so that the bundle can require that version as minimum, but unfortunately the build is failing for php 7.1 and 7.2

dbu:
hm, actually, why did i merge this into the 1.4 branch? this is a "new feature" type of change and should go to version 2.2 i think. i added some discussion on https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/425 as well - would you mind much if i revert this merge and you add the change in the master branch?

Issue #142, opened 2014-12-01
Akamai support?


andrerom:
Just to start the discussion, anyone looked into feasibility of Akamai support?

Might also be interesting to support other CDN's as well, but Akamai is typically preferred choice among media companies and they often use it already.

andrerom:
Small update here, @kmadejski did a small prototype for eZ on this and there are some good news and some bad news:
- Tagging is now supported, so it's possible to purge by that
- We didn't see anyway so far to accomplish x-user-context-hash kind of feature, so it would be be limited to traffic to anonymous users it seems


For me those two features are kind of key for FOSHttpCache, so maybe that rules out adding support for it here unless someone sees a way.

dbu:
i am okay with having akamai support without the context hash in it. FOSHttpCache is also about cache tagging and invalidation of paths / banning. as long as documentation is clear what can be used and how (example config for akamai etc) i think its cool to have it.

kmadejski:
@dbu thanks for the reply. 
I think I can easily share what I did, but my current schedule is extremely busy. I will try to look into that in a couple of weeks.

dbu:
@kmadejski would be awesome if you can share / show it somewhere. a pull request with documentation would be very cool of course, but a blog post just explaining what you did and showing the files you did without integrating them into the repository here could be a start as well. then someone else who needs akamai support could maybe contribute the change, if you license your code accordingly.

PR #403, opened 2018-01-19
[WIP][RFC] Fastly support

hpatoio:
Adding support for Fastly 
Goes with https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/427

For details about Fastly API:

 - https://docs.fastly.com/api/purge#soft_purge_2e4d29085640127739f8467f27a5b549
 - https://docs.fastly.com/api/purge#purge_d8b8e8be84c350dd92492453a3df3230 

Related discussion: #142

### DIFF
--- a/src/ProxyClient/Fastly.php
+++ b/src/ProxyClient/Fastly.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of the FOSHttpCache package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\Exception\InvalidArgumentException;
+use FOS\HttpCache\ProxyClient\Invalidation\BanCapable;
+use FOS\HttpCache\ProxyClient\Invalidation\PurgeCapable;
+use FOS\HttpCache\ProxyClient\Invalidation\RefreshCapable;
+use FOS\HttpCache\ProxyClient\Invalidation\TagCapable;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Fastly HTTP cache invalidator.
+ *
+ * @author Simone Fumagalli <simone.fumagalli@musement.com>
+ */
+class Fastly extends HttpProxyClient implements TagCapable
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function invalidateTags(array $tags)
+    {
+
+        $headers= [
+            'Fastly-Key' => $this->options['authentication_token'],
+            'Accept' => 'application/json'
+        ];
+
+        if (true === $this->options['soft_purge']) {
+            $headers['Fastly-Soft-Purge'] = 1;
+        }
+
+        foreach ($tags as $tag) {
+            $this->queueRequest(Request::METHOD_POST, sprintf("/service/%s/purge/%s", $this->options['service_identifier'], $tag), $headers, false);
+        }
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureOptions()
+    {
+        $resolver = parent::configureOptions();
+
+        $resolver->setRequired([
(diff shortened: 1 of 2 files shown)


dbu:
thanks for starting this! i am super busy before php benelux, will try to have a look at this next week.

do you have specific questions at this point? what remains to be done (apart from test and style issues)?

hpatoio:
> do you have specific questions at this point? 

No, not really

> what remains to be done (apart from test and style issues)?

The FastlyProxyClient need implements:

- [x] TagCapable
- [ ] PurgeCapable
- [ ] RefreshCapable

And then documentation.

dbu:
cool. then i will ignore that pull request for now, please ping me when you need input or want me to review something. i would love to have fastly support in this library!

src/ProxyClient/Fastly.php
```diff
@@ -0,0 +1,78 @@
...
+            'Fastly-Key' => $this->options['authentication_token'],
+            'Accept' => 'application/json'
+        ];
+
+        if (true === $this->options['soft_purge']) {
+            $headers['Fastly-Soft-Purge'] = 1;
+        }
+
+        foreach (\explode(",", $this->options['service_identifier']) as $fastlyServiceId) {
+            foreach ($tags as $tag) {
```
andrerom: In the case of hard purge, it's possible to invalidate 256 tags at a time now:
https://docs.fastly.com/api/purge#purge_db35b293f8a724717fcf25628d713583

andrerom:
I think we might have another todo here. From what I could see last week, UserContextInvalidator also needs to be changed to not rely on BanCable for this to be possible.

One alternative would be to add tags for user id and session id on the hash request cache. Any thoughts on this @dbu ?

dbu:
so it would be possible to do user context caching with fastly, if we find a way around the ban issue? that would be quite cool indeed.

how do we do that with the symfony cache kernel thingy? that one does not have full banning implemented, but does have tags. do we not invalidate the hash lookup for that case?

dbu:
tagging the hash lookup response with the session id would be a very good idea imho. with the x-cache extension, invalidating the tag would also be much better for varnish than the ban request that we currently do. (apart from that ban request being completely broken and banning the whole server each time)

Toflar:
The Symfony Proxy is not `BanCapable` but I'm not sure what the question here is?

andrerom:
> The Symfony Proxy is not BanCapable but I'm not sure what the question here is?

Q: How Use Context hash invalidation is done with Symfony Proxy, given [UserContextInvalidator](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/src/UserContextInvalidator.php#L39) in bundle expects BanCapable, which Symfony Proxy does not implement.

At least I got issue with that when trying to do [Fastly POC](https://github.com/andrerom/sf-london-2018-httpcache-demo/commit/747a2eb4b004b2133526b4cafb92648226dd6d81#diff-3584dcd2a732e6f34941b9194c2ff416R109) using `cache_manager.custom_proxy_client`.

---

@dbu While on that, I also ran into another possible thing we should think about for full Fastly support. And that is mapping FOS's `reverse_proxy_ttl ` header to Fastly's `Surrogate-Control` header: https://docs.fastly.com/guides/performance-tuning/controlling-caching#changing-caching-times-for-different-users

Toflar:
I'm not using the User Context at all, I'm using our https://github.com/terminal42/header-replay-bundle. So I think this probably doesn't work?

andrerom:
+100

andrerom:
> do you plan to work on any of the subjects we discussed here @andrerom to wrap up fastly support?

I havn't planned for that, [what I did for demo at symfony live can be found here](https://github.com/andrerom/sf-london-2018-httpcache-demo/commit/63d3f0a2148d7f10f3bfd9da951ff4b61f9aa8f4). It's not tested against Fastly, but it has support for purging several tags at once when doing hard purging which can also be added here, but besides that it's very similar to the code here.

src/ProxyClient/Fastly.php
andrerom: Seems they are also able to do this on Soft purge, via support there they seem to plan to update doc to make it more clear.

So given Faslty has API rate limits, we should change to use the API linked to above regardless of soft/purge so we can do up to 256 at a time.

dbu:
continued in #451

PR #404, opened 2018-02-01
Update X-Cache-Tags pattern to prevent objects being invalidated when  their tag ends with the tag being invalidated

PeteLawrence:
If you have two objects, one with a tag of 'foobar' and another with a tag of 'bar', currently if you invalidate objects that have a tag of 'bar', you actually invalidate both of these objects since the first objects tag ends in 'bar'.

This pull request should prevent that from happening by only invalidating objects that match the full key.


dbu:
tagged 2.1.1, thanks again for the fix!

PR #408, opened 2018-02-21
Expand AnonymousRequestMatcher by user_identifier_headers option

blankse:
Is needed by https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/425


src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -13,36 +13,62 @@
...
+     *
+     * - user_identifier_headers: List of request headers that authenticate a non-anonymous request.
+     * - session_name_prefix:     Prefix for session cookies. Must match your PHP session configuration.
+     *
+     * @param array $options Options to overwrite the default options
+     *
+     * @throws \InvalidArgumentException if unknown keys are found in $options
      */
-    public function __construct($sessionNamePrefix)
+    public function __construct(array $options = array())
```
dbu: i am afraid i already tagged a version with this class. please remove the type declaration and if $options happens to be a string, treat it as `['session_name_prefix' => $options]` to be BC.

src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -13,36 +13,62 @@
...
+                    // ignore empty cookie header
+                    // ignore cookie header when session_name_prefix is set
+                    continue;
+                }
+
+                return false;
+            }
         }
 
         foreach ($request->cookies as $name => $value) {
```
dbu: the cookies header is hardcoded here - which i think is correct. makes me think we should not pass around Cookies in the list of authorization headers at all.  and then we can get rid of the checks in the loop above.

src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -13,36 +13,62 @@
...
+     * @throws \InvalidArgumentException if unknown keys are found in $options
      */
-    public function __construct($sessionNamePrefix)
+    public function __construct(array $options = array())
     {
-        $this->sessionNamePrefix = $sessionNamePrefix;
+        $resolver = new OptionsResolver();
+        $resolver->setDefaults(array(
+            'user_identifier_headers' => array('Cookie', 'Authorization', 'HTTP_AUTHORIZATION', 'PHP_AUTH_USER'),
+            'session_name_prefix' => 'PHPSESSID',
```
dbu: we always pass values from either the defaults in the listener or the symfony config, right? if thats correct i'd rather make these required and have no defaults to not repeat the defaults in too many places.

src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -13,36 +13,62 @@
...
  */
 class AnonymousRequestMatcher implements RequestMatcherInterface
 {
-    private $sessionNamePrefix;
+    /**
+     * The options configured in the constructor argument or default values.
+     *
+     * @var array
+     */
+    protected $options;
```
dbu: why not private?

src/SymfonyCache/UserContextSubscriber.php
```diff
@@ -51,6 +51,7 @@ class UserContextSubscriber implements EventSubscriberInterface
      *                            match the setup for the Vary header in the backend application.
      * - user_hash_uri:           Target URI used in the request for user context hash generation.
      * - user_hash_method:        HTTP Method used with the hash lookup request for user context hash generation.
+     * - user_identifier_headers: List of request headers that authenticate a non-anonymous request.
```
dbu: this also needs to do to the documentation in the Resources/docs folder.

src/UserContext/AnonymousRequestMatcher.php
blankse: I can't see a tagged version with this class.

dbu: oh, you are right, i missed to merge the 1.x branch to the master branch. so indeed, this is good as is. nice, was not happy that we would need a BC layer for something that has only just been added :-D

src/UserContext/AnonymousRequestMatcher.php
dbu: i think we should remove the function in the bundle, as it should be doing the exact same thing as the function here. or is it doing something different?

blankse:
@dbu I changed the PR

src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -13,36 +13,51 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Matches anonymous requests using a list of identification headers.
  */
 class AnonymousRequestMatcher implements RequestMatcherInterface
 {
-    private $sessionNamePrefix;
+    /**
+     * The options configured in the constructor argument or default values.
```
dbu: there are no default values. i think we can just drop this line and only keep the `@var array`.

src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -13,36 +13,51 @@
...
+    private $options;
 
     /**
-     * @param string $sessionNamePrefix Prefix for session cookies. Must match your PHP session configuration
+     * When creating this matcher, you can configure a number of options.
+     *
+     * - user_identifier_headers: List of request headers that authenticate a non-anonymous request.
+     * - session_name_prefix:     Prefix for session cookies. Must match your PHP session configuration.
+     *
+     * @param array $options Options to overwrite the default options
```
dbu: there are no defaults (which is good). lets instead say "Configuration for the matcher. All options are required because this matcher is usually created by the UserContextSubscriber which provides the default values."

blankse:
@dbu I changed the phpdocs

blankse:
Thanks :+1:

PR #409, opened 2018-03-13
Varnish xkey continued

dbu:
@andrerom i did some cosmetic tweaks to #332 and noticed that we did not mention that the user also needs to change how to tag the response. can you please check if the commit i added makes sense to you?


dbu:
hooray, thanks a lot for your contribution!

now i have to sort out the problem with symfony 4 overwriting the headers when the user is logged in destroying user context caching, and then we are ready for 2.2!

PR #410, opened 2018-03-13
Expand AnonymousRequestMatcher by user_identifier_headers option

dbu:
continuation of #408

### DIFF
--- a/src/UserContext/AnonymousRequestMatcher.php
+++ b/src/UserContext/AnonymousRequestMatcher.php
@@ -13,20 +13,38 @@ namespace FOS\HttpCache\UserContext;
 
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Matches anonymous requests using a list of identification headers.
  */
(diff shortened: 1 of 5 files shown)


src/UserContext/AnonymousRequestMatcher.php
```diff
@@ -13,36 +13,53 @@
...
-            return false;
+        foreach ($this->options['user_identifier_headers'] as $header) {
+            if ($request->headers->has($header)) {
+                return false;
+            }
         }
 
         foreach ($request->cookies as $name => $value) {
-            if (0 === strpos($name, $this->sessionNamePrefix)) {
+            if (0 === strpos($name, $this->options['session_name_prefix'])) {
```
dbu: @blankse wdyt, would it make sense to make it possible to ignore cookies? the old code the bundle had did make it possible by not having the Cookie header in the list of headers... could we say that if the prefix is `''` we do not check cookies? or allow type boolean as well and set the prefix to false?

dbu: i pushed a change to allow that. @blankse can you take a look at e60ad47a4dfe6db929a395bd6c98af9f868c2a2e and tell me if you like it or not?

blankse: @dbu LGTM :+1:

dbu:
i tagged 1.4.5 with this and will now merge the changes up to the master branch so they will be in 2.2 when that is released.

PR #412, opened 2018-03-15
Symfony HttpCache Tag Invalidation: Do not check for missing host

trickreich:
Fixed bug in Symfony Tag Invalidation.

Do not check if host is missing in request creation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+2.1.2
+-----
+
+#### Symfony HttpCache
+
+* Fixed bug in Symfony tag invalidation.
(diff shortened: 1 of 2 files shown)


dbu:
thanks for the fix!

tagged 2.1.3 and pulled the changes to master as well.

dbu:
great, glad to hear its useful!

dbu:
gna, i think travis-ci did not run on this branch at all - as all was green i did not check. there is somethingsomething with how we set up mocking that seems to explode now. @trickreich if you would have a bit of time to look into https://travis-ci.org/FriendsOfSymfony/FOSHttpCache/builds/353840335 i would appreciate it. probably its just some mock method not expecting the extra parameter.

trickreich:
@dbu I'll try to fix them
See https://github.com/FriendsOfSymfony/FOSHttpCache/pull/413 for the process

PR #419, opened 2018-04-12
[RFC] Allow to directly route proxy requests in case of Symfony proxy

Toflar:
For simple single node setups I would like to simplify the Symfony proxy setup.
Right now FOSHttpCache executes a real external `PURGE` or `PURGETAGS` request which forces you to configure and know the target domain which is not necessarily easy for apps like CMS'es that are usually set up on the server directly without knowing the target domain beforehand.

TODO
* [x] Update CHANGELOG

This should be easier for simple single node setups so here's my approach.

Docs:

1. Your `Kernel` has to implement the `HttpCacheProvider` (you can use the `HttpCacheAwareKernel` trait to simplify things):

```php
class MyKernel implements HttpCacheProvider
{
    use HttpCacheAwareKernel;
}
```

2. In the `index.php` you set the `HttpCache` to your `MyKernel` so the bundle is able to access the `HttpCache` instance from the DIC like so:

```php
$kernel = new MyKernel('prod', false);
$httpCache = new HttpCache($kernel);
$kernel->setHttpCache($httpCache);

$request = Request::createFromGlobals();
$response = $httpCache->handle($request);
```

Configuration for the bundle is simple. You only need to add `127.0.0.1` to the servers and use the new `KernelClient` service (see https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/440):

```diff
fos_http_cache:
    proxy_client:
        symfony:
            tags_header: My-Cache-Tags
            tags_method: TAGPURGE
            header_length: 1234
            purge_method: PURGE
            http:   
                servers:
                    - '127.0.0.1'
                http_client: 'fos_http_cache.symfony.kernel_client'
```

The `KernelClient` handles the request by making sure, the IP is set to `127.0.0.1`, converting via PSR-7 requests and responses and sending it directly to the `HttpCache`.

I've integrated everything and it works like a charm. If you like the concept, I'll be happy to finish that with tests.


dbu:
what if instead of the whole cache kernel, we would pass along the store instead? we could write a new ProxyClient instead of a pseudo http client. that client would operate on the store directly. 

apart from permission and sanity checks, the purge listeners simply forward the information to the cache:

* tags is: $tags = explode(',', $request->headers->get($this->tagsHeader)); $store->invalidateTags($tags);
* purge is: $store->purge($request->getUri());

hm, ok, refresh does an actual request. looks like your approach is more consistent. 

open question: what would you think of a combined approach? we inject the HttpCache and do the "simulated" request for refresh, but do direct interactions with the cache store for purge and invalidateTags?

dbu:
and thanks for looking into this! i guess its confusing for many users, and on multi-system setups, you hopefully use a real caching proxy like varnish.

btw, one detail: we do not use `Trait` and `Interface` suffix in this library.

Toflar:
> open question: what would you think of a combined approach? we inject the HttpCache and do the "simulated" request for refresh, but do direct interactions with the cache store for purge and invalidateTags?

Right now, it's a consistent behaviour. It's always a request routed through the kernel, whether it's a real request or not. So it also allows for other listeners and subscribers to do something beforehand or afterwards which whould not be possible anymore if we purge directly. Not saying it's a bad idea, right now the concept would just stay exactly the same. It's just a shortcut with no real network traffic if you will :) I just think there are already so many configuration options, having even more would potentially lead to more confusion. As you can see from the bundle PR, the only thing that needs to be done is adding a new service to keep things simple, nothing more. So configuration remains just the same and just as easy and I kind of like that :)

dbu:
yep, i agree. so lets follow your suggestion of doing this through simulated requests.

please add a bit of class documentation to the relevant classes to explain why we do it like this.

Toflar:
Okay, let's finish this then :-) I've added tests and documentation.
I'm having trouble with not using `Interface` and `Trait` as suffixes because I don't know how to name the interface. Actually both would have the same name if I removed `Interface`. Any proposal for the names?

composer.json
```diff
@@ -24,6 +24,7 @@
         "php": "^5.6 || ^7.0.0",
         "symfony/event-dispatcher": "^2.3 || ^3.0 || ^4.0",
         "symfony/options-resolver": "^2.7 || ^3.0 || ^4.0",
+        "symfony/http-foundation": "^2.4 || ^3.0 || ^4.0",
```
dbu: i think this should be in require-dev & suggest, just like http-kernel itself. its only needed when using the HttpCache (and i guess only when using the new client)

composer.json
```diff
@@ -36,11 +37,15 @@
...
-        "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0"
+        "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0",
+        "symfony/psr-http-message-bridge": "^1.0",
+        "zendframework/zend-diactoros": "^1.7"
     },
     "suggest": {
         "friendsofsymfony/http-cache-bundle": "For integration with the Symfony framework",
-        "monolog/monolog": "For logging issues while invalidating"
+        "monolog/monolog": "For logging issues while invalidating",
+        "symfony/psr-http-message-bridge": "If you need direct routing for Symfony.",
```
dbu: we should be more specific here: "For simplified setup for cache invalidation in a single node system with symfony HttpCache"

composer.json
```diff
@@ -36,11 +37,15 @@
         "php-http/mock-client": "^0.3.2",
         "phpunit/phpunit": "^5.7 || ^6.0",
         "symfony/process": "^2.3 || ^3.0 || ^4.0",
-        "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0"
+        "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0",
+        "symfony/psr-http-message-bridge": "^1.0",
+        "zendframework/zend-diactoros": "^1.7"
```
dbu: we already use php-http for other things. lets use the message factory interface from php-http/message instead (see http://docs.php-http.org/en/latest/message/message-factory.html)

doc/proxy-clients.rst
```diff
@@ -184,6 +184,42 @@ dispatcher as explained above and pass it to the Symfony client::
 
     $symfony = new Symfony($httpDispatcher);
 
+
+The `HttpDispatcher` sends real HTTP requests using any instance of `HttpAsyncClient`
+available in your application (see HTTPlug_ for more information). If you
+have a single node Symfony application setup you might want to directly route
```
dbu: s/node/server/ 

application setup, you can send these requests directly to the cache kernel inside the same PHP process instead of sending actual HTTP requests over the network.

doc/proxy-clients.rst
```diff
@@ -184,6 +184,42 @@ dispatcher as explained above and pass it to the Symfony client::
...
+these requests through the kernel directly instead of having it execute real
+HTTP requests for e.g. performance reasons or just because it seems easier to
+setup as you don't have to know the domain/ip your application is reachable at.
```
dbu: This makes your setup easier as you don't need to know the IP of your server and will also save server resources.

doc/proxy-clients.rst
```diff
@@ -184,6 +184,42 @@ dispatcher as explained above and pass it to the Symfony client::
...
+For that to work, you can pass an instance of `KernelClient` to your `$httpDispatcher`.
+It takes your kernel as the constructor argument which needs to implement
+`HttpCacheAwareKernelInterface` so the instance of `HttpCache` can be accessed
```
dbu: accessed. (this is english, not german. lets make shorter sentences to make things easier to understand ;-) )

src/SymfonyCache/HttpCacheAwareKernelInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Symfony\Component\HttpKernel\HttpCache\HttpCache;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+interface HttpCacheAwareKernelInterface extends HttpKernelInterface
+{
+    /**
+     * @param HttpCache $httpCache
+     */
+    public function setHttpCache(HttpCache $httpCache);
```
dbu: imho this should not be part of the interface. its an implementation detail for your bootstrap. (its fine to have it in the trait though - but when bootstrapping, the user knows what trait the actual kernel class implements)

src/SymfonyCache/KernelClient.php
```diff
@@ -0,0 +1,110 @@
...
+use Http\Promise\FulfilledPromise;
+use Psr\Http\Message\RequestInterface;
+use Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory;
+use Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Zend\Diactoros\ServerRequest;
+
+/**
+ * An implementation of HttpAsyncClient that allows direct routing through the
```
dbu: i would not call this "routing", but talk about "calling"

src/SymfonyCache/KernelClient.php
```diff
@@ -0,0 +1,110 @@
...
+ * Symfony HttpCache kernel without executing a real HTTP request.
+ * It uses the HttpFoundationFactory and the Zend DiactorosFactory to convert
+ * between PSR-7 requests and responses. Both are optional dependencies of this
+ * package thus existence of the respective classes is checked for in the
+ * constructor of this client. It is only needed if you have a single node
```
dbu: It is only needed if => It can only be used if

and i would put this information above the details about converting requests. (ah and that will need to change anyways when we use php-http instead)

src/SymfonyCache/KernelClient.php
```diff
@@ -0,0 +1,110 @@
...
+ * setup of Symfony and serves as kind of a shortcut for easier configuration.
+ * If you use Varnish or have a multiple node Symfony setup, this class is entirely
+ * useless to you and you can happily ignore it.
```
dbu: s/class is entirely useless/client can not be used/

src/SymfonyCache/HttpCacheAwareKernel.php
```diff
@@ -0,0 +1,41 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\SymfonyCache;
+
+use Symfony\Component\HttpKernel\HttpCache\HttpCache;
+
+/**
+ * Trait for enhanced Symfony reverse proxy based on the Symfony kernel component.
```
dbu: this could say: "Trait to implement the HttpCacheProvider interface."

src/SymfonyCache/HttpCacheAwareKernelInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\SymfonyCache;
+
+use Symfony\Component\HttpKernel\HttpCache\HttpCache;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+interface HttpCacheAwareKernelInterface extends HttpKernelInterface
```
dbu: why do we extend the kernel interface here? i think the consuming code does not care what is providing the http cache. this could be `HttpCacheProvider` and the trait `HttpCacheAware`

doc/proxy-clients.rst
```diff
@@ -184,6 +184,42 @@ dispatcher as explained above and pass it to the Symfony client::
...
+    // You can use the HttpCacheAwareKernel trait to simplify things.
+    $kernel = new App\Kernel();
+
+    $httpCache = new HttpCache($kernel);
+
+    // This is where your kernel now becomes aware of the HttpCache instance
+    $kernel->setHttpCache($httpCache);
+
+    // Now let's create the HttpDispatcher with our KernelClient
+    $kernelClient = new KernelClient($kernel);
```
dbu: why not inject the HttpCache directly here? would that not be a lot simpler? (we then need some solution for the bundle, of course. but the library seems not to need the indirection i think)

src/SymfonyCache/HttpCacheAwareKernelInterface.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    /**
+     * @param HttpCache $httpCache
+     */
+    public function setHttpCache(HttpCache $httpCache);
+
+    /**
+     * @return HttpCache
```
dbu: do we need a HttpCache specifically? i think the client only cares about `handle`, so we could use `KernelInterface` as the thing (but name it $httpCache nonetheless)

src/SymfonyCache/KernelClient.php
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Converts a PSR-7 request to a Symfony HttpFoundation request, runs
+     * it through the HttpCache kernel and converts the response back to a PSR-7
+     * response.
+     *
+     * {@inheritdoc}
+     */
+    public function sendAsyncRequest(RequestInterface $request)
+    {
+        $symfonyRequest = Request::createFromGlobals();
```
dbu: should we not do `new Request` here? otherwise we might sneak accidental things from whatever request was going on that is triggering the invalidation.

src/SymfonyCache/KernelClient.php
```diff
@@ -0,0 +1,110 @@
...
+            [],
+            $request->getUri(),
+            $request->getMethod(),
+            $request->getBody(),
+            $request->getHeaders(),
+            $symfonyRequest->cookies->all(),
+            $query
+        );
+
+        $symfonyRequest = $this->httpFoundationFactory->createRequest($serverRequest);
```
dbu: why first create a PSR server request and then convert it to a symfony request? we do have to convert the request we get from the http dispatcher but i think we could create a symfony server request directly, no?

src/SymfonyCache/KernelClient.php
```diff
@@ -0,0 +1,110 @@
...
+ * between PSR-7 requests and responses. Both are optional dependencies of this
+ * package thus existence of the respective classes is checked for in the
+ * constructor of this client. It is only needed if you have a single node
+ * setup of Symfony and serves as kind of a shortcut for easier configuration.
+ * If you use Varnish or have a multiple node Symfony setup, this class is entirely
+ * useless to you and you can happily ignore it.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class KernelClient implements HttpAsyncClient
```
dbu: could we instead of having a "mock" http client, create a Dispatcher interface that the HttpDispatcher implements, and that this class would implement? and then refactor a base class out of HttpDispatcher for the validatation and queuing mechanism, but with a different `flush`.

composer.json
Toflar: I think it's okay to leave it here. 2.7 and 2.8 are the only supported Symfony version left of the 2.x branch. The 2.3 branch has lost its support for a year now. Nobody will be using http foundation lower than 2.4, I'm fairly certain of that :)

composer.json
Toflar: You're right, done in b04699281a4eda1121edbd41a8331f7afdbf116e.

composer.json
Toflar: Good input, done in db97a74d860673d69a6445a73535ffe1ff533164.

src/SymfonyCache/HttpCacheAwareKernelInterface.php
Toflar: What should not be part of this interface?

composer.json
Toflar: Ah well, I still need a `ServerRequest` to get a proper Symfony `Request` instance. Not sure I can get rid of Diactoros here, wdyt?

src/SymfonyCache/HttpCacheAwareKernelInterface.php
Toflar: It should still have a `handle()` method and thus be a valid kernel ;-)

doc/proxy-clients.rst
Toflar: I have no clue how that should work for the bundle so that's the way I would like to go.

src/SymfonyCache/KernelClient.php
Toflar: Yeah wasn't sure about that. I've used `new Request()` first but then it's just probably missing a lot of other info (although that is pretty surely entirely irrelevant info for the cache purging process).

src/SymfonyCache/KernelClient.php
Toflar: I don't understand this comment.

dbu: if i am not mistaken, we have a method that gets a psr request, creates a symfony request, creates a diactoros server request and then converts the diactoros server request to a symfony server request that is passed to the handle. or am i misreading this?

composer.json
dbu: i am not concerned about the version constraint, but about adding a new dependency that will be installed even when you don't use symfony at all.

src/SymfonyCache/HttpCacheAwareKernelInterface.php
dbu: the setter

src/SymfonyCache/HttpCacheAwareKernelInterface.php
dbu: no, we never call handle on this, we only call getHttpCache

src/SymfonyCache/KernelClient.php
dbu: we don't do that in the other invalidation requests, so i think it should be safe. if we need specific things, we should get the original request from the RequestStack service and copy the things. (Host header should probably be present, for example)

composer.json
Toflar: Okay will move to dev deps.

src/SymfonyCache/KernelClient.php
Toflar: You are misreading I guess. We get a PSR-7 request and need a Symfony Request.

src/SymfonyCache/HttpCacheAwareKernelInterface.php
Toflar: Ah, I see! Removed in 6964a8b3aa914bbad5442bb9c8a979505c7afe35.

src/SymfonyCache/HttpCacheAwareKernelInterface.php
Toflar: That's right, removed in ca31460557d3d63577048b82fb632c3cf98ecac8.

src/SymfonyCache/KernelClient.php
```diff
@@ -0,0 +1,111 @@
...
+ * as kind of a shortcut for easier configuration.
+ * If you use Varnish or have a multiple node Symfony setup, this client is entirely
+ * useless and cannot be used.
+ * It HttPlug libraries to convert between PSR-7 requests and responses. Both are
+ * optional dependencies of this package thus existence of the respective classes
+ * is checked for in the constructor of this client.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class KernelClient implements HttpAsyncClient
```
dbu: >> HttpDispatcher > Dispatcher

> That was my first idea but that doesn't work for BC reasons. Every proxy client expects an instance of HttpDispatcher right now, not of any interface.

this is constructor injection only. so we could change the constructor of HttpProxyClient to say Dispatcher (which works even if some custom extending class says HttpDispatcher because that is a Dispatcher). and the interface has both public methods invalidate & flush, so they are compatible. i think we can sneak this through without any risks for BC?

src/SymfonyCache/KernelClient.php
Toflar: There is no request stack at this point. Well there is but no current request. We are talking about the `kernel.terminate` event there. I used to have this in the first iteration but then noticed that and thus my fallback was the globals :)

src/ProxyClient/Dispatcher.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\Exception\ExceptionCollection;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Queue and send HTTP requests with a Httplug asynchronous client.
+ *
+ * @author David Buchmann <mail@davidbu.ch>
```
dbu: you could put your name here

src/ProxyClient/Dispatcher.php
```diff
@@ -0,0 +1,45 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\Exception\ExceptionCollection;
+use Psr\Http\Message\RequestInterface;
+
+/**
+ * Queue and send HTTP requests with a Httplug asynchronous client.
```
dbu: this needs to be adjusted. maybe just "Queue and send HTTP invalidation requests."

src/SymfonyCache/HttpCacheAwareKernel.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+namespace FOS\HttpCache\SymfonyCache;
+
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+/**
+ * Trait to implement the HttpCacheProvider interface.
+ */
+trait HttpCacheAwareKernel
```
dbu: s/HttpCacheAwareKernel/HttpCacheAware/

src/SymfonyCache/KernelDispatcher.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * @var array
+     */
+    private $queue = [];
+
+    /**
+     * KernelClient constructor.
+     *
+     * @param HttpCacheProvider $kernel
+     */
```
dbu: please remove this autogenerated phpdoc, it does not add anyhting thats not in the code already.

src/SymfonyCache/KernelDispatcher.php
```diff
@@ -0,0 +1,116 @@
...
+        // Add headers
+        $headers = $invalidationRequest->getHeaders();
+        foreach ($headers as $name => $values) {
+            $name = strtolower($name);
+
+            if ('cookie' === $name) {
+                foreach ($values as $value) {
+                    foreach (explode(';', $value) as $cookieString) {
+                        $cookie = Cookie::fromString($cookieString);
+                        $request->cookies->add([$cookie->getName() => $cookie->getValue()]);
```
dbu: is the symfony request overwriting the cookies header if we do not parse things like this?

src/SymfonyCache/KernelDispatcher.php
```diff
@@ -0,0 +1,116 @@
...
+ * useless and cannot be used.
+ *
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
+ */
+class KernelDispatcher implements Dispatcher
+{
+    /**
+     * @var HttpCacheProvider
+     */
+    private $kernel;
```
dbu: i'd call this $httpCacheProvider instead of $kernel. we don't care what provides us with the HttpCache.

src/SymfonyCache/KernelDispatcher.php
```diff
@@ -0,0 +1,115 @@
...
+        // Add headers
+        $headers = $invalidationRequest->getHeaders();
+        foreach ($headers as $name => $values) {
+            $name = strtolower($name);
+
+            if ('cookie' === $name) {
+                foreach ($values as $value) {
+                    foreach (explode(';', $value) as $cookieString) {
+                        $chunks = explode('=', $cookieString, 2);
+                        $request->cookies->add([trim($chunks[0]) => trim($chunks[1])]);
```
dbu: if we simply set the cookie header like the other headers, is symfony overwriting it?

Toflar: The `Request` class of Symfony unfortunately has no handling for `cookie` in the request class. It expects you to set them from the outside (usually `$_COOKIE`) so I had to implement custom handling here. I kind of expected that to happen which is why I put it into the unit test right from the beginning :D

dbu: on http level, cookies is just another header. is setting that header destroying things? (i guess we do care, because `refresh` in combination with user context - although that would be a bit random, it would only refresh for the group of the client triggering the refresh)

Toflar:
Should be ready for another review.

src/SymfonyCache/KernelDispatcher.php
dbu: i would have expected that setting the cookie header would work, but i see, this is parsed from the $_COOKIE only. i think now we do the correct thing, convert the $invalidationRequest cookies into symfony request cookies so they are processed :+1:

Toflar:
> instead of the provider, we could set the cache as a request attribute, no? but that would be a bit weird as well...

Request attributes are the new superglobals... ;-)

Changelog entry is added. I'll test the integration and update the bundle accordingly so please do not merge yet, maybe it requires changes (but I don't think so).

doc/proxy-clients.rst
```diff
@@ -184,6 +184,39 @@ dispatcher as explained above and pass it to the Symfony client::
...
     $symfony = new Symfony($httpDispatcher);
 
+
+The `HttpDispatcher` sends real HTTP requests using any instance of `HttpAsyncClient`
+available in your application (see HTTPlug_ for more information). If you
+have a single server Symfony application setup, you can send these requests
+directly to the cache kernel inside the same PHP process instead of sending actual
+HTTP requests over the network. This makes your setup easier as you don't need
+to know the IP of your server and will also save server resources.
+For that to work, you can use the `KernelDispatcher` instead of the `HttpDispatcher`.
```
dbu: in rst. you need double backticks for code. single backtick is only for references.

dbu:
thanks a lot! i cleaned up the doc a bit and squashed and merged.

i also bumped the branch alias version to 2.3. i won't release immediately, in case other things come up in the near future.

dbu:
merged in d3b7a1cf596be16129cbe4a3e8ca507cbdd682a0

PR #421, opened 2018-07-04
Disable escaping of tags when using Xkey

### LINKED ISSUE #420: Tags being incorrectly(?) escaped when using Varnish's Xkey module
Currently when Varnish::invalidateTags($tags) is called, the tags contained within $tags are all escaped using preg_quote. eg. `$escapedTags = array_map('preg_quote', $this->escapeTags($tags));`
  
This makes sense when invalidating tags using the original BAN method since the contents of the X-Cache-Tags header gets used in a regular expression match within the fos_ban.vcl file, but not when using the new Xkey method, as the escaped tag may no longer match the original tag if it contains any special characters.

eg. We tag an API response containing details of a particular customer with a tag of 'customer-1234', however when we try to invalidate this tag, the tag becomes 'customer\\-1234' within the invalidation request that gets made to Varnish, and as a result the API response does not get cleared.

I'll happily create a PR if others agree that this is how it should be working, and not me misunderstanding something.

PeteLawrence:
fix #420

### DIFF
--- a/tests/Unit/ProxyClient/VarnishTest.php
+++ b/tests/Unit/ProxyClient/VarnishTest.php
@@ -87,7 +87,7 @@ class VarnishTest extends TestCase
     public function testPurgekeys()
     {
         $options = [
-            'tag_mode' => 'purgekeys'
+            'tag_mode' => 'purgekeys',
         ];
 
         $varnish = new Varnish($this->httpDispatcher, $options);
@@ -109,7 +109,7 @@ class VarnishTest extends TestCase
     {
         $options = [
             'tag_mode' => 'purgekeys',
-            'tags_header' => 'xkey-purge'
+            'tags_header' => 'xkey-purge',
         ];
 
         $varnish = new Varnish($this->httpDispatcher, $options);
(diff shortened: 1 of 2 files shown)


tests/Unit/ProxyClient/VarnishTest.php
```diff
@@ -84,6 +84,49 @@ function (RequestInterface $request) {
...
+        );
+
+        $varnish->invalidateTags(['post-1', 'post,type-3']);
+    }
+
+    public function testHardPurgekeys()
+    {
+        $options = [
+            'tag_mode' => 'purgekeys',
+            'tags_header' => 'xkey-purge'
```
XWB: Missing `,`

tests/Unit/ProxyClient/VarnishTest.php
```diff
@@ -84,6 +84,49 @@ function (RequestInterface $request) {
         $varnish->banPath('/articles/.*', 'text/html', $hosts);
     }
 
+    public function testPurgekeys()
+    {
+        $options = [
+            'tag_mode' => 'purgekeys'
```
XWB: Missing `,`

dbu:
merged in 88e4d4280aea8ff1941402834a5fa01411823c3d
2.3.1 is underway.

PR #422, opened 2018-07-27
Allow cleaning up cache tags headers for the Symfony proxy

Toflar:


src/SymfonyCache/CleanupCacheTagsListener.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Remove the cache headers.
+     *
+     * @param CacheEvent $e
+     */
+    public function cleanResponse(CacheEvent $e)
+    {
+        if (null === ($response = $e->getResponse())) {
```
XWB: `if (null === $response = $e->getResponse())` should be fine.

doc/symfony-cache-configuration.rst
```diff
@@ -240,14 +245,17 @@ To get cache tagging support, register the ``PurgeTagsListener`` and use the
         SurrogateInterface $surrogate = null,
         array $options = []
     ) {
+        $tagsHeader = 'X-Cache-Tags';
```
dbu: i think best practice would be to define this as a `const` on the class. or use `TagHeaderFormatter::DEFAULT_HEADER_NAME` but if this example wants to show how to change the header it makes sense to have our own constant

src/SymfonyCache/CleanupCacheTagsListener.php
```diff
@@ -0,0 +1,60 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\SymfonyCache;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Listener that allows to cleanup the cache tags header so it's not exposed
```
dbu: i would prefer to not repeat "cleanup" but say something like:

"Listener to remove the cache tags header before the response is delivered to the client so it's not exposed to the world."

src/SymfonyCache/CleanupCacheTagsListener.php
```diff
@@ -0,0 +1,60 @@
...
+{
+    /**
+     * @var string
+     */
+    private $tagsHeader;
+
+    /**
+     * @param string $tagsHeader The header that is used for cache tags
+     */
+    public function __construct($tagsHeader = PurgeTagsListener::DEFAULT_TAGS_HEADER)
```
dbu: its the same value, but this should be `TagHeaderFormatter::DEFAULT_HEADER_NAME` because its about the header used to mark tags on a response. the purge constant is about which header is used to tell what tags to purge... its a bit confusing that we chose the same default for both those headers...

src/SymfonyCache/CleanupCacheTagsListener.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->tagsHeader = $tagsHeader;
+    }
+
+    /**
+     * Remove the cache headers.
```
dbu: lets say "Remove the tags header." - otherwise this is confusing. or we could even call the method `removeTagsHeader` and remove the whole phpdoc because then the method name exactly tells what it does.

dbu:
can you please add a changelog entry?

is there anything to add to the bundle as well? or does the symfony cache kernel doc not talk about the setup and just point to the library?

Toflar:
Updated and included everything that was commented on.
The bundle does say nothing about the kernel and its subscribers, it's all in the library here 😄

Toflar:
Can you re-run the tests? They do not fail because of the unit tests apparently 😄

dbu:
thanks! i merged on the cli. changed the branch alias and tweaked the changelog as 2.4 is not released yet

PR #423, opened 2018-07-30
Do not add the same tag multiple times

Toflar:


tests/Unit/ResponseTaggerTest.php
```diff
@@ -132,4 +132,15 @@ public function testNonStrictEmptyTag()
         $this->assertTrue($tagHandler->hasTags());
         $this->assertEquals('post-1', $tagHandler->getTagsHeaderValue());
     }
+
+    public function testUniqueTags()
+    {
+        $headerFormatter = new CommaSeparatedTagHeaderFormatter('FOS-Tags');
+
+        $tagHandler = new ResponseTagger(['header_formatter' => $headerFormatter, 'strict' => true]);
+        $tagHandler->addTags(['post-1', 'post-2', 'post-1']);
```
dbu: for the sake of completeness, i'd like to have another line here that does `$tagHandler->addTags(['post-2', 'post-3']);`. with the current test, `array_merge($this->tags, array_unique($filtered))` would pass but its not correct.

src/ResponseTagger.php
```diff
@@ -123,7 +123,7 @@ public function addTags(array $tags)
             throw new InvalidTagException('Empty tags are not allowed');
         }
 
-        $this->tags = array_merge($this->tags, $filtered);
+        $this->tags = array_unique(array_merge($this->tags, $filtered));
```
dbu: i am pondering whether we should do the unique only at the end when building the tags header. not sure whats better performance - doing unique only at the end in getTagsHeaderValue would add less overhead for people who's system is not adding duplicate tags. on the other hand, a system with massive duplicates would profit from a rolling de-duplication. also depends on whether addTags is called a lot or only a few times... wdyt?

Toflar: I asked myself the same question but in the end I guess it's such a minor thing, nobody will ever notice 😄 I put it into the `addTags()` because there's already some filtering going on there. I just figured it would be better to add it there so it's all in one place. But again, at the end of the day it doesn't really matter I think 😊

dbu: yeah, agreed

dbu:
can you please add the changelog entry? otherwise this looks good to me

Toflar:
Done.

dbu:
i tagged the changes as 2.4.0

thanks a lot for the contributions!

Toflar:
Yeyy, thank you!!!

PR #424, opened 2018-08-09
[RFC] Added MaxHeaderValueLengthFormatter

Toflar:
This is an idea on how we could try to solve the maximum length of the header values in the `ResponseTagger`. We already have `TagHeaderFormatters` so I thought I'll write one that takes an inner formatter and splits the tags of this one to make sure it does not exceed the max length. This would allow to use this formatter to limit the header value length of any other formatter (the built in CSV or your very own) very easily (a pretty cool concept I find 😄 ).

For this to work, however, I had to adjust the return value of the `TagHeaderFormatter` interface and I know that in theory this is a BC break. **BUT:** I wanted to discuss this here because it's really only a theoretical BC break. So what I did is I changed this:

```php
    /**
     * Get the value for the HTTP tag header.
     *
     * This concatenates all tags and ensures correct encoding.
     *
     * @param array $tags
     *
     * @return string
     */
    public function getTagsHeaderValue(array $tags);
```

to this:

```php
    /**
     * Get the value for the HTTP tag header.
     *
     * This concatenates all tags and ensures correct encoding.
     *
     * @param array $tags
     *
     * @return string|string[]
     */
    public function getTagsHeaderValue(array $tags);
```

So basically, it now allows an array of strings to be returned. This means all the existing formatters that return a string are just fine. Only the ones **calling** the method must also deal with an array return value. And this in fact should've **always been the case** because both, the PSR-7 `Response::withHeader()` as well as the Symfony `$response->headers->set()` accept an array to support multi-value headers. So, in theory this is a BC break but I don't think anyone will be affected.

Wdyt?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,10 @@ matrix:
     - php: 7.2
       env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1
 
+    # Test Symfony LTS versions
+    - php: 7.2
+      env: DEPENDENCIES="symfony/lts:^3 toflar/psr6-symfony-http-cache-store:^1.0"
+
       # Latest commit to master
     - php: 7.2
       env: STABILITY="dev"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ cache:
 env:
   global:
     - VARNISH_VERSION=5.1
-    - DEPENDENCIES="toflar/psr6-symfony-http-cache-store:^1.0"
+    - DEPENDENCIES="toflar/psr6-symfony-http-cache-store:^1.1.2"
 
 matrix:
   fast_finish: true
@@ -30,7 +30,7 @@ matrix:
     - php: 7.2
       env: DEPENDENCIES="symfony/lts:^2"
     - php: 7.2
-      env: DEPENDENCIES="symfony/lts:^3 toflar/psr6-symfony-http-cache-store:^1.0"
+      env: DEPENDENCIES="symfony/lts:^3 toflar/psr6-symfony-http-cache-store:^1.1.2"
 
       # Latest commit to master
     - php: 7.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,15 +26,10 @@ matrix:
     - php: 7.2
       env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1
 
-      # Test Symfony LTS versions
-    - php: 7.2
-      env: DEPENDENCIES="symfony/lts:^2"
-    - php: 7.2
-      env: DEPENDENCIES="symfony/lts:^3 toflar/psr6-symfony-http-cache-store:^1.1.2"
-
       # Latest commit to master
     - php: 7.2
       env: STABILITY="dev"
+
   allow_failures:
     # Dev-master is allowed to fail.
     - env: STABILITY="dev"
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,9 @@
         "symfony/process": "^3.4 || ^4.0",
         "symfony/http-kernel": "^3.4 || ^4.0"
     },
+    "conflict": {
+        "toflar/psr6-symfony-http-cache-store": "<1.1.2"
+    },
     "suggest": {
         "friendsofsymfony/http-cache-bundle": "For integration with the Symfony framework",
         "monolog/monolog": "For logging issues while invalidating"
--- a/composer.json
+++ b/composer.json
@@ -27,8 +27,7 @@
         "php-http/client-implementation": "^1.0.0",
         "php-http/client-common": "^1.1.0",
         "php-http/message": "^1.0.0",
-        "php-http/discovery": "^1.0",
-        "toflar/psr6-symfony-http-cache-store": "^1.1.2"
+        "php-http/discovery": "^1.0"
     },
     "require-dev": {
         "mockery/mockery": "~0.9.5",
--- a/composer.json
+++ b/composer.json
@@ -22,12 +22,13 @@
     ],
     "require": {
         "php": "^5.6 || ^7.0.0",
-        "symfony/event-dispatcher": "^2.3 || ^3.0 || ^4.0",
-        "symfony/options-resolver": "^2.7 || ^3.0 || ^4.0",
+        "symfony/event-dispatcher": "^3.4 || ^4.0",
+        "symfony/options-resolver": "^3.4 || ^4.0",
         "php-http/client-implementation": "^1.0.0",
         "php-http/client-common": "^1.1.0",
         "php-http/message": "^1.0.0",
-        "php-http/discovery": "^1.0"
+        "php-http/discovery": "^1.0",
+        "toflar/psr6-symfony-http-cache-store": "^1.1.2"
     },
     "require-dev": {
         "mockery/mockery": "~0.9.5",
@@ -35,8 +36,8 @@
         "php-http/guzzle6-adapter": "^1.0.0",
         "php-http/mock-client": "^0.3.2",
         "phpunit/phpunit": "^5.7 || ^6.0",
-        "symfony/process": "^2.3 || ^3.0 || ^4.0",
-        "symfony/http-kernel": "^2.3 || ^3.0 || ^4.0"
+        "symfony/process": "^3.4 || ^4.0",
+        "symfony/http-kernel": "^3.4 || ^4.0"
     },
     "suggest": {
         "friendsofsymfony/http-cache-bundle": "For integration with the Symfony framework",
@@ -54,7 +55,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.4.x-dev"
+            "dev-master": "2.5.x-dev"
         }
     }
 }
--- a/src/ProxyClient/Symfony.php
+++ b/src/ProxyClient/Symfony.php
@@ -58,13 +58,13 @@ class Symfony extends HttpProxyClient implements PurgeCapable, RefreshCapable, T
             'purge_method' => PurgeListener::DEFAULT_PURGE_METHOD,
             'tags_method' => PurgeTagsListener::DEFAULT_TAGS_METHOD,
             'tags_header' => PurgeTagsListener::DEFAULT_TAGS_HEADER,
-            'tags_path' => '/',
+            'tags_invalidate_path' => '/',
             'header_length' => 7500,
         ]);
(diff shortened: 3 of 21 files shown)


tests/Unit/TagHeaderFormatter/MaxHeaderValueLengthFormatterTest.php
```diff
@@ -0,0 +1,73 @@
...
+        );
+    }
+
+    public function tooLongProvider()
+    {
+        return [
+            [3, ['foo', 'bar', 'baz'], ['foo', 'bar', 'baz']],
+            [6, ['foo', 'bar', 'baz'], ['foo', 'bar', 'baz']], // with the , that equals 7
+            [7, ['foo', 'bar', 'baz'], ['foo,bar', 'baz']],
+            [50, ['foo', 'bar', 'baz'], 'foo,bar,baz'], // long enough, must not be an array
```
dbu: why is it an array for 7 but not for 8+? that seems odd to me

src/TagHeaderFormatter/MaxHeaderValueLengthFormatter.php
```diff
@@ -0,0 +1,120 @@
...
+        $this->inner->getTagsHeaderName();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTagsHeaderValue(array $tags)
+    {
+        $values = (array) $this->inner->getTagsHeaderValue($tags);
+        $newValues = [[]];
```
dbu: should this not be a single `[]`?

tests/Unit/TagHeaderFormatter/MaxHeaderValueLengthFormatterTest.php
Toflar: Max length `6` is too short for `foo,bar` (7 chars) so it must result in two header values and 7 must result in one.

src/TagHeaderFormatter/MaxHeaderValueLengthFormatter.php
Toflar: Nope, this is because of `array_merge(...$newValues);`. You could also call `array_merge()` in every loop but it's a performance gain if you only merge it once :)

Toflar:
> i think the plan makes sense. i would be more relaxed to merge this if we can build a functional test with all caches that support tagging to prove that multiple tag headers will indeed work as expected (e.g. invalidation happens both if tag is in the first or in the second header)

As far as I can see, no proxy currently tests cache tag invalidation. I could only help with the SF proxy test anyway where I would need to add a dependency to my alternative PSR-6 store.

> i am a bit worried about the recursive splitting function you implemented. we potentially build several headers without then using them if they are too long. i think i would prefer to change the csv implementation to do that.

I don't think this makes sense. The way I built it, the `MaxHeaderValueLengthFormatter` can be applied to any other formatter you like. No matter if it's CSV or anything else. 
Also, let's be realistic here. It's so fast, it's just irrelevant. And chances that you actually hit the maximum are already low. Plus, even if you hit the maximum, chances that it needs multiple runs are really, really small because we split the whole tags array into halves.
I like that you think about these kind of things but I don't think there's anything to worry about here, really 😄

tests/Unit/TagHeaderFormatter/MaxHeaderValueLengthFormatterTest.php
dbu: sorry, i missed the the line with 7 still has 2 elements. all fine with that then.

dbu:
okay, agreed that the header should usually not need to be split many times.

and indeed, we miss functional tests for tagging. we do have purge/refresh/ban in
https://github.com/FriendsOfSymfony/FOSHttpCache/tree/master/tests/Functional/ProxyClient but miss tag tests. would be great if we could add them to be sure the multiple headers really work as expected. do you think you could add something for that? i guess they should go into a new test case in https://github.com/FriendsOfSymfony/FOSHttpCache/tree/master/tests/Functional/Symfony

can you please add something to the documentation for how to use this? and do you plan on adding this to the bundle afterwards?

Toflar:
Pheeew, that was a hard nut to crack 😄 But I've added functional tests for cache tag invalidation in case of Symfony now and covered both cases, single and multiple headers. Of course that only works if my PSR-6 cache store is installed so I'm not sure if everything is going to work out fine on travis yet.

> can you please add something to the documentation for how to use this?

Will do, once the PR is in ready-to-merge-state so we don't document stuff that's not true 😄 

> and do you plan on adding this to the bundle afterwards?

Yes I do.

Toflar:
@dbu maybe you can help with the failing tests?

src/ProxyClient/Symfony.php
```diff
@@ -58,11 +58,13 @@ protected function configureOptions()
             'purge_method' => PurgeListener::DEFAULT_PURGE_METHOD,
             'tags_method' => PurgeTagsListener::DEFAULT_TAGS_METHOD,
             'tags_header' => PurgeTagsListener::DEFAULT_TAGS_HEADER,
+            'tags_path' => '/',
```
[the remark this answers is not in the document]
Toflar: I agree here, changed in https://github.com/FriendsOfSymfony/FOSHttpCache/pull/424/commits/e6bfbf2ceb19737c704ba0d19feb4cc310b92b57.

Toflar:
I've quickly tried to check but I would need to set up a docker container to replicate php 5.6 which is EOL very, very soon anyway too. Tbh, I would like to invest my time in more productive things than keeping support for versions nobody should be using anyway. If anybody really needs it, I think it should be them investing the time, not us 😄 Thus, I've bumped the minimum required versions.
One test is still failing but that looks like a time related thing which could be fixed by restarting the build.

Looks like this is going to make it so I'll be adding some docs soon.

dbu:
the php 5.6 build with symfony 3.4 seems to work. agree that we should not waste too much time on supporting outdated things. people stuck on legacy can still use the 2.4 tag. i guess FOSHttpCache 3.0 should be PHP ^7.1 . (not in a hurry to do that, though)

composer.json
```diff
@@ -22,21 +22,22 @@
...
-        "symfony/event-dispatcher": "^2.3 || ^3.0 || ^4.0",
-        "symfony/options-resolver": "^2.7 || ^3.0 || ^4.0",
+        "symfony/event-dispatcher": "^3.4 || ^4.0",
+        "symfony/options-resolver": "^3.4 || ^4.0",
         "php-http/client-implementation": "^1.0.0",
         "php-http/client-common": "^1.1.0",
         "php-http/message": "^1.0.0",
-        "php-http/discovery": "^1.0"
+        "php-http/discovery": "^1.0",
+        "toflar/psr6-symfony-http-cache-store": "^1.1.2"
```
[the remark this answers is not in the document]
Toflar: Ah that was accidentally, fixed, sorry.

.travis.yml
```diff
@@ -26,15 +26,10 @@ matrix:
     - php: 7.2
       env: VARNISH_VERSION=4.1 VARNISH_MODULES_VERSION=0.9.1
 
-      # Test Symfony LTS versions
-    - php: 7.2
-      env: DEPENDENCIES="symfony/lts:^2"
-    - php: 7.2
-      env: DEPENDENCIES="symfony/lts:^3 toflar/psr6-symfony-http-cache-store:^1.0"
```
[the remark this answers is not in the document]
Toflar: As we've bumped to 3.4 as a minimum requirement I thought `--prefer-lowest` would already cover that as "lowest" === "3.4" now. Wrong?

Toflar:
Docs ready for review.

.travis.yml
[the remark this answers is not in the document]
dbu: i'd keep both so that we know if its only the oldest symfony 3.4 that fails or also the newest 3.4. (and also the other way round, we did have some regressions in the past, where some change in a patch version of old symfony broke stuff)

.travis.yml
```diff
@@ -7,7 +7,7 @@ cache:
 env:
   global:
     - VARNISH_VERSION=5.1
-    - DEPENDENCIES="toflar/psr6-symfony-http-cache-store:^1.0"
+    - DEPENDENCIES="toflar/psr6-symfony-http-cache-store:^1.1.2"
```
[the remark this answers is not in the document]
Toflar: Done.

Toflar:
Docs updated.

dbu:
can you please add the changelog entry and make the part about the proxy having to support multiple headers a `note`?

the lowest version build failed, i restarted it to see if its a fluke. i wonder if those tests first check that we have a 2xx response first, or whether the real problem is some error from the server... we should probably output the response body when the assertion fails, to help debugging.

dbu:
hah, now we see more. so on php 5.6 the toflar store is not found. which seems odd - do we not add it to the project?

the other tests have a problem with mockery and the getHeaders method...

tests/Functional/Fixtures/web/symfony.php
```diff
@@ -14,13 +14,14 @@
...
 use Symfony\Component\HttpKernel\HttpCache\Store;
+use Toflar\Psr6HttpCacheStore\Psr6Store;
 
 $loader = require_once __DIR__.'/../../../../vendor/autoload.php';
 
 $symfonyProxy = new SymfonyProxy();
 
 $kernel = new AppKernel();
-$kernel = new AppCache($kernel, new Store($symfonyProxy->getCacheDir()), null, ['debug' => true]);
+$kernel = new AppCache($kernel, new Psr6Store(['cache_directory' => $symfonyProxy->getCacheDir(), 'cache_tags_header' => 'X-Cache-Tags']), null, ['debug' => true]);
```
dbu: the lowest version build clears the dependencies. i think we should only try to use the Psr6 store if class_exists and otherwise use the normal store.

andrerom:
> @andrerom afaik ez was using an older symfony version. do you see a problem with this change?

We are still on 1.x, started looking into moving to 2.x now that xkey is in, but it's invalidates most of our existing code so no blockers from us ;)

dbu:
yep. why is the test for purge talking about the tags? do we have a wrong dependency here?

Toflar:
No, no. I forgot half of it, I'm on it :)

dbu:
cheers!

do you want to work on the bundle integration before i tag the release? could be a nice further sanity check ;-)

Toflar:
I will be but I'm on holidays for 3 weeks now so that'll take a while 😄 So no need to tag that right now. Thanks!

dbu:
alright! enjoy your holidays then!

PR #425, opened 2018-09-10
Symfony Client: Support Tags: Check if needed dependency is installed

trickreich:
Fix https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/issues/478


dbu:
added a changelog entry and merged manually!
thanks, i will release 2.5.1 with this fix now.

trickreich:
I think it's not possible to check if the store is a instance of Psr6, or am I wrong?

trickreich:
And: Thanks for the fast response! :+1:

dbu:
yes, i would not see how that would be possible. we could make it a configuration option to set but that would be weird too.

btw, this is going to be 2.5.0, i did not realise that 2.5.0 has not yet been released :-)

and thanks again for the improvement!

trickreich:
Yeah.. I think this is the best solution here.

Ok, thanks. Have you already a plan when you are going to release it?

Because currently the log is full of errors in our applications.

PR #430, opened 2018-10-02
clear response tagger after tagging to not leak tags from one response to the next

dbu:
as discussed in https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/486

once this is merged, i will tag a patch release and update the bundle SymfonyResponseListener to also call `clear`

### DIFF
--- a/src/ResponseTagger.php
+++ b/src/ResponseTagger.php
@@ -21,6 +21,10 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
 /**
  * Service for Response cache tagging.
  *
+ * Record tags with this class and then either get the tags header or have the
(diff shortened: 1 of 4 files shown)


dbu:
@andrerom does this make sense?

src/ResponseTagger.php
```diff
@@ -143,10 +158,13 @@ public function tagResponse(ResponseInterface $response, $replace = false)
             return $response;
         }
 
+        $value = $this->getTagsHeaderValue();
```
andrerom: `$tagsHeaderValue` if we want to be explicit

dbu:
thank you, readability is important!

PR #436, opened 2018-10-30
Fixed HttpCacheProvider interface not allowing to return null

Toflar:
That's actually a bug in the interface documentation because just because a kernel is `HttpCacheAware` does not mean it really is able to return an HttpCache instance.
It's just aware of it meaning the kernel is able to receive one and return it back but still, if nobody ever gives it one, it cannot magically create one out of the blue :)

### DIFF
--- a/src/SymfonyCache/HttpCacheAware.php
+++ b/src/SymfonyCache/HttpCacheAware.php
@@ -19,7 +19,7 @@ use Symfony\Component\HttpKernel\HttpKernelInterface;
 trait HttpCacheAware
 {
     /**
-     * @var HttpKernelInterface
+     * @var HttpKernelInterface|null
      */
     private $httpCache;
 
--- a/src/SymfonyCache/HttpCacheAware.php
+++ b/src/SymfonyCache/HttpCacheAware.php
@@ -24,7 +24,7 @@ trait HttpCacheAware
     private $httpCache;
 
     /**
-     * @return HttpKernelInterface
+     * @return HttpKernelInterface|null
      */
     public function getHttpCache()
     {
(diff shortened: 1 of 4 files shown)


src/SymfonyCache/KernelDispatcher.php
```diff
@@ -96,6 +97,10 @@ public function flush()
 
         $httpCache = $this->httpCacheProvider->getHttpCache();
 
+        if (null === $httpCache) {
+            throw new ProxyUnreachableException('Cannot reach Symfony HttpCache as the Kernel does not know about it.');
```
dbu: could we add something more here to help with the most common setup error? "Kernel did not return a HttpCache instance. Did you forget $kernel->setHttpCache($cacheKernel) in your front controller?"

src/SymfonyCache/HttpCacheAware.php
```diff
@@ -24,7 +24,7 @@
     private $httpCache;
```
dbu: then this `@var` should also have `|null`

Toflar: Right, fixed!

src/SymfonyCache/KernelDispatcher.php
Toflar: Good hint, done :)

tests/Unit/SymfonyCache/KernelDispatcherTest.php
```diff
@@ -64,8 +52,43 @@ public function testFlush()
...
-        $dispatcher->invalidate($psr7Request);
+        $dispatcher->invalidate($this->getRequest());
 
         $dispatcher->flush();
     }
+
+    public function testFlushWithoutHttpCache()
+    {
+        $this->expectException(ProxyUnreachableException::class);
+        $this->expectExceptionMessage('Cannot reach Symfony HttpCache as the Kernel does not know about it.');
```
dbu: needs to be adjusted

Toflar:
Ah, lol. Idiot 😄 Ready then :)

dbu:
added a changlog entry and merged on the CLI. will tag 2.5.3 with this shortly.

thanks a lot!

Issue #435, opened 2018-10-30
Introduction of PurgeAllCapable?


Toflar:
Sometimes it would be really nice to just purge the whole reverse proxy cache.
Right now this doesn't seem possible because you always have to either purge by tags or URL
Sure, I could just tag every response with the `fos-cache-purge-all` tag and then purge this one but I don't think it's a good idea to just tag them all. I mean, I'm not sure if it is or not, hence this issue.

For the Symfony Proxy Client it could just purge the `http_cache` folder, so that would be easy to implement and definitely more efficient that tagging all the responses. However, I'm not sure about Varnish. Is there anything more efficient you think?
Seems like for Litespeed this would also be the recommended way to go: https://www.litespeedtech.com/support/forum/threads/resolved-ls-cache-purge-all-php-script.14921/

Wdyt?

dbu:
with varnish, you can ban without criteria. this is not super efficient, but its only one single purge request. usually i restart the varnish process, to reset everything.

i am not so sure if it would make sense that the web application itself can do a complete reset. it feels like there are few valid use cases. with the symfony cache kernel, people can build their own solution and with varnish they can ban everything or trigger a restart. therefore i feel like we should rather not have this in the library.

Toflar:
So maybe a bit more background: For Contao and basically any CMS that integrates the FOSCacheBundle and sticks to sending the correct response headers, the huge advantage is that you can configure any reverse proxy the bundle supports. So you can start off small with Symfony and later when traffic grows go bold and switch to Varnish. The CMS supports both of them because it uses the FOSHttpCacheBundle services to tag, invalidate etc. So imho it must be the same for eZ here.
And for any CMS one thing applies: Users manage content and not we as administrators do. So I have to give them some maintenance tasks they can execute when they create new content. One of which is to clear the whole reverse proxy cache.
What it currently does in Contao is it clears the `http_cache` folders content which means that maintenance job is useless if you use the Varnish configuration. So what it should do instead is asking the `CacheInvalidator` (or `CacheManager`) to just `::invalidateAll()` and in case of Symfony it would purge the folder and in case of Varnish it would send the respective request etc.

See where I'm coming from?

dbu:
okay, i see your point. and we can provide a generic way of doing this. 

[...]

PR #437, opened 2018-10-30
[WIP] ClearCapable

Toflar:
Quick hack to outline the concept as discussed in https://github.com/FriendsOfSymfony/FOSHttpCache/issues/435.


src/SymfonyCache/PurgeListener.php
```diff
@@ -13,17 +13,20 @@
...
  *
  * @author David Buchmann <mail@davidbu.ch>
+ * @author Yanick Witschi <yanick.witschi@terminal42.ch>
  *
  * {@inheritdoc}
  */
 class PurgeListener extends AccessControlledListener
 {
     const DEFAULT_PURGE_METHOD = 'PURGE';
+    const DEFAULT_PURGE_ALL_HEADER = 'Purge-All-Cache-Entries';
```
dbu: how about "CLEAR_CACHE" instead of PURGE_ALL?

src/ProxyClient/Invalidation/ClearCacheCapable.php
```diff
@@ -0,0 +1,28 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\ProxyClient\Invalidation;
+
+use FOS\HttpCache\ProxyClient\ProxyClient;
+
+/**
+ * An HTTP cache that supports clearing all of its
+ * cache completely.
```
dbu: s/its cache completely/its cache entries/

can we add a second paragraph? saying that this is here to allow a more efficient operation than banning everything, and might work with caches that do not support banning.

Toflar:
This is ready for review then 😄 I've added tests and docs. Imho the implementation in Varnish should be done in a separate PR (it's a new feature anyway) and I can't help much because I'm lacking the knowhow here.

src/CacheInvalidator.php
```diff
@@ -266,6 +274,24 @@ public function invalidateRegex($path, $contentType = null, $hosts = null)
...
+    /**
+     * Clear the cache completely.
+     *
+     * @throws UnsupportedProxyOperationException if HTTP cache does not support clearing the cache completely
+     *
+     * @return $this
+     */
+    public function clearCache()
+    {
+        if (!$this->cache instanceof ClearCacheCapable) {
```
dbu: imho this should fallback to BAN everything, if the cache is BanCapable

src/CacheInvalidator.php
```diff
@@ -105,6 +111,8 @@ public function supports($operation)
                 }
 
                 return $supports;
+            case self::CLEAR:
+                return $this->cache instanceof ClearCacheCapable;
```
dbu: if we implement fallback to ban, instanceof BanCapable would be ok too

src/SymfonyCache/PurgeListener.php
```diff
@@ -83,6 +96,24 @@ public function handlePurge(CacheEvent $event)
         $response = new Response();
         $store = $event->getKernel()->getStore();
 
+        // Purge whole cache
+        if ($request->headers->has($this->clearCacheHeader)) {
+            if (!$store instanceof Psr6StoreInterface) {
```
dbu: does the default cache store of symfony not allow pruning? if not, we should mention that in the documentation

src/ProxyClient/Invalidation/ClearCacheCapable.php
```diff
@@ -0,0 +1,34 @@
...
+ * An HTTP cache that supports clearing all of its
+ * cache entries.
+ *
+ * Proxy clients supporting this interface can allow
+ * for a more efficient delete-all operation rather
+ * than banning everything. Also it serves as
+ * an alternative to proxies that do not support
+ * banning cache entries at all.
+ */
+interface ClearCacheCapable extends ProxyClient
```
dbu: to be consistent with the other interfaces and our naming, how about `ClearCapable`?

src/CacheInvalidator.php
Toflar: But there is no general method like `banEveryting()` right? I mean, how am I supposed to know if the proxy supports "banning everything"? And if so, how? Does it use a special header with `*` or not?

src/SymfonyCache/PurgeListener.php
Toflar: Unfortunately, it does not. To be fair, there's very little reason not to use my store implementation 😄
How about https://github.com/FriendsOfSymfony/FOSHttpCache/pull/437/commits/aedf0771865a2df110780d447deadbb5a419757d?

src/ProxyClient/Invalidation/ClearCacheCapable.php
Toflar: Right, it's all about the `cache` anyway here so that makes no sense. I've renamed it. However, not sure if I should also rename the method from `clearCache()` to `clear()`?

dbu: lets rename the method on the proxy client to `clear` but keep the cache invalidator method `clearCache` i'd say.

src/CacheInvalidator.php
dbu: the way ban is defined, if you specify no criteria, it matches everything and thus bans everything.

Toflar: So `->ban()` with no headers speaking in code?

dbu: genau

src/ProxyClient/Invalidation/ClearCacheCapable.php
Toflar: Done :)

src/CacheInvalidator.php
Toflar: Tried to implement this but not sure it's a good idea to be honest. 
If we do this I have absolutely no possibility to find out if a proxy really supports clearing or if it just falls back to banning. It might cause issues because I'm never informed about the fact that I'm using a slow fallback cache clearing mechanism.

Imho if someone wants to do that, they should do this:

```php
try {
    $cacheInvalidator->clearCache();
} catch (UnsupportedProxyOperationException $e) {
    $cacheInvalidator->invalidate();
}
```

dbu: fair point indeed. instead of try-catch, they can also check the capability and call the right method.

Toflar:
Jep :)

src/ProxyClient/Symfony.php
```diff
@@ -94,4 +97,18 @@ public function invalidateTags(array $tags)
 
         return $this;
     }
+
+    /**
+     * Clear the cache completely.
+     *
+     * @return $this
+     */
+    public function clearCache()
```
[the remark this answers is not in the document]
Toflar: Was a bit in a hurry sorry. Should be all green now!

PR #439, opened 2018-10-31
do nothing on empty queue

### LINKED ISSUE #438: Version 2.5.3 breaks console application
The new `ProxyUnreachableException` exception which has been added in 7ca3f9e104d3bdc35b6fcdb3cdd57086929006ab yesterday breaks my console application. I think it is fine not to have an HTTP cache on the command line, therefore this case needs to be handled differently IMHO.

dbu:
fix #438


dbu:
@Toflar are you ok with this change? imho it makes sense

Toflar:
That makes perfect sense to me!

PR #444, opened 2018-11-21
LiteSpeed support

Toflar:
WIP for Open Source Litespeed (OLS) and possibly the commercial Litespeed Web Server (LSWS).

Cache tagging is only implemented in OLS but not at all in LSWS. The changes worked with OLS but the latest OLS version seems to have done BC breaks / regressions with cache tagging.

The original author does not continue on this - if you need this, you will need to wrap up this merge request or pay someone to do it...

---

This is a possible implementation for LSWS (https://www.litespeedtech.com/products/litespeed-web-server) support.


doc/litespeed-configuration.rst
```diff
@@ -0,0 +1,94 @@
...
+the proxy but instead we have to send a request to an endpoint where the response provides
+the correct LiteSpeed-specific headers which then trigger purging actions.
+You can read more on these headers in the `LiteSpeed response headers documentation`_.
+
+To do so, we generate a simple PHP file with a random file name containing the appropriate ``header()`` calls.
+After generation, we request this file and delete it again right away.
+
+For this reason you have to configure two parameters:
+
+* The location on your server where the file should be generated to (must be publicly accessible).
```
dbu: and writable by the server

doc/litespeed-configuration.rst
```diff
@@ -0,0 +1,94 @@
...
+* The base URL on which the generated file can be requested.
+
+Configuring LiteSpeed WebServer itself
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Enabling LiteSpeed to support public caching on your server is pretty much straight forward.
```
dbu: how straight forward this is depends on how familiar you are with stuff ;-)

it seems this is an apache module? so if you are currently on nginx, its not straightforward at all... also, some people love configuring apache in .htaccess while others only do it in the /etc files.

pet peeve aside: should we explain how to set up and configure LiteSpeed in the FOSHttpCache documentation? i feel we should not, it seems out of scope to me. we should only talk about specifics that are relevant for us (like: what about multi server setup? would you then setup LightSpeed on an apache that acts as proxy? or have a http cache on each node? (in which case you'd need to send the request to each node and have the file on each node)

doc/litespeed-configuration.rst
```diff
@@ -0,0 +1,94 @@
...
+* Your domain is called ``www.example.com``
+* Your domain points to ``/var/www/public``
+
+
+Your proxy client instance has to look like so::
+
+    use FOS\HttpCache\ProxyClient\HttpDispatcher;
+    use FOS\HttpCache\ProxyClient\LiteSpeed;
+
+    $servers = ['https://www.example.com'];
```
dbu: could we use an ip here to be more clear?

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,124 @@
...
+
+        return $resolver;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function invalidateTags(array $tags)
+    {
+        $this->headerLines[] = 'X-LiteSpeed-Purge: '.implode(', ', preg_filter('/^/', 'tag=', $tags));
```
dbu: if a lot goes wrong, could we end up with php server injection with a carefully crafted tag? i think we need some escaping here

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,124 @@
...
+
+        $url = '/'.$filename;
+
+        $this->queueRequest('GET', $url, []);
+
+        $result = parent::flush();
+
+        // Reset
+        $this->headerLines = [];
+        unlink($this->options['target_dir'].'/'.$filename);
```
dbu: imho createFile should return the absolute path so we can unlink whatever was returned.

doc/proxy-clients.rst
```diff
@@ -24,6 +24,7 @@ Client        Purge   Refresh Ban     Tagging Clear
 ============= ======= ======= ======= ======= =======
 Varnish       ✓       ✓       ✓       ✓
 NGINX         ✓       ✓
+LiteSpeed     ✓               ✓       ✓       ✓
```
dbu: the client does not implement the BanCapable interface - can we support BAN or not?

Toflar:
I think you got something about LiteSpeed wrong in general. LiteSpeed is a drop-in-replacement for Apache. It's not an Apache module but it supports `.htaccess` configuration. It's web server and proxy all in one and it's super fast (it's actually a pity many people don't know about it).
Do you want to review your comments again? It should answer a few things like `multi-node` setup questions and `is this an apache module?` 😄

dbu:
> I think you got something about LiteSpeed wrong in general. LiteSpeed is a drop-in-replacement for Apache.

maybe we should mention that in the lead then ;-)

multi-node setup is still very much a question though. apache can act as a reverse proxy. if you'd use litespeed that way, you would need to make sure the request to selve goes to the correct backend, or your generated header file is not there. if you use multiple litespeed behind a loadbalancer, you'd need to make sure that each of them gets the request (that they are all in the list of servers) so they see the invalidation resposne. and also that the file is generated on each instance.

doc/litespeed-configuration.rst
dbu: okay, got it. then lets say we need to enable caching in the LiteSpeed server and explain only that part.

lets not talk about ESI here as FOSHttpCache is not about ESI. (if you want to add a note with just a link to their doc, or a detailed blog post about LiteSpeed ESI & Symfony, that would be fine)

Toflar:
I think I'd rather mention that multi-node stuff is not a covered use-case at all. I won't work on support for that because I won't need it.

dbu:
> multi-node not covered

i think thats good. lets state that explicitly in the documentation, and mention that if you need more than one server with your app, varnish is a good solution for the caching reverse proxy, and its fully supported already.

litespeedtech:
Hi, dbu,

This is Lauren from LiteSpeedTech. We also have WebADC product, which follows the same headers. So this solution is generic, if it works with LiteSpeed Web Server, it will work with WebADC. WebADC is a full featured Load balancer with LSCache, ESI, HTTP/2, QUIC support built in. We already have many clients used it production environment, especially for big Magento stores. 

Thanks for great effort Toflar put in. We can together improve this. This is something we always wanted to do, but haven't got chance.

We have a Free Starter license (allow for 1 Domain, 2GB memory), this allows more devs can give a try in his local environment. Hopefully we can get more feedback.

Thanks,
Lauren

doc/litespeed-configuration.rst
Toflar: Right, will add this.

doc/litespeed-configuration.rst
Toflar: Right, will remove the part about ESI and Symfony.

doc/litespeed-configuration.rst
Toflar: Why would an IP be more clear? I don't understand this 😄

src/ProxyClient/LiteSpeed.php
Toflar: The invalidation request comes from the `CacheInvalidator`. This is trusted to me?

src/ProxyClient/LiteSpeed.php
Toflar: Not really, then I have to strip the whole relative path to the target_dir again when generating the public url.

doc/proxy-clients.rst
Toflar: I'll update the docs. Both Purge and Ban could be implemented by `PURGE` requests as for example in Varnish. Unfortunately this is not possible for tags which is why I've implemented this file-generation-thingy in the first place.

Toflar:
> the header file is a good solution, but seems against the philosophy of lightspeed. it should actually only be needed if we invalidate from a non-web context. when there is a web request that triggers the things that trigger invalidation, we would ideally set the headers on the web response.

It's the only method that is supported to invalidate cache tags and I actually find it a pretty nice solution. It's really easy to setup and it works across different applications or frameworks. Setting it on the response would again require a framework-specific implementation (e.g. a listener for Symfony).

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,124 @@
...
+        file_put_contents($this->options['target_dir'].'/'.$filename, $content);
+
+        return $filename;
+    }
+
+    private function generateUrlSafeRandomFileName()
+    {
+        $filename = 'fos_cache_litespeed_purger_';
+
+        if (function_exists('random_bytes')) {
```
Toflar: @dbu we're not talking crypto-stuff here and we're deleting the file right away so probably this part here is way to defensively coded 😄 Maybe a simple `sha1(mt_rand())` is perfectly good enough?

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function flush()
+    {
+        $filename = $this->createFile();
+
+        $url = '/'.$filename;
```
dbu: this is assuming we can put the files into the root directory. what if i need them in some sub directory?

lets rename target_dir to web_root to be clear what it is. right now, i'd expect /var/www/public/my_dir to work as target_dir, but it would end up with a 404

dbu:
hi @litespeedtech Lauren , thanks for looking at this. is WebADC the commercial licensed version of LiteSpeed? or is it a pure reverse proxy without the capability to run web applications itself, but follows the same instructions as LiteSpeed?

and what do you think of the discussion about hooking on the actual reply to the request for invalidation header, when possible?

Toflar:
> and what do you think of the discussion about hooking on the actual reply to the request for invalidation header, when possible?

It would be better of course. It's faster. But then we need to have access to the response somehow and this is something completely new we haven't used anywhere else before.

src/ProxyClient/LiteSpeed.php
Toflar: Why? If you configure `https://example.com/my_dir` as base, this would work perfectly fine?

src/ProxyClient/LiteSpeed.php
Toflar: Okay done in 8e627ab82650ba1b47260f617d9c210ba10c30c0.

litespeedtech:
@dbu WebADC is a different product than LSWS, it's a Layload balancer with LSCache. That's for our client who needs a cluster solution, they will need to use WebADC, as caching needs to be at the first layer.
Ideally if it's doable via direct response header, that will be great. That's our current cache plugin does. [Here](https://litespeedtech.com/products/cache-plugins/magento-acceleration/litemage2) we have a graph that showing the advantage over varnish. To serve same amount of traffic, our internal tests showing litespeed uses about 1/3 of CPU & memory of what varnish consumes.
Another advantage of LiteSpeed ESI implementation, is we supports esi:inline. If one page has many esi:include, we can use inline to do all with one response. This is used in our Prestashop plugin. But this is easier to do at cache plugin layer, not easy to do at framework level without much knowledge of the real app.

src/ProxyClient/LiteSpeed.php
dbu: thanks!

src/ProxyClient/LiteSpeed.php
dbu: ah, you are correct. can you do that in the example, specify the base url with a directory? and besides the example list the options and explicitly say how they interact? right now, the user has to be very familiar with FOSHttpCache or study the implementation to understand how this would work.

Toflar: Hmm, thinking about it, actually `$baseUri` never makes sense for LiteSpeed, am I right? We generate the file with the values we get. So if we get an `->purge(`/foobar`);` call, we generate `header('X-LiteSpeed-Purge: /foobar')` no matter the `$baseUri`. Wdyt?

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,110 @@
...
+ */
+
+namespace FOS\HttpCache\ProxyClient;
+
+use FOS\HttpCache\ProxyClient\Invalidation\ClearCapable;
+use FOS\HttpCache\ProxyClient\Invalidation\PurgeCapable;
+use FOS\HttpCache\ProxyClient\Invalidation\TagCapable;
+
+/**
+ * LiteSpeed Web Server (LSWS) invalidator.
```
dbu: should we mention the WebDAC too here, as it should work for both? also in the doc?

Toflar: I have no clue about it, never heard, never tried. So I won't be able to contribute.

src/ProxyClient/LiteSpeed.php
dbu: hm. we need to use the right base path in what we tell litespeed to purge. if the symfony app is installed at /foo and you purge /bar the web url is /foo/bar, right?

Toflar: I've checked with the LSWS support. So when you call e.g. `https://foobar.com/about-us` LS will of course cache this at `foobar.com/about-us`. If you want to purge this, you have to set the header `X-LiteSpeed-Purge: /about-us` on `foobar.com`. LSWS does not allow to purge cross-domain.
So `$baseUri` actually **must** be empty and `$servers` might be multiple ones if your site runs on multiple domains (which shouldn't be done anyway because of duplicate content).
We could ensure we only have the path in `public function purge($url...)`. So we might want to make absolute URLs relative ones automatically and document the part about having multiple servers and always leaving `$baseUri` empty. Do you understand what I mean? :)

dbu: thats why i asked for $servers to be set to IPs to make the example clearer ;-)

so $baseUri is how your server is being talked to, including `foobar.com`. $servers is the list of IPs where the caching proxy runs. imho $baseUri must be the domain (plus path, if symfony does not run at `/`) and $server is the IP or DNS name of the machine running LSWS. 

multiple domain names for the same site are not supported with Varnish either. i don't think we want to support that, because it is extremly bad practice if you care about caching.

i think the target_dir, base_uri and server are all we need to specify things. we should explain how they work together with an example where symfony does not run at `/` of the domain, to avoid confusion there.

Toflar: > i think the target_dir, base_uri and server are all we need to specify things. we should explain how they work together with an example where symfony does not run at `/` of the domain, to avoid confusion there.

I'm trying but I don't get it myself. Imho `$baseUri` is prepended automatically to all relative `->purge($url)` calls. Is that not correct?

dbu: it is prepended, yes. when the requests are actually sent, we use the $servers array to update the request: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/a2fce472f38defcdb444d6c3c14257ff79c66392/src/ProxyClient/HttpDispatcher.php#L225-L233

the crucial line is https://github.com/FriendsOfSymfony/FOSHttpCache/blob/a2fce472f38defcdb444d6c3c14257ff79c66392/src/ProxyClient/HttpDispatcher.php#L231

in the end, HTTP uses a TCP connection to an IP and the payload contains a `Host: ` header. that host is what baseUri is about, and what the caching proxy looks at. the domain name in `$servers` is used to resolve the IP address (if its not directly an IP address already).

Toflar: Yeah, and we never prepend. Purging must be relative in our case (we can ensure this) and then servers must be the correct domains imho.

dbu: imho we should follow the same logic as with the other caching proxies.

if symfony is not running at `/` we need to have the web path both in the path to the generated php file and prepend in the purge instructions inside that generated php file. i would use the base_uri for that path information. if we want to allow the generated php file to be created in a subpath, we would then need a separate parameter for that i think.

dbu: ah sorry, how do you mean purging is relative? relative to the web path where symfony has its root?

Toflar: > I've checked with the LSWS support. So when you call e.g. https://foobar.com/about-us LS will of course cache this at foobar.com/about-us. If you want to purge this, you have to set the header X-LiteSpeed-Purge: /about-us on foobar.com. LSWS does not allow to purge cross-domain.

So `$purgeUri` is nonsense for LSWS. You cannot have `X-LiteSpeed-Purge: foobar.com/about-us`. You can only have it on the domain you request the generated purger file from.

dbu: this is the same when purging from varnish. we create the request with the correct host name and the correct path: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/b9d4bef48a678ee8b1b86391e9883c71e14d27b8/src/ProxyClient/HttpProxyClient.php#L93

i think to have the same feature scope, we should handle it when the $uri passed to `purge` contains the domain. so we would collect purge instructions per domain and on flush create those files and send the purge requests to $servers with the correct Host header for each domain.

Toflar: That doesn't work. I mean, for the urls, fine. But then to which file would you add the tags that need to be purged? All of them? Again, I don't think it works in a multi node setup so it never makes sense to provide multiple servers and so `$baseUri` is never needed.

dbu: do you come to symfony con next week? i think it would be easier to discuss this face-to-face ;-)

Toflar: I'll be there by Wednesday at 3pm or so :)

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,110 @@
...
+     */
+    public function flush()
+    {
+        $filename = $this->createFile();
+
+        $url = '/'.$filename;
+
+        $this->queueRequest('GET', $url, []);
+
+        $result = parent::flush();
```
dbu: flush might throw an ExceptionCollection. deletion should be in a finally block

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,110 @@
...
+        $content = '<?php'."\n\n";
+
+        foreach ($this->headerLines as $header) {
+            $content .= sprintf('header(\'%s\');', addslashes($header))."\n";
+        }
+
+        // Generate a reasonably random file name, no need to be cryptographically safe here
+        $filename = 'fos_cache_litespeed_purger_'.substr(sha1(uniqid('', true).mt_rand()), 0, mt_rand(10, 40)) . '.php';
+
+        file_put_contents($this->options['target_dir'].'/'.$filename, $content);
```
dbu: document_root (public web directory of this application) and optional target_dir (defaults to /)

Toflar:
Failing test is unrelated, maybe you need to retrigger the build. I did not update the docs yet as I want to have your feedback on the implementation first 😄

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+class LiteSpeed extends HttpProxyClient implements PurgeCapable, TagCapable, ClearCapable
+{
+    private $headerLines = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clear()
+    {
+        $this->addHeaderLine('X-LiteSpeed-Purge', '*');
```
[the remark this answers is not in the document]
Toflar: Added a comment.

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function purge($url, array $headers = [])
+    {
+        $urlParts = parse_url($url);
+        $host = null;
+
+        if (isset($urlParts['host'])) {
```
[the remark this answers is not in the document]
Toflar: Right! Done.

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+            $host = $urlParts['host'];
+            $url = $urlParts['path'];
```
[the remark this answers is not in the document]
Toflar: Done.

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        $this->addHeaderLine('X-LiteSpeed-Purge', $url, $host);
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
```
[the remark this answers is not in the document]
Toflar: Added.

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+     */
+    public function flush()
+    {
+        $filenames = [];
+
+        $url = '/';
```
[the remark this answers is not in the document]
Toflar: Done.

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+
+        if ($this->options['target_dir']) {
+            $url .= $this->options['target_dir'].'/';
+        }
+
+        foreach ($this->headerLines as $host => $lines) {
+            $filename = $this->createFileForHost($host);
```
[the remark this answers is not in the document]
Toflar: Adjusted.

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+
+            $this->queueRequest('GET', $url.$filename, []);
```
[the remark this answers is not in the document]
Toflar: But `$host` can be `/`?

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,167 @@
...
+            foreach ($filenames as $filename) {
+                unlink($this->getFilePath().'/'.$filename);
+            }
+        }
+    }
+
+    private function addHeaderLine($header, $value, $host = null)
+    {
+        if (null === $host) {
+            $host = $this->options['base_uri'];
```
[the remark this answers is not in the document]
Toflar: Can you elaborate what you mean or maybe even commit yourself directly (pushes are allowed).

tests/Unit/ProxyClient/LiteSpeedTest.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * @var string
+     */
+    private $documentRoot;
+
+    protected function setUp()
+    {
+        $this->httpDispatcher = \Mockery::mock(HttpDispatcher::class);
+
+        $deleteFilesAndFolders = function ($path) use (&$deleteFilesAndFolders) {
```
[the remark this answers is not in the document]
Toflar: Same thoughts :) I would've used it if it was a depencency already. Maybe if we need more FS related stuff we might want to require it as a dev-dependency.

Toflar:
I fail to install openlitespeed on travis really. I don't see how I can contribute functional tests here.

Toflar:
Sooo I have some pretty good news to share here! I managed to get in touch with the super helpful folks at LiteSpeed and they helped me out with all sorts of questions. They listened to me and will now integrate purging tags by a PURGE request which is super awesome! So I‘ll revamp the integration pretty soon as I think it will get a lot closer then to what we have already for e.g. Varnish🙌🙌🙌

dbu:
very cool! sorry, i never got around to look into this. did you manage to run litespeed on travis-ci meanwhile, or is that still an issue?

Toflar:
> or is that still an issue?

That is still an issue. It has to run via `sudo` which is one thing I got fixed. So that's okay but I just can't figure out how to make it point to the correct root dir using `$TRAVIS_BUILD_DIR`. My bash foo is not good enough and my `sed` stuff in `install-openlitespeed.sh` is not working.

Toflar:
I don't know man. That `sed` command is just not working. I've tried 100 variants. Works just fine on Ubuntu but I can't get it to work on Travis.

Toflar:
Ah stupid me. Fixed 😄 Okay, now OLS is running just fine. I have to figure out why I get 404 responses but I guess I'll be able to do so at some point 😄

Toflar:
Yeah! Finally got OLS running on Travis. We are now able to run the functional tests against it but of course, I need to wait for GA of the new `PURGE` endpoint.

Toflar:
So @dbu I was informed that the `PURGE` endpoint will also be vhost specific.
So imagine you have `domain.org` and `domain.com` and both cache a `/foobar` path (or a set a `foobar-1` cache tag). To purge the paths relatively on both domains, we need to send 2 `PURGE` requests to each of the two domains. Same applies for cache tags.
This is currently not possible with our `HttpDispatcher` because it does take multiple servers but only one baseUri:

```php
$httpDispatcher = new HttpDispatcher(
    ['http://127.0.0.1'],
    'localhost'
);
```

In fact what we need here would be this:

```php
$httpDispatcher = new HttpDispatcher(
    ['http://127.0.0.1'],
    ['domain.org', 'domain.com']
);
```

And then during the `HttpDispatcher::fanOut()` we have to create a request per baseUri. Should be fully BC. I can just make the string argument an array with one baseUri.
Any objections?

Toflar:
> Any objections?

I've implemented it now so you can see the code directly.

doc/litespeed-configuration.rst
```diff
@@ -0,0 +1,137 @@
...
+      checkPublicCache                 1
+
+      # This disables the private cache
+      checkPrivateCache                0
+
+      # Also consider the query string in caches
+      qsCache                          1
+
+      # Enable checking for a cached entry if there's a cookie on the request
+      reqCookieCache                   1
```
dbu: should we add a line here to mention that you need to enable user context caching if you set this to 1 and otherwise risk mixing up your cache? or maybe we should default 0 and then have a commented line with 1 and explain to use that with user context caching?

doc/litespeed-configuration.rst
```diff
@@ -0,0 +1,137 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+LSWS can also be configured on server level, however, as LSWS supports ``.htaccess`` you may want to configure it
+there to give your application the flexibility of having different configurations for certain paths.
+
+Configure your ``.htaccess`` as follows::
+
+    <IfModule LiteSpeed>
+        CacheEnable public /
+        # TODO: The rest of the directives
```
dbu: are the directives named the same as with OLS? then i'd just put a comment that says "Look at the recommended OLS parameters."

doc/litespeed-configuration.rst
```diff
@@ -0,0 +1,137 @@
...
+    </IfModule>
+
+Also refer to the LSWS docs if you need more details about the different configuration values.
+
+Configuring the library
```
dbu: i'd say "Configuring FOSHttpCache to work with LiteSpeed" - library is a bit vague for the context switch

src/ProxyClient/HttpDispatcher.php
```diff
@@ -78,19 +78,20 @@ class HttpDispatcher implements Dispatcher
...
      *                                         E.g. ['127.0.0.1:6081']
-     * @param string               $baseUri    Default application hostname, optionally
+     * @param string|string[]      $baseUris   Default application hostnames, optionally
      *                                         including base URL, for purge and refresh
-     *                                         requests (optional). This is required if
+     *                                         requests (optional). At least one is required if
      *                                         you purge and refresh paths instead of
-     *                                         absolute URLs
+     *                                         absolute URLs. A request will be sent for each
+     *                                         base URL.     *
```
dbu: please remove the trailing `*`

src/ProxyClient/HttpDispatcher.php
```diff
@@ -104,15 +105,19 @@ public function __construct(
         $this->uriFactory = $uriFactory ?: UriFactoryDiscovery::find();
 
         $this->setServers($servers);
-        $this->setBaseUri($baseUri);
+
+        // Support both, a string or an array of strings (array_filter to kill empty base URLs)
```
dbu: i am not happy with the array_filter, as that could hide problems. i'd explicitly check `'' === $baseUri` and in that case create [].

src/ProxyClient/HttpDispatcher.php
```diff
@@ -187,52 +192,65 @@ protected function getServers()
...
-                }
-            } else {
-                // Relative URI
-                if ('' !== $this->baseUri->getHost()) {
-                    $uri = $uri->withHost($this->baseUri->getHost());
-                }
-
-                if ($this->baseUri->getPort()) {
-                    $uri = $uri->withPort($this->baseUri->getPort());
+        /** @var RequestInterface[] $requests */
```
dbu: we should not need this comment, we only put request interface into this variable and code checkers should see that i think.

src/ProxyClient/HttpDispatcher.php
```diff
@@ -187,52 +192,65 @@ protected function getServers()
...
+        $requests = [$request];
```
dbu: we do not use that in the case of having baseUris. i would prefer to structure the code like this:
```php
if (0 !== \count) {
    $requests = $this->requestToBaseUris($request);
} else {
    $requests = [$request];
}
```

src/ProxyClient/HttpDispatcher.php
```diff
@@ -257,19 +275,21 @@ private function setServers(array $servers)
...
-            $this->baseUri = null;
+        if (0 === \count($baseUris)) {
+            $this->baseUris = [];
 
             return;
         }
 
-        $this->baseUri = $this->filterUri($uriString);
+        foreach ($baseUris as $baseUri) {
+            $this->baseUris[] =  $this->filterUri($baseUri);
```
dbu: does this detect when $baseUri is empty? i think its good if we produce a clear error message in that case.

src/ProxyClient/LiteSpeed.php
```diff
@@ -0,0 +1,82 @@
...
+        $this->sendPurgeRequest([
+            'X-LiteSpeed-Purge' => implode(', ', preg_filter('/^/', 'tag=', $tags)),
+        ]);
+
+        return $this;
+    }
+
+    private function sendPurgeRequest(array $headers)
+    {
+        // TODO: LiteSpeed is likely going to hard code this URL, otherwise it has to be configurable
```
dbu: this says "TODO" - should we merge it like this for now? or can we link to some issue tracker of litespeed?

src/Test/LiteSpeedTest.php
```diff
@@ -0,0 +1,119 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\Test;
+
+use FOS\HttpCache\ProxyClient\HttpDispatcher;
+use FOS\HttpCache\ProxyClient\LiteSpeed;
+use FOS\HttpCache\Test\Proxy\LiteSpeedProxy;
+
+// TODO: update docs in case anything is configurable
```
dbu: another TODO

.travis.yml
```diff
@@ -56,7 +56,6 @@ install:
 before_script:
   # Install Varnish
   - |
-    set -e
```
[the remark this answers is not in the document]
Toflar: I removed it because travis will not execute `after_script` and `after_failure` if you do this :)

doc/litespeed-configuration.rst
Toflar: I would prefer not to mention it because I haven't tested if that would really work.

doc/litespeed-configuration.rst
Toflar: I don't know yet. To be tested.

doc/litespeed-configuration.rst
Toflar: Ok will do.

src/ProxyClient/HttpDispatcher.php
Toflar: Oh that was a leftover, will do.

src/ProxyClient/HttpDispatcher.php
Toflar: Not phpstorm.

src/ProxyClient/HttpDispatcher.php
Toflar: It cannot be empty. Empty baseUri's were ignored silently up until now and they still are.

src/ProxyClient/LiteSpeed.php
Toflar: Nooo, this is far from being done! It has to wait until LiteSpeed releases the new features so I can actually integrate it. All we have in this PR is just preparation work :)

src/ProxyClient/HttpDispatcher.php
Toflar: Done.

src/ProxyClient/HttpDispatcher.php
Toflar: Done in 5f160db1d7059137fabb1a709e9b6fa36c5939c1. Do you like that better?

.travis.yml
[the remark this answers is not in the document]
dbu: i see. its on a CI, so worst case you have to read a bit more when it fails to figure out what the first error was.

doc/litespeed-configuration.rst
dbu: without user context, we should really not enable caching when there are cookies, otherwise you get cache mixups when there are session cookies

src/ProxyClient/HttpDispatcher.php
dbu: yep. added another nitpick, but this looks more robust

doc/litespeed-configuration.rst
Toflar: We're not caching. We're serving entries from the cache even if you're logged in. Depends on your use case but if handled correctly, that's a valid use case. E.g. all your public pages can still be served from the cache even if one of your visitors is logged in. However, you have to make sure that these "public" pages are really public or if they contain parts that differ when one is logged in, they are served via ESI. 
So really it depends on the use case and I agree, you have to be pretty fit in terms of caching to understand it all. But on the other hand, we cannot write an essay here. Do you think we should disable it in the docs and just mention that one might want to enable this depending on the use case?

tests/Unit/ProxyClient/HttpDispatcherTest.php
```diff
@@ -171,6 +171,8 @@ public function testBanWithoutBaseUri()
 
         $request = $this->messageFactory->createRequest('BAN', '/', ['X-Url' => '/foo/.*']);
         $httpDispatcher->invalidate($request, false);
+
+        // TODO: This is already a risky case because it doesn't test anything. Should be fixed in master?
```
Toflar: @dbu I added https://github.com/FriendsOfSymfony/FOSHttpCache/issues/447 and will update this PR and remove this TODO because it's not related to this PR at all.

doc/litespeed-configuration.rst
dbu: so the use case would be that you have some pages that do not depend on the session in at all, and you configure the cache to allow caching of everything and the backend is supposed to cache-control with that in mind? 

if i got that correctly, none of our other documentations suggest this strategy. i would not put that into such an example with a list of configuration options that people might blindly copy-paste and end up with a very risky system if they are not aware what they did.

if we want to cover this use case in the doc, it should be a new section, explaining for each cache how to configure the behaviour and clearly warning of the consequences if the backend does not set cache instructions correctly.

Toflar: No, I think you still didn't get the use case right 😄
Let's say we have an `/about-us.html` which is public. The page has a private section where you can log in so when you visit `/about-us.html` you potentially have a cookie set.
So in this case I don't want OLS to ignore the cache for `/about-us.html` because I know it's not going to be different.
But yes, I agree you have to know every response of your app and you have to always keep that in mind. So I'm probably going to change this for the docs, I can still enable it for myself 😄

dbu: hm, how is that different from what i described? you are (maybe) logged in and go to a page and we tell the cache that if the backend wants that page cached, it should cache it even if there are credentials.

as said, it would be ok to cover that use case in the docs (as kindof a simplified version of user context that does not need the hash request). but we should not have it here as a "default" configuration, its too dangerous for that.

Toflar: Agreed.

doc/litespeed-configuration.rst
```diff
@@ -0,0 +1,156 @@
...
+      checkPublicCache                 1
+
+      # This disables the private cache
+      checkPrivateCache                0
+
+      # Also consider the query string in caches
+      qsCache                          1
+
+      # Disable checking for a cached entry if there's a cookie on the request
+      reqCookieCache                   0
```
dbu: we don't support user context caching with litespeed, right?

dbu:
i notice that this seems mostly done. any chance we can wrap it up and merge it?

Toflar:
It just works with OLS. Apparently there's no interest to integrate the same logic in LSWS which makes absolutely no sense to me. Looks like LiteSpeed has two different products which do not share any of the code base, not even the functional tests :(

dbu:
strange. well, lets wrap it up for OLS and add a note that LSWS behaves different and would need more work?

Toflar:
Yeah, I'll add the docs and write that LSWS currently does not support tag invalidation.

doc/litespeed-configuration.rst
Toflar: Right. User context caching requires some kind of preflight request to the app to find out the context which requires implementation in the proxy.

Toflar:
I really doubt that I want to finish this PR at this point.

First of all, I set out to do this PR because I had customers running behind LSWS and as there seems to be no effort at all in implementing cache tag purging there, there's no benefit for me or my customers. So if I was to finish this, it would be pure goodwill because I cannot use it myself anyway.

And looking at the failing unit tests after updating to the latest stable OLS release, there are random failures again. So I feel like I am the one that is functional testing OLS and that's definitely not the purpose of this PR. I mean I could of course dig into why exactly OLS doesn't behave like it should anymore but then again, I don't use it myself...why would I?

dbu:
pitty, but fair enough. totally understand you. 

i suggest we keep this around for a while still, in case somebody wants to continue working on it. i updated the description and marked the PR as needing a new contributor.

dbu:
i close this as it is pretty old.

if somebody wants to pick things up, please feel free to do so, i am not against having a proxy client for litespeed in here. note that with all the refactorings that happened since this was started, the proxy client might has to look different than in this PR.

PR #451, opened 2019-05-06
Add Fastly support

andrerom:
Closes #403

Changes:
- Use batch tag purging to reduce risk of hitting API limits
- Support PurgeCapable
- Drop multi services, does not seems to be a good fit for Purge, & unsure what the use case is
- Added unit testing

Follow Up topics:
- Add support in Bundle
- Abstract Reverse proxy TTL to use Surrogate Control on Fastly and optionally others  #455 

Context of this:
- eZ Platform (v3) is being worked on by @ViniTou to remove several of our HttpCache abstractions and move to FosHttpCache 2.x, which means we among other things we'd like Fastly support on FosHttpCache level instead ;)


dbu:
thanks for picking this one up again. great if we can get this in!

while having some demo or something to validate that it works with the actual api, i agree that the automated tests should not talk with actual fastly. could you use the MockClient of httplug to expect requests and return canned responses so we can do a functional test within our application boundary?

src/ProxyClient/Fastly.php
```diff
@@ -0,0 +1,130 @@
...
+        if (true === $this->options['soft_purge']) {
+            $headers['Fastly-Soft-Purge'] = 1;
+        }
+
+        foreach (\array_chunk($tags, 256) as $tagchunk) {
+            $this->queueRequest(
+                Request::METHOD_POST,
+                sprintf("/service/%s/purge", $this->options['service_identifier']),
+                $headers + [
+                    'Surrogate-Key' => implode(' ', $tagchunk),
```
vvuksan: You may want to change the surrogate keys to JSON payload instead of sending them as a header. 

https://github.com/fastly/fastly-magento2/blob/master/Model/Api.php#L197

andrerom: Due to escaping? It will still be limited to 256 tags right?

vvuksan: It's still limited to 256. If payload is a problem you can just send it as a header. It was just a little bit cleaner.

dbu: as long as we are BC (which we should be able to) i think it makes sense to add an optional body to `queueRequest`. parameter ordering will be a bit "funny" because of the validateHost flag we already have, but so be it...

src/Test/Proxy/FastlyProxy.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\Test\Proxy;
+
+class FastlyProxy extends AbstractProxy
```
dbu: is this WIP?

andrerom: Yes, I'm unsure what we want to do on testing.

@vvuksan sent me tips on maybe using this: https://fiddle.fastlydemo.net/
But I'm unsure if it is suitable for automated functional testing.
So I'm wondering if it's best to just aim for unit testing for now.

@vvuksan Is that what you had in mind as well? Or do you have suggestions on how we can also use fiddle for automated functional tests (CI/Travis)?

Toflar: I have the same issue for my LiteSpeed PR (which I should polish, I know...). OpenLiteSpeed is no issue because you can install it during CI but the enterprise version (which happens to not be the same code base somehow) cannot be installed because one needs a license. I've already talked to them asking for some kind of OSS license but no luck so far. I guess we'll have certain integrations without functional tests then :(

andrerom: If you need ESI, then [indeed it seems](https://www.litespeedtech.com/products/litespeed-web-server/editions) you won't be able to run the tests on the open source version.

Toflar: We don't need ESI for our tests but the thing is, you just cannot install the Enterprise version due to the license. And there's no guarantee things work because they do in the OSS variant as they share no code. It might sound weird (at least to me it does) but the implementations are in fact completely different :) Anyway, not the discussion I wanted to start here. I just wanted to show another example of where functional tests on CI are probably not going to happen.

andrerom: yes, my point was that the code difference is not our issue, it's theirs. So if they behave different that is a bug on their side (imho). Aka, unless you rely on enterprise features we can "safely" use oss version for functional.

Toflar: Ah, I agree yes :)

dbu:
I suggest that we do some functional tests with the httplug mock client
and/or the vcr-plugin. That can at least ensure we know what we send out,
and "documents" what we expect as interaction with fastly.
With php-vcr it would be doable to switch tests to talk to an actual fastly
account with credentials and check if it really works.

andrerom:
> I suggest that we do some functional tests with the httplug mock client
and/or the vcr-plugin. 

Any existing FOSHttpCache examples of that?

dbu:
no such examples, so far we have only unit tests and full integration tests (we install nginx and varnish in travis-ci). 
you can set the vcr plugin to record mode and run it with an actual fastly, to have it generate the fixtures. maybe we can use a constant that can be set to activate passthrough mode and configuration  for a fastly account so that phpunit.xml can be used to locally run tests against an actual fastly instance, while using the fixtures in the ci.

andrerom:
@dbu / @Toflar Unit tests added.

I'm considering adding refresh using two requests, one to soft purge the url and then second one to load it. WDYT?

Toflar:
> Only thing I'm considering from my POV is adding `refresh()` using two requests, one to soft purge the url and then second one to load it. WDYT?

I did that in my LiteSpeed PR as well so if it works, why not? 😄

dbu:
sure, if we can't do it in one then that sounds fine. with varnish / symfony you can do a request forcing a backend lookup, but when that is not available, doing 2 requests seems fine to me. needs to be sure there is no race condition though (doing the request before the cache got purged)

andrerom:
Added.

@vvuksan Would you be able to verify if what's added here* successfully will trigger a refresh of a url? Or if you would recommend another way? e.g undocumented API or something ;)

\* TL;DR;
1. soft PURGE of url
2. HEAD request for url

andrerom:
@dbu / @Toflar Ready for review afaik.

dbu:
build errors should go away once https://github.com/symfony/symfony/pull/32802/ is tagged

src/ProxyClient/Fastly.php
```diff
@@ -0,0 +1,158 @@
...
+        $headers = [
+            'Fastly-Key' => $this->options['authentication_token'],
+            'Accept' => 'application/json',
+        ];
+
+        if (true === $this->options['soft_purge']) {
+            $headers['Fastly-Soft-Purge'] = 1;
+        }
+
+        foreach (\array_chunk($tags, 256) as $tagChunk) {
```
dbu: how do we know this number? i think in other clients, we did this more sophisticated with the actual header length. i'd assume we need to do the same here, to avoid over long headers when purging lots of tags with long names.

src/ProxyClient/Fastly.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureOptions()
+    {
+        $resolver = parent::configureOptions();
+
+        $resolver->setRequired([
+            'authentication_token',
+            'service_identifier',
```
dbu: i think we usually documented additional configuration options in phpdoc - could you check with the varnish client and document in a similar place?

andrerom: It’s in their docs as hard limit, I can add a link.

dbu: for the varnish proxy client, we have this section in the phpdoc: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/01d87faf5cc48e2164b958b0a64ed7ff2b8b71ea/src/ProxyClient/Varnish.php#L24, i'd like to do the same here. can be the same text as in the doc.

dbu:
oh, and link to fastly docs is of course a good idea!

can you please also add an entry in the changelog and add info in the documentation? https://github.com/FriendsOfSymfony/FOSHttpCache/tree/master/doc

i'd squash all commits, even though that will mash together the git attribution to just one of you. hope you don't mind? otherwise i squash to 2 commits to keep your things separate.

@Toflar do you see anything else or should i merge when we have the documentation? i plan a new minor version once this is released.

tests/Unit/ProxyClient/FastlyTest.php
```diff
@@ -0,0 +1,177 @@
...
+
+                    return true;
+                }
+            ), false
+        );
+
+        $fastly->invalidateTags(['post-1', 'post,type-3']);
+    }
+
+    public function testInvalidateTags_HeadersSplit()
```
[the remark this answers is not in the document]
dbu: i would veto that in src/ but for tests/ it seems ok to me. i think i mostly have seen things like testInvalidateTagsHeaderSplit, but with `_` it is actually much more readable imho. i am fine keeping it like this.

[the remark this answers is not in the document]
andrerom: I usually don't do this in eZ code, so unsure why I did it here.
Could have been in the moment thinking to keep the root of the tests the same, and variants of the tests separated by `_`

I'm fine either way, but I prefer to stay consistent with the rest of the code base convention if any.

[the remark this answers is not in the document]
Toflar: We use camel case everywhere. This mixture of snake and camel case looks strange to me but then again, who cares 😄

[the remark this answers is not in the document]
andrerom: I agree, I'll drop it in next update

src/ProxyClient/Fastly.php
andrerom: It’s in their docs as hard limit, I'll add a link.

dbu: but if each tag would be 40 characters (because i use bad tags, or uuid or something), we would end up with a 10k header line, no? that is likely to break, depending on intermediary systems.

andrerom: RTFM ;)
Added link now

doc/proxy-clients.rst
```diff
@@ -158,6 +159,44 @@ default ``xkey-softpurge``.
...
 
+Fastly Client
+~~~~~~~~~~~~~~
+
+The Fastly client sends HTTP requests with the ``HttpDispatcher``. Create the
+dispatcher as explained above and pass it to the Fastly client::
+
+    use FOS\HttpCache\ProxyClient\Fastly;
+
+    $varnish = new Fastly($httpDispatcher);
```
dbu: $fastly?

andrerom:
Fixed a few things, notable:
- If you want doc / text to be different, please use GH suggest feature ;)
- Realized that I was missing API endpoint for invalidation() and clear() => *So unless unclear, there has been no functional tests here yet, that is a must before release. Contribution on that would be great, otherwise I'll need to check internally who might be able to do that.*

doc/proxy-clients.rst
```diff
@@ -158,6 +159,44 @@ default ``xkey-softpurge``.
...
+.. note::
+
+    Unlike other supported proxies there is no configuration needed for the proxy itself as all invalidation is done
+    against `Fastly Purge API`_. But for optimal use make sure to tune configuration together with Fastly.
+
+You need to pass the following options to the Fastly client:
+
+* ``service_identifier``: Identifier for your Fastly service account.
+* ``authentication_token``: User token for authentication against Fastly APIs.
+* NB: To be able to clear all cache(`->clear()`), you'll need a token for user with Fastly "Engineer permissions".
```
dbu: very cool! this is exactly what such documentation should help people with, remove wtf moments :+1:

doc/proxy-clients.rst
```diff
@@ -158,6 +159,44 @@ default ``xkey-softpurge``.
...
+A full example could look like this::
+
+    $options = [
+        'service_identifier' => '<my-app-identifier>',
+        'authentication_token' => '<user-authentication-token>',
+        'soft_purge' => false
+    ];
+    $requestFactory = new MyRequestFactory();
+
+    $varnish = new Fastly($httpDispatcher, $options, $requestFactory);
```
dbu: $fastly

src/ProxyClient/Fastly.php
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        // Split tag invalidations across several requests within Fastly's tag batch invalidations limits.
+        foreach (\array_chunk($tags, self::TAG_BATCH_PURGE_LIMIT) as $tagChunk) {
+            $this->queueRequest(
+                Request::METHOD_POST,
+                sprintf(self::API_ENDPOINT . '/service/%s/purge', $this->options['service_identifier']),
+                $headers + [
+                    // Note: Can be changed to rather use json payload if queueRequest is changed to allow passing body
+                    'Surrogate-Key' => implode(' ', $tagChunk),
```
dbu: with long tags, this header could grow longer than some intermediate http handlers support.

i would adapt this to use https://github.com/FriendsOfSymfony/FOSHttpCache/blob/01d87faf5cc48e2164b958b0a64ed7ff2b8b71ea/src/ProxyClient/Varnish.php#L82 (but never more than the TAG_BATCH_PURGE_LIMIT)

src/ProxyClient/Fastly.php
```diff
@@ -0,0 +1,180 @@
...
+            $headers['Fastly-Soft-Purge'] = 1;
+        }
+
+        // Split tag invalidations across several requests within Fastly's tag batch invalidations limits.
+        foreach (\array_chunk($tags, self::TAG_BATCH_PURGE_LIMIT) as $tagChunk) {
+            $this->queueRequest(
+                Request::METHOD_POST,
+                sprintf(self::API_ENDPOINT . '/service/%s/purge', $this->options['service_identifier']),
+                $headers + [
+                    // Note: Can be changed to rather use json payload if queueRequest is changed to allow passing body
```
dbu: did we discuss that? i think we could add an optional body argument in HttpProxyClient::queueRequest, or not?

would be less fragile than the header and we could simply use the TAG_BATCH_PURGE_LIMIT constant without further worries.

src/ProxyClient/Fastly.php
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * {@inheritdoc}
+     *
+     * @see https://docs.fastly.com/api/purge#purge_db35b293f8a724717fcf25628d713583
+     */
+    public function invalidateTags(array $tags)
+    {
+        $headers = [
+            'Fastly-Key' => $this->options['authentication_token'],
```
dbu: as every request needs to be authenticated, we could overwrite the `queueRequest` method and add this header there instead of in all methods.

doc/proxy-clients.rst
andrerom:
```suggestion
    $fastly = new Fastly($httpDispatcher, $options, $requestFactory);
```

doc/proxy-clients.rst
andrerom:
```suggestion
    $fastly = new Fastly($httpDispatcher);
```

dbu:
> Please use GH suggest feature so we can speed up here ;)

oh, will try that next time, seems nice.

> Realized that I was missing API endpoint for invalidation() and clear() => So unless unclear: There has been no functional tests here yet, that is a must before release. Contribution on that would be great, otherwise I'll need to check internally who might be able to do that.

i never used fastly. maybe @hpatoio wants to try to do that?

hpatoio:
> i never used fastly. maybe @hpatoio wants to try to do that?
Ciao. Cool you finalized my PR. We are not using this bundle anymore (went for a different solution) but I'm happy to help. What do you exactly need ? Sorry but the conversation in the PR is a bit long and a recap could help

src/ProxyClient/Fastly.php
andrerom: I'd rather move to use body then..

dbu:
hey hpatoio, good to see you (even if just virtually) - andré said that there are some tests missing. @andrerom can you explain in more detail what tests you would want? end-to-end tests with an actual fastly seems difficult to achieve to me...

andrerom:
> end-to-end tests with an actual fastly seems difficult to achieve to me...

this is what I had in mind

hpatoio:
> end-to-end tests with an actual fastly seems difficult to achieve to me...
Got it. For NGINX/Varnish we create services usign the `.sh` and the we run the e2e tests. For Fastly you wanna do the same but using a real Fastly service. Going to figure out something.

dbu:
can we wrap this up somehow? i'd be ok to merge without functional test and add a warning in the doc that fastly support is experimental.

Toflar:
To be fair I think there's nothing else we can do other than checking that the API requests are correct. We don't need to functional test the API of Fastly, that's their responsiblity. So imho this doesn't need to be experimental :)

andrerom:
Ok with me, rebased and pushed.

dbu:
@Toflar well, without an end-to-end test we can't be sure that our assumptions about the fastly api are correct. but yeah, utlimately that is up to everybody building an application with these components, as otherwise we still can't be sure if it works. i really would not want to run something that talks to fastly in our travis-ci setup.

@andrerom can you please apply the cs fixer thingies? the travis failure is "only" the python sphinx installation to test the doc. that is a known problem and not related to your changes.

src/ProxyClient/Fastly.php
dbu: @andrerom this question is still open. do you want to change it or keep as is?

src/ProxyClient/Fastly.php
dbu: @andrerom do you want to move this to body, or shall we merge as is and refactor later?

src/ProxyClient/Fastly.php
andrerom: done

andrerom:
@dbu CS fixes and change to use request body has been pushed.

doc/proxy-clients.rst
Toflar:
```suggestion
* NB: To be able to clear all cache(``->clear()``), you'll need a token for user with Fastly "Engineer permissions".
```

dbu:
thanks a lot for your persistence!

Toflar:
Thanks André - great stuff!

PR #454, opened 2019-05-30
fix: add a clear error message for xkey Varnish config when mode header is missing

dkarlovi:


dkarlovi:
Yeah, guess why I added it. :laughing:

PR #460, opened 2019-07-29
Fix tests on travis

andrerom:
Found so far:
- xenial is now default on travis, this does not seems to work with Varnish 3, stick to trusty for now
   - Bump back to xenial once we can remove Varnish3
- New version of `guzzlehttp/psr7/src/MessageTrait` is a bit more strict, and detects invalid parameters used in UserContextTestCase.php

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: php
 
 cache:
--- a/tests/Functional/Varnish/UserContextTestCase.php
+++ b/tests/Functional/Varnish/UserContextTestCase.php
@@ -52,12 +52,12 @@ abstract class UserContextTestCase extends VarnishTestCase
         $this->assertContextCache($cachedResponse2->getHeaderLine('X-HashCache'));
         $this->assertHit($cachedResponse2);
 
-        $headResponse1 = $this->getResponse('/user_context.php', ['Cookie' => ['0=foo'], [], 'HEAD']);
+        $headResponse1 = $this->getResponse('/user_context.php', ['Cookie' => ['0=foo']], 'HEAD');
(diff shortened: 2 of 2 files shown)


andrerom:
Unsure why this one fails:
```
2) FOS\HttpCache\Tests\Unit\SymfonyCache\EventDispatchingHttpCacheTest::testAddListener
Undefined index: trace_level
/home/travis/build/FriendsOfSymfony/FOSHttpCache/vendor/symfony/http-kernel/HttpCache/HttpCache.php:130
/home/travis/build/FriendsOfSymfony/FOSHttpCache/vendor/symfony/http-kernel/HttpCache/HttpCache.php:228
/home/travis/build/FriendsOfSymfony/FOSHttpCache/src/SymfonyCache/EventDispatchingHttpCache.php:98
/home/travis/build/FriendsOfSymfony/FOSHttpCache/src/Test/EventDispatchingHttpCacheTestCase.php:318
```

dbu:
unless debug is false. i think when it is false the addTraces can still be called and will then fail in this way. it only happens with newer php versions, maybe its a recent regression that does not pop up because people are either in debug mode or ignore the undefined index in production?

dbu:
its been added here: https://github.com/symfony/symfony/commit/9a2fcc9392c3da4718ae2c4d4a5877a9962a70f8#diff-3b14ed7783f3f7d7daf4536a9f3f1230

dbu:
https://github.com/symfony/symfony/issues/32801

dbu:
i will merge anyways, its much better than the current state :+1: 

thanks @andrerom !

dbu:
i went ahead and did a pull request to have it fixed in symfony https://github.com/symfony/symfony/pull/32802

PR #450, opened 2019-04-29
lock sphinx version

dbu:
the build failed.

meanwhile, when specifying python: 3.6, we get a bunch of ssl errors when trying to install.

PR #463, opened 2019-11-05
Use LegacyEventDispatcherProxy for Symfony 4.3

XWB:
This should fix the following deprecation message:

> User Deprecated: Calling the "Symfony\Component\EventDispatcher\EventDispatcherInterface::dispatch()" method with the event name as first argument is deprecated since Symfony 4.3, pass it second and provide the event object first instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+2.8.0
+-----
+
+### General
+
+* Use `LegacyEventDispatcherProxy` for Symfony >= 4.3 to avoid deprecation messages.
+
 2.7.0
 -----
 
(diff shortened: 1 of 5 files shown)


XWB:
It seems Travis fails due to an unrelated error:

> Undefined index: trace_level

dbu:
weird, that was a bug in symfony that i thought i fixed already once: https://github.com/symfony/symfony/pull/32802

dbu:
do you have time to debug that one locally? 

looking at https://github.com/symfony/symfony/blob/v4.3.6/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L97 i can't see how this could ever happen with v4.3.6 of symfony/http-kernel so we would have to debug if the kernel somehow does not get instantiated properly or some assumption about the options is wrong.

btw looking at my "fix" again, i notice that trace_level already was initialized in the defaults and my change was pointless, just complicating the code a bit.

dbu:
dug around the code a bit, i think the problem is with https://github.com/FriendsOfSymfony/FOSHttpCache/blob/01d87faf5cc48e2164b958b0a64ed7ff2b8b71ea/src/Test/EventDispatchingHttpCacheTestCase.php#L48-L50

we create a partial mock. dangerous territory :scream: 
i guess we should change getHttpCachePartialMock to always mock the addTraces method to avoid the issue.

XWB:
@dbu Nice find. Disabling the original constructor is a classic pitfall :) I just pushed another commit that fixes the unit tests.

There is still one test failing but seems to be a totally different issue.

dbu:
thank you!

yeah the sphinx build would be to check the documentation. i try to fix it in #450 but did not find why it was not working, and then did not find more time to continue...

PR #465, opened 2019-11-06
Bump branch alias

XWB:


dbu:
thanks! do you think your deprecation fixes should lead to a new minor version? i think a patch release would be enough as there is no new functionality - am i missing something?

PR #467, opened 2019-11-06
Update dispatch call

XWB:
I may have forgotten something in https://github.com/FriendsOfSymfony/FOSHttpCache/pull/463

### DIFF
--- a/src/SymfonyCache/EventDispatchingHttpCache.php
+++ b/src/SymfonyCache/EventDispatchingHttpCache.php
@@ -145,7 +145,16 @@ trait EventDispatchingHttpCache
     {
         if ($this->getEventDispatcher()->hasListeners($name)) {
             $event = new CacheEvent($this, $request, $response, $requestType);
-            $this->getEventDispatcher()->dispatch($name, $event);
+
+            // LegacyEventDispatcherProxy exists in Symfony >= 4.3
+            if (class_exists(LegacyEventDispatcherProxy::class)) {
+                // New Symfony 4.3 EventDispatcher signature
+                $this->getEventDispatcher()->dispatch($event, $name);
+            } else {
+                // Old EventDispatcher signature
+                $this->getEventDispatcher()->dispatch($name, $event);
+            }
+
             $response = $event->getResponse();
         }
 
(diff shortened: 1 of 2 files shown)


src/SymfonyCache/EventDispatchingHttpCache.php
```diff
@@ -145,7 +145,16 @@ protected function dispatch($name, Request $request, Response $response = null,
     {
         if ($this->getEventDispatcher()->hasListeners($name)) {
             $event = new CacheEvent($this, $request, $response, $requestType);
-            $this->getEventDispatcher()->dispatch($name, $event);
+
+            // LegacyEventDispatcherProxy exists in Symfony >= 4.3
+            if (class_exists(LegacyEventDispatcherProxy::class)) {
+                // New Symfony 4.3 EventDispatcher signature
+                $this->getEventDispatcher()->dispatch($event, $name);
```
dbu: isn't the point of the legacy proxy that we then can use the same code regardless of dispatcher version?

XWB: Apparently not. The `LegacyEventDispatcherProxy` triggers a deprecation notice if you use the old syntax: https://github.com/symfony/event-dispatcher/blob/4.3/LegacyEventDispatcherProxy.php#L63

And `LegacyEventDispatcherProxy` is not available in Symfony versions below 4.3 so there is no other way around this hack.

dbu: but if we do all that, why do we use the legacy dispatcher at all? does it do anything for us, or could we just conditionally use the new order if we are on a new enough version?

dbu: @XWB ping ;-)

dbu:
thanks @XWB ! @andrerom wrapped this up in #468

src/SymfonyCache/EventDispatchingHttpCache.php
XWB: @dbu Because if somebody extends `EventDispatchingHttpCache`, the `LegacyEventDispatcherProxy` will make sure the old dispatcher syntax will continue to work.

PR #472, opened 2019-11-28
test with php 7.4

dbu:
* [x] Restart when php 7.4 is available on travis-ci

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,19 +8,18 @@ cache:
 env:
   global:
     - VARNISH_VERSION=5.1
+    - SYMFONY_PHPUNIT_VERSION=8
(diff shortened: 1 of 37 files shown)


Toflar:
Any reason you're not using the Symfony PHPUnit bridge which already does all the compatibility stuff for you? 🙈

dbu:
what compatibility does it do? does it backport functionality like the assertStringContainsString? is just having the bridge available enough, or will i need to do something specific?

i should probably use it anyways, because some things in src/ are about phpunit < 7 while others are exclusively for 7 and not 8. and right now we only test 5.7 and 8. providing phpunit integration for others makes me want to also run a test with phpunit 6 and 7.

Toflar:
Yeah exactly. My gut feeling tells me that it might help you big time in this PR but needs to be verified. Just chipping in an idea here :)

PR #475, opened 2020-01-16
Fixed Symfony PurgeListener was pruning instead of purging

Toflar:
What a stupid mistake and why did nobody notice this so far? 😄 

Basically the clear-all mode of the Symfony integration only pruned (cleaned up expired entries) up until now. But it should actually clear it = deleting all the entries!
This wasn't even possible with my store implementation so [I quickly fixed this and released v2.2.0](https://github.com/Toflar/psr6-symfony-http-cache-store/commit/8cb5d97cb46a95b6b8e93f24cfeacbed12cca646).


Toflar:
Okay, so this PR also fixes an issue with Symfony 3/4/5 compat. One failing test is not related to this PR and the other one is PHP 7.0 compat because the PSR6 store is not compatible with 7.0.
Any reason why we are still compatible with 7.0?

src/SymfonyCache/PurgeTagsListener.php
```diff
@@ -119,8 +119,13 @@ public function handlePurgeTags(CacheEvent $event)
 
         $tags = [];
 
-        foreach ($request->headers->get($this->tagsHeader, '', false) as $v) {
-            foreach (explode(',', $v) as $tag) {
+        // Compatibility with Symfony 3+
+        $allHeaders = $request->headers->all();
```
dbu: i would prefer to have a reflection check if `all` takes a parameter and if it does, you can `$request->headers->all($this->tagsHeader);` and otherwise keep the old way with `get`. and add a note that when we drop support for symfony < 4.4 we can drop the check.

i prefer that over this code that is harder to read and won't get cleaned up.

PR #480, opened 2020-05-07
Allowed server URI parts are extended: added user and pass

SerheyDolgushev:
Allow providing HTTP access credentials for the servers.


SerheyDolgushev:
Failing checks do not seem to be related to the current PR.

PR #482, opened 2020-05-11
Use userInfo from server configuration when duplicating request

ITernovtsiy:
Currently, information about "userInfo" not copied, and we see 401 response returned from cache proxy.

### DIFF
--- a/src/ProxyClient/HttpDispatcher.php
+++ b/src/ProxyClient/HttpDispatcher.php
@@ -223,13 +223,18 @@ class HttpDispatcher implements Dispatcher
 
         // Create a request to each caching proxy server
         foreach ($this->getServers() as $server) {
-            $requests[] = $request->withUri(
-                $uri
-                    ->withScheme($server->getScheme())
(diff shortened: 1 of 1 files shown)


Toflar:
Thanks for contributing! Can you add some tests so we can see what's actually going wrong now?

SerheyDolgushev:
It is related to https://github.com/FriendsOfSymfony/FOSHttpCache/pull/480

dbu:
makes sense with #480.

but indeed, a test for the whole scenario would be very welcome. can you do one please?

SerheyDolgushev:
@dbu Thanks a lot for a quick merge!  :)

ITernovtsiy:
yes, I'll work on tests later and open a separate PR, thanks!

SerheyDolgushev:
> makes sense with #480.
> 
> but indeed, a test for the whole scenario would be very welcome. can you do one please?

We will create an internal ticket to implement a test for this change, so we will not forget about it. And we will submit a separate PR for the test.

Issue #284, opened 2016-03-22
make it easier to troubleshoot error-400 responses


gggeek:
In UserContextSubscriber, if the user-hash header is missing, an error 400 will be returned to the user:
- with no indication of the root of the problem to the user
- with no indication in the logs, either

It would be nice if this was easier to troubleshoot

gggeek:
As a side note: a single line in the apache config can be used to trigger the usage of httpcache in a  context where there is a reverse proxy in front, and the error 400 to appear

dbu:
i agree that good error responses are an important time safer for developers. glad if we can improve something.

are you talking about https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/master/EventListener/UserContextSubscriber.php (in the bundle repository, not this one)? i don't see where we would trigger the error 400. glad if you can do a pull request with a simple solution (maybe simply have a description on the exception so that its logged / shown in debug mode?)

gggeek:
I think that it is in https://github.com/FriendsOfSymfony/FOSHttpCache/blob/8019bad9b3c1746263c5d83df844914983fa8fae/src/SymfonyCache/UserContextSubscriber.php, line 102.

Will try to find the tome to think about a good message which compromises between information and security

dbu:
ah, i see. right, lets add something there. glad if you can do a pull
request. i think as long as its a static message and not using anything
from the request it should be ok.
can this happen in other situations than when custom development on the
AppCache or an extended event subscriber class lead to some mismatch?

gggeek:
a million dollar question... :-)

dbu:
ping @gggeek

dbu:
@gggeek i aim to release 2.0 in september. it would be great if you find time to come up with an idea here. now we still can do BC breaks if needed. once we have a beta release, we won't do BC breaks anymore for non-critical issues.

gggeek:
@dbu I ll try to give it a look over the weekend...

dbu:
i will close this issue for now. but if you find time to look into this and provide an improvement PR, please do @gggeek . the release did not happen but i am about to finish the rewriting of the HTTP layer to use HTTPlug.

PR #484, opened 2020-07-15
v1.4: Add generic error response to request tampering response

ryanolee:
# Problem
After spending many more hours than I would like to say on an issue ultimately pertaining to #284 I think it would be a fairly good idea of giving some indicator of what went wrong.

# Proposed solution
Given the fairly odd place at which this code can be executed in the symfony request cycle it is fairly difficult to find a good way to do logging here without making significant changes. Beyond that the handler is run so early there is no guarentee symfony's core error handling will be set up by the time this code is executed. Blackholeing the error however makes it immensely difficult (beyond setting up xdebug :sweat: ) to find where the error is been given. 

Therefore returning a response with the error that is generic enough for a attacker to not necessarily know where it is coming from, even with targeted searches https://github.com/search?q=bad+request&type=Code , but verbose enough for a developer actively searing their code base for the exception to find it .

Let me know if you need any more input from me on this! Thanks!

### DIFF
--- a/src/SymfonyCache/UserContextSubscriber.php
+++ b/src/SymfonyCache/UserContextSubscriber.php
@@ -117,7 +117,7 @@ class UserContextSubscriber implements EventSubscriberInterface
             if ($request->headers->get('accept') === $this->options['user_hash_accept_header']
                 || null !== $request->headers->get($this->options['user_hash_header'])
             ) {
-                $event->setResponse(new Response('Bad request given', 400));
+                $event->setResponse(new Response('bad request', 400));
 
                 return;
             }
--- a/src/SymfonyCache/UserContextSubscriber.php
+++ b/src/SymfonyCache/UserContextSubscriber.php
@@ -117,7 +117,7 @@ class UserContextSubscriber implements EventSubscriberInterface
             if ($request->headers->get('accept') === $this->options['user_hash_accept_header']
                 || null !== $request->headers->get($this->options['user_hash_header'])
             ) {
-                $event->setResponse(new Response('', 400));
+                $event->setResponse(new Response('Bad request given', 400));
 
                 return;
             }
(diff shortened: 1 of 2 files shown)


dbu:
hi, ryanolee, thanks for the pull request. yeah this indeed not a good developer experience. i think at the time of writing, we assumed this would only fail when somebody is trying to hack the site, but at least when doing custom hashes, it could happen because of a mistake. indeed logging is not trivial, as it happens in the http cache which is created in the front controller and not with the bundle. i see no risk in outputting this generic text and it could indeed help by giving something to search in the code. alternatively, we could think if we want to output some numeric error codes in addition, but that becomes a bit more searchable.

why are you doing the pull request against the 1.4 branch? the current release of this library is 2.9... from your PR title i see that you targetted version 1 on purpose. i am a bit worried of changing the 1.x version for this, when i expect people to use 2.x in new projects.

ryanolee:
Hi @dbu ,
Thanks for the quick reply! I can understand why we might not necessarily want to update an otherwise stable version of the packaged given it has not been really touched at that version in the better part of two years :thinking:  

The reason I am specifically targeting 1.4 here is because this issue raised itself as part of integrating varnish onto an existing ez publish 5 site.  Specifically I have a hard dependency on  ezsystems/ezpublish-kernel@6.5.0 Which requires friendsofsymfony/http-cache-bundle@">=1.2, <=1.3.6" giving a version of friendsofsymfony/http-cache-bundle@1.3.6 which as a constraint for friendsofsymfony/http-cache@"~1.4" giving a version of friendsofsymfony/http-cache@1.4.2. 

Given there is no particularly easy ez upgrade path between ezpublish 5 -> ezplatform the site is currently limited to that version of the kernel. Hence making it difficult to use any other version of package on my end :+1: .
Thanks!
-Ryan

dbu:
thanks! i merged manually (adding a changelog, squashing the commits and chaning "Bad Request" to uppercase) in aec6cbf31cb6947f4d44653aba093472777c0e17. will tag a release and merge it up to version 2.

ryanolee:
Thanks @dbu !

dbu:
you are welcome. thanks for contributing!

Issue #487, opened 2020-11-20
PHP 8.0 compatibility


ausi:
Are there any plans for adding compatibilty with PHP 8.0?

Would you like me to create a pull request for this?

dbu:
yes please. and this one should be easier than the cmf stack, i hope.

Toflar:
I think the phpunit stuff will be PITA. @dbu maybe it's time for a new major version here too? Drop all the stuff that's Symfony < 4.4 just like I did for the PSR-6 store? I mean, you're the main maintainer of that lib but imho PHP 8 would be reason enough to finally get rid of the old stuff and lower the maintenance burden on your side?

dbu:
i don't mind dropping support for php 7.1 and symfony < 4.4 if that helps. i don't know how deeply we go into symfony, not sure if removing symfony 3.4 support makes much of a difference.

doing that does not require a new major version, just a new minor version. as long as we don't change any interface signatures / method signatures. it would be nice to start doing strict typing, but maybe better do that later and for now focus on getting a php 8 supporting release out.

ausi:
Closed in favor of #488

PR #488, opened 2020-12-02
PHP 8.0 Compatibility

ausi:
As discussed in #487

This is currently a draft. ☺️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,8 @@ matrix:
     - php: 7.3
     - php: 7.4
       env: DOCCHECK=true COVERAGE=true PHPUNIT_FLAGS="-v --coverage-text" VARNISH_MODULES_VERSION=0.12.1
+    - php: 8.0
+      env: COMPOSER_FLAGS="--ignore-platform-req=php" SYMFONY_PHPUNIT_VERSION=8
 
       # Test with Varnish 4
     - php: 7.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,7 +90,7 @@ script:
   - if [[ "$DOCCHECK" = true ]]; then make -C doc spelling; fi
 
 after_script:
-  - wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover
+  - if [[ $COVERAGE = true ]]; then wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover coverage.clover; fi
 
 after_failure:
   - cat /tmp/fos_nginx_error.log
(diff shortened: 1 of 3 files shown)


ausi:
All dependencies except `php-http/guzzle6-adapter` are compatible with PHP 8. I’m not sure if we can get rid of `php-http/guzzle6-adapter` easily? But because it is only a dev dependency it should not hold us back from marking this package compatible with php 8 (if everything else works).

dbu:
thanks! 

i suggest to use guzzle7-adpater, as guzzle 6 is not compatible with php 8. guzzle 7 works with php 7.2 - 8.0.
lets drop php 7.1 support to keep things simple. (and users on php 7.1 can keep using the current version, its not like we have a lot of new features happening at the moment)

i fixed the styleci error in a separate branch, please rebase.
can you please put the branch alias in composer.json to 2.10.x? (it should currently be 2.9.x but was forgotten to be updated). and please add a note in the CHANGELOG.md file

ausi:
I was unable to fix the `sphinxcontrib-spelling requires Python '>=3.6' but the running Python is 2.7.6` error in
<https://travis-ci.com/github/FriendsOfSymfony/FOSHttpCache/jobs/452324683#L856>

But I think it is not related to this pull request

.travis.yml
```diff
@@ -10,21 +10,25 @@ env:
...
+    - php: 7.2
       env: VARNISH_VERSION=3.0 SYMFONY_PHPUNIT_VERSION=5.7 COMPOSER_FLAGS="--prefer-lowest" DEPENDENCIES=""
 
-    - php: 7.1
-      env: SYMFONY_PHPUNIT_VERSION=7
     - php: 7.2
     - php: 7.3
     - php: 7.4
       env: DOCCHECK=true COVERAGE=true PHPUNIT_FLAGS="-v --coverage-text" VARNISH_MODULES_VERSION=0.12.1
+      python: 3.6
```
dbu, line 26: lets drop this line as it did not help. we can ignore the failure and i will comment out the doccheck in a separate PR

dbu (changes requested):
awesome, thanks a lot! ok to remove the python version again? we can try to fix that in a different PR

.travis.yml
ausi: Done in a05cb960adf73894e936e746413eb6e890b69cf9

ausi:
Everything else is green now 🎉

dbu (approved):
thanks a lot!

i will tag the release after merging this.

i skip the doccheck for now, if nobody fixed it in the last 1.5 years, chances are it won't soon be fixed anyways...

ausi:
Thank you @dbu

Should I create a pull request for https://github.com/FriendsOfSymfony/FOSHttpCacheBundle too?

dbu:
if you have the time to do it, i'd appreciate it :+1:

ausi:
I will look into it.

PR #497, opened 2021-06-18
improve caching of the user context hash lookup in Varnish

### LINKED ISSUE #496: [Performance] Varnish user context hash vcl should use the hash in vcl_hash
We use Varnish and user context hash on very high traffic website. And we experienced various performance and stability issue I wanted to share with you. All credits for the finding does not belong to me but to my client and his exchanges with Varnish support. I think it's important enough to share it here.

User context hash is based on a new header that we vary on. This header itself is calculated on the /_fos_user_context_hash call that itself is vary against the cookie or authorization header.

If we use user context hash it's because we want the cardinality of an user context page to be as low as posible to avoid pointless computation.
But the cardinality of the _fos_user_context_hash is linked to the number of visitor, and it can be very big. [We relates on the Vary header](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/5d1f62825e995afdf86a720e47484ed2543283ac/src/EventListener/UserContextListener.php#L165) to make the distinction and find in all the cached variants.

The vary header is like a sql where on a not indexed field, against a possible enormous table if you have a traffic like this one.
To fix it we just need to index this field.

```
sub vcl_hash {
    if (req.url == "/_fos_user_context_hash") {
        hash_data(req.http.<Cookie or Autorization>);
    }
    # VS: NOTE: No return, fallthrough to builtin VCL
}
```

Before on our load test:

image

After:

image

On production, we reduced our number of varnish servers by ten.

The comportment is [now explain in the next release varnish documentation](https://github.com/varnishcache/varnish-cache/pull/3520/files#diff-09291b164556649a3f302cac0242fe37da79c35c2e9bb06bd1866536689f4ac0R137) along with a warning. The warning is triggered if there is more than 10 variants. And I hope everybody using user context hash has more than 10 logged in users otherwise...

dbu:
Thanks to Bastian Jaillot (@bastnic) for reporting the performance
analysis and proposing this solution.

Further details can be found at https://github.com/FriendsOfSymfony/FOSHttpCache/issues/496

fix #496

### DIFF
--- a/doc/varnish-configuration.rst
+++ b/doc/varnish-configuration.rst
@@ -295,6 +295,10 @@ To enable this feature, add the following to ``your_varnish.vcl``:
             call fos_user_context_recv;
         }
 
+        sub vcl_hash {
+            call fos_user_context_hash;
+        }
+
         sub vcl_backend_response {
             call fos_user_context_backend_response;
         }
@@ -310,6 +314,10 @@ To enable this feature, add the following to ``your_varnish.vcl``:
 
         sub vcl_recv {
             call fos_user_context_recv;
+        }
+
+        sub vcl_hash {
+            call fos_user_context_hash;
         }
 
         sub vcl_fetch {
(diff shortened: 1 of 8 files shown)


dbu:
@bastnic may i ask you to review this pull request?

doc/varnish-configuration.rst
```diff
@@ -341,6 +349,13 @@ request with :ref:`a proper user hash <return context hash>`.
     set the ``req.url`` to a fixed URL. Otherwise Varnish would cache every
     hash lookup separately.
 
+    The ``fos_user_context_hash`` should be used to separate the cache of the
+    hash lookup. If you don't do that, Varnish can run into performance issues
```
bastnic, line 353: maybe link [the code change](https://github.com/varnishcache/varnish-cache/pull/3520/files#diff-09291b164556649a3f302cac0242fe37da79c35c2e9bb06bd1866536689f4ac0R137] or the original PR?

bastnic (approved):

doc/varnish-configuration.rst
dbu: good idea, i link that PR now

PR #499, opened 2021-06-25
Fix exception creation in `HttpDispatcher`

jdreesen:
[`FOS\HttpCache\Exception\InvalidArgumentException`](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/39999a6c60d8bdc756436dd0739202d897d29351/src/Exception/InvalidArgumentException.php) just extends the base one, without any custom code. Which means the first parameter should be the message, followed by the code and then the previous exception.

### DIFF
--- a/src/ProxyClient/HttpDispatcher.php
+++ b/src/ProxyClient/HttpDispatcher.php
@@ -142,7 +142,7 @@ class HttpDispatcher implements Dispatcher
                 try {
                     $promises[] = $this->httpClient->sendAsyncRequest($proxyRequest);
                 } catch (\Exception $e) {
-                    $exceptions->add(new InvalidArgumentException($e));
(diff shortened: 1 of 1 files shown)


dbu (approved):
oops, you are right!

thanks a lot!

dbu:
released as https://github.com/FriendsOfSymfony/FOSHttpCache/releases/tag/2.10.1

src/ProxyClient/HttpDispatcher.php
```diff
@@ -156,7 +156,7 @@ public function flush()
                 $exceptions->add(ProxyUnreachableException::proxyUnreachable($exception));
             } catch (\Exception $exception) {
                 // @codeCoverageIgnoreStart
```
dbu, line 158: famous last words :-(

PR #501, opened 2021-07-07
phpstan static analysis with github workflow

dbu:


phpstan.neon.dist
```diff
@@ -0,0 +1,10 @@
+parameters:
+    level: 1
+    paths:
+        - src
+    excludePaths:
+        analyseAndScan:
+            # contains code to support legacy phpunit versions
+            - src/Test/Legacy/*
```
dbu, line 8: according to https://phpstan.org/user-guide/ignoring-errors#excluding-whole-files this should work, but we get a fatal error when parsing files in src/Test/Legacy: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/501/checks?check_run_id=3006267281

@ondrejmirtes may i bother you with this? is the documentation incorrect, or am i doing something wrong? according to the output, we are on phpstan 0.12.91
the src/BaseEvent.php is indeed ignored, so i don't think the base path is wrong.

ondrejmirtes: Try to run with `--debug` to see which file actually fails with this - it might still get loaded for some reason when analysing a non-excluded file. It's possible to circumvent `excludePaths` for example if Composer is set up to autoload files from this directory.

.github/workflows/static.yml
```diff
@@ -0,0 +1,24 @@
...
+  phpstan:
+    name: PHPStan
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Pull in optional dependencies
+        run: composer require symfony/http-kernel symfony/event-dispatcher symfony/process phpunit/phpunit psr/log toflar/psr6-symfony-http-cache-store
```
dbu, line 19: TODO: require mockery/mockery and also analyse the tests folder

phpstan.neon.dist
ondrejmirtes: See - it fails during analysis of `/github/workspace/src/Test/WebServerListener.php`.

dbu: thanks a lot!

i did not realize that it happens that way. we do some horrible stuff in those files to support various versions of phpunit... in regular php execution, it only follows one of the code paths because we are either on an old or a new phpunit version. makes sense that the static analysis gets tripped by this. i will exclude the things and look into cleaning this up - i guess nobody still uses this with pre 6.0 phpunit versions :see_no_evil:

PR #506, opened 2021-12-10
Fix php 8.1 deprecation

alexander-schranz:


dbu (commented):
thanks! can't we narrow the return type in all PHP versions that we support here, so that we could even today declare the return types of those methods?

alexander-schranz:
For BC reasons we can not yet add return types here else it could break if somebody maybe extend from them. The are here just defined as phpdoc types.

dbu:
one of the reasons why we should make such things final (which of course is only possible in a major version)

Issue #457, opened 2019-06-20
Generic CDN support


diimpp:
Hello,

I'm tasked with implementing CloudFlare cache invalidation and I've noticed, that `HttpProxyClient` design is not really compatible with generic CDN cache invalidation flow.

What usually happens with CDN is rest post request with json body. Single or multiple urls, headers, etc.

And for example it's not possible to supply body to the request factory of the HttpProxyCache
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/ProxyClient/HttpProxyClient.php#L94

I do realise, that I can just implement `ProxyClient` interface and do the rest of the implementation as I need to.
And that `HttpProxyClient` is not generic solution, but a specialized version for direct communication with proxy servers line nginx or varnish, but it's kind of feels wrong to have `HttpProxyClient` name wise, while it doesn't allow to send body.

Not sure what this issue is about. :) RFC for renaming `HttpProxyClient`? Add some mention on CDN's or custom ProxyClient implementation?

CF purge request looks like so
```bash
curl -X POST "https://api.cloudflare.com/client/v4/zones/023e105f4ecef8ad9ca31a8372d0c353/purge_cache" \
     -H "X-Auth-Email: user@example.com" \
     -H "X-Auth-Key: abcdefg" \
     -H "Content-Type: application/json" \
     --data '{"files":["http://www.example.com/css/styles.css",{"url":"http://www.example.com/cat_picture.jpg","headers":{"Origin":"cloudflare.com","CF-IPCountry":"US","CF-Device-Type":"desktop"}}]}'
```

Related https://github.com/FriendsOfSymfony/FOSHttpCache/issues/142
https://github.com/FriendsOfSymfony/FOSHttpCache/pull/403

dbu:
hi, great to hear this! @andrerom is looking into fastly support (see #451), which is also a CDN. i think he noticed similar needs. @Toflar is working on litespeed support (see #444), which again asks us to make some concepts more flexible. i would love for FOSHttpCache to be more flexible for these scenarios. please start by looking at the fastly PR to see if that already addresses some of your concerns. we could also extract some refactorings from there and merge early (e.g. allow request body), so that you both can continue the work on CDN support.

andrerom:
Once #451 is in, we can consider adding the following which now seems to have cache tagging support:
- Akamai: https://developer.akamai.com/blog/2019/03/25/introducing-ability-purge-cache-tag
[...]

PR #500, opened 2021-07-03
use github workflow

dbu:
fix #492

TODO
* figure out varnish setup problems
* edit README to show github actions status instead of travis status

PR #507, opened 2021-12-16
Cloudflare support

simonrjones:
**Background**
I'm working on the w3.org redesign project and we'd like to add Cloudflare support to FOSHttpCache so we can use this library to help control full page caching. 

**Changes**
* Add clear, purge and invalidateTags methods for Cloudflare API
* Purge by path queues requests and sends in one request (can send 30 purge path requests per API request)
* Add docs & test (using Mockery)

**Notes**
I've tested this with real-world code (w3.org) for clear and purge. I need to test clearing tags but manual API tests work fine. 

I can't build the docs locally so I'm not sure if I've got the formatting right for this. Also, the docs confused me a little since I had to specify the Cloudflare API URL in the HttpDispatcher to get this working. Not sure if this is correct & expected when working with an external CDN like Cloudflare:

     $httpDispatcher = new HttpDispatcher(['https://api.cloudflare.com']);
     $cloudflare = new Cloudflare($httpDispatcher);

For details about Cloudflare API:

    https://api.cloudflare.com/#zone-purge-all-files
    https://developers.cloudflare.com/cache/how-to/purge-cache#h_6d756ac9-c476-45e8-a5d4-e2a6e45d9dc7

Related discussion: #457


src/ProxyClient/Cloudflare.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * JSON encode data.
+     *
+     * @param $data
+     *
+     * @return false|string
+     */
+    protected function encode($data)
+    {
+        return json_encode($data, JSON_UNESCAPED_SLASHES);
```
diimpp, line 57: Would it be better to throw exception here instead of `false` on encoding failure?
Otherwise `false` will be sent as body of invalidation request?

simonrjones: That's a good idea, but `JSON_THROW_ON_ERROR` is only available since PHP 7.3 so is that OK? This library is currently marked as PHP 7.2+ in the composer.json file. An alternative is to detect a false return value and manually throw an exception.

simonrjones:
I am not certain why the PHPStan, Travis CI and Scrutinizer status checks are failing. Can anyone give me any tips on this, or are there issues with the current GitHub action setup?

diimpp:
Judging by commit history, CI is dead since May.

src/ProxyClient/Cloudflare.php
diimpp: Personally I would go with return value check, something like that

```
$json = json_encode($data);
if (false === $json && json_last_error() !== JSON_ERROR_NONE) { // Technically only of those checks is needed.
    throw new \InvalidArgumentException('Cannot encode "$data": ' . json_last_error_msg());
}
```

simonrjones: AFAIK json_encode always returns false on error, so I've just used that check.

simonrjones:
@diimpp The CI seems to fail on:

- static / PHPStan src - I'm not sure what this does since phpstan-tests already runs phpstan and passes fine?
- I can't really see where [Travis is failing](https://app.travis-ci.com/github/FriendsOfSymfony/FOSHttpCache/jobs/552574390) unless this is from the risky & skipped tests? But if you have GitHub Actions now which run PHPUnit is there a need for TravisCI as well?
- Scrutinizer fails with the following error. This looks like a configuration issue.

> The code coverage data was not received within the specified time. Please make sure your third-party service is configured to send code coverage data.

diimpp:
> @diimpp The CI seems to fail on:
> 
>     * static / PHPStan src - I'm not sure what this does since phpstan-tests already runs phpstan and passes fine?
> 
>     * I can't really see where [Travis is failing](https://app.travis-ci.com/github/FriendsOfSymfony/FOSHttpCache/jobs/552574390) unless this is from the risky & skipped tests? But if you have GitHub Actions now which run PHPUnit is there a need for TravisCI as well?
> 
>     * Scrutinizer fails with the following error. This looks like a configuration issue.
> 
> 
> > The code coverage data was not received within the specified time. Please make sure your third-party service is configured to send code coverage data.

/cc @dbu

dbu:
thanks a lot for the pull request! and happy to hear this library is powering w3.org :blush: 

travis-ci changed their offering and many project are moving away to github actions since then. i moved my other projects, but got stuck in #500 with trying to set up varnish in github actions. i will give it another go.

scrutinizer fails because it would be the phpunit build that should push data to it...

phpstan build probably fell apart with the symfony 6 release, i will look into that as well.

the rendered documentation is built with an action in readthedocs - when you click on "Details" in the build action for readthedocs, you can check the docs for this branch.

src/ProxyClient/Cloudflare.php
```diff
@@ -0,0 +1,169 @@
...
+        if (false === $json) {
+            throw new \InvalidArgumentException(sprintf('Cannot encode "$data": %s', json_last_error_msg()));
+        }
+
+        return $json;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
```
dbu, line 69: might want to add "Tag invalidation only available with cloudflare enterprise account" here as well.

src/ProxyClient/Cloudflare.php
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * JSON encode data.
+     *
+     * @param $data
+     *
+     * @return string
+     *
+     * @throws InvalidArgumentException
+     */
+    protected function encode($data)
```
dbu, line 57: why protected and not private?

doc/proxy-clients.rst
```diff
@@ -296,6 +298,57 @@ And adapt your bootstrapping code to use the cache kernel::
...
+    $httpDispatcher = new HttpDispatcher(['https://api.cloudflare.com']);
+    $cloudflare = new Cloudflare($httpDispatcher, $options);
+
+.. note::
+
+    Cloudflare supports different cache purge methods depending on your account. All Cloudflare accounts support purging
+    the cache by URL and clearing all cache items. You need a `Cloudflare Enterprise`_ account to purge by cache tags.
+
+    For purging the cache by URL please see the docs `Cloudflare Purge by URL`_ for information about headers you can
+    pass to clear the cache correctly.
```
dbu, line 340: i am a bit confused by this sentence. does the user have to configure something specifically, or do invalidation requests in a specific way? or is that more of an implementation note? if the later, i would put it in the class phpdoc of the cloudflare ProxyClient

src/ProxyClient/Cloudflare.php
```diff
@@ -0,0 +1,169 @@
...
+ * - authentication_token  API authorization token, requires Zone.Cache Purge permissions
+ *
+ * @author Simon Jones <simon@studio24.net>
+ */
+class Cloudflare extends HttpProxyClient implements ClearCapable, PurgeCapable, TagCapable
+{
+    /**
+     * @see https://api.cloudflare.com/#getting-started-endpoints
+     */
+    private const API_ENDPOINT = 'https://api.cloudflare.com/client/v4';
```
dbu, line 32: the current logic in the HttpDispatcher (clearly driven by how things work with local caching proxies) is that it is configured with a list of server domain names/ips and sends a copy of the request to each proxy. so e.g. a refresh request for mydomain.com/path would be sent to the host varnish-proxy1.mydomain.com and varnish-proxy2.mydomain.com with a `Host` header of mydomain.com and the /path.

so this domain will be overwritten by the dispatcher. it would be better to have /client/v4 here, without the domain.

alternatively we would need to refactor the HttpDispatcher for this use case, to keep the domain intact. that would make more sense. i think best we would create a base dispatcher that does not have the multiplex logic at all, and have the current HttpDispatcher extend that. then it will be possible to use the base dispatcher for this proxy client instead of the HttpDispatcher that provides unnecessary functionality.

dbu (commented):
i looked at the code, and seems good to me. i commented on a few nitpicks, and replied to your question about the HttpDispatcher.

dbu:
i fixed the build in the master branch. can you rebase please?

src/ProxyClient/Cloudflare.php
simonrjones: >  it would be better to have /client/v4 here, without the domain.

Thanks for confirming, have updated this.

doc/proxy-clients.rst
simonrjones: It contains details on how Cloudflare purges the cache by URL and which headers you need to pass to clear the cache correctly (if these are set in HTTP responses). 

I have added this to the code phpdoc, but I think this is useful to know about in the docs too since it may confuse users if the cache is not cleared as they expect. I've reworded to try to make this clearer.

doc/proxy-clients.rst
```diff
@@ -296,6 +298,57 @@ And adapt your bootstrapping code to use the cache kernel::
...
+
+The `Cloudflare`_ client sends HTTP requests with the ``HttpDispatcher``. Create the
+dispatcher as explained above, making sure you set the server to the Cloudflare API `https://api.cloudflare.com`, and
+pass it to the Cloudflare client::
+
+    use FOS\HttpCache\ProxyClient\Cloudflare;
+    use FOS\HttpCache\ProxyClient\HttpDispatcher;
+
+    $httpDispatcher = new HttpDispatcher(['https://api.cloudflare.com']);
+    $cloudflare = new Cloudflare($httpDispatcher);
```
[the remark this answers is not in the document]
simonrjones: I was following the example from Fastly, but I've updated this now, thanks.

simonrjones:
Rebased master & updated my branch with a few changes as suggested above.

dbu:
finished in #516 - thanks a lot @simonrjones for this contribution!

PR #514, opened 2021-12-22
Fix badges

villfa:
I've replaced the Travis badge by one from the Github actions.

Also the Scrutinizer badge displays an "unknown" status. That's because the code rating is not enabled. See: https://scrutinizer-ci.com/docs/code_rating_system

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,6 +1,13 @@
 imports:
   - php
 
+build:
+  nodes:
+    analysis:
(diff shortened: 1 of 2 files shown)


dbu:
oh, missed the badge when i switched. thanks! and thanks for the configuration of the rating. i guess it goes to 10, so 9.91 sounds good?

villfa:
Yes, it's a pretty good score :)

PR #516, opened 2022-01-12
Simonrjones cloudflare support

dbu:
wrapping up #507 

@simonrjones i squashed the commits and rebased on the master branch to get rid of the additional commits in the diff. i have reformatted the documentation a bit, and done some tweaks to the code (see the separate, second commit). can you please check if that all makes sense or if i messed up something?

### DIFF
--- a/doc/proxy-clients.rst
+++ b/doc/proxy-clients.rst
@@ -26,11 +26,13 @@ Varnish       ✓       ✓       ✓       ✓
 Fastly        ✓       ✓               ✓       ✓
 NGINX         ✓       ✓
 Symfony Cache ✓       ✓               ✓ (1)   ✓ (1)
+Cloudflare    ✓                       ✓ (2)   ✓
 Noop          ✓       ✓       ✓       ✓       ✓
 Multiplexer   ✓       ✓       ✓       ✓       ✓
 ============= ======= ======= ======= ======= =======
 
-(1): Only when using `Toflar Psr6Store`_.
+| (1): Only when using `Toflar Psr6Store`_.
+| (2): Only available with `Cloudflare Enterprise`_.
 
 If needed, you can also implement your own client for other needs. Have a look
 at the interfaces in namespace ``FOS\HttpCache\ProxyClient\Invalidation``.
@@ -90,7 +92,8 @@ Varnish Client
 ~~~~~~~~~~~~~~
 
 The Varnish client sends HTTP requests with the ``HttpDispatcher``. Create the
-dispatcher as explained above and pass it to the Varnish client::
+dispatcher as explained :ref:`above <HTTP client configuration>` and pass it to
+the Varnish client::
 
     use FOS\HttpCache\ProxyClient\Varnish;
 
@@ -163,11 +166,12 @@ Fastly Client
 ~~~~~~~~~~~~~~
 
 The Fastly client sends HTTP requests with the ``HttpDispatcher``. Create the
-dispatcher as explained above and pass it to the Fastly client::
+dispatcher as explained :ref:`above <HTTP client configuration>` and pass it to
+the Fastly client::
 
     use FOS\HttpCache\ProxyClient\Fastly;
 
-    $varnish = new Fastly($httpDispatcher);
+    $fastly = new Fastly($httpDispatcher);
 
 .. note::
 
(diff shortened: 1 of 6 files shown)


src/ProxyClient/Cloudflare.php
```diff
@@ -0,0 +1,180 @@
...
+        foreach (\array_chunk($this->purgeByUrlsData, self::URL_BATCH_PURGE_LIMIT) as $urlChunk) {
+            $this->queueRequest(
+                'POST',
+                sprintf(self::API_ENDPOINT.'/zones/%s/purge_cache', $this->options['zone_identifier']),
+                [],
+                false,
+                $this->json_encode(['files' => $urlChunk])
+            );
+        }
+        $this->purgeByUrlsData = [];
```
dbu, line 137: i added this - we need to clean up this array, in case the invalidation is done from some long running job that periodically flushes.

src/ProxyClient/Cloudflare.php
```diff
@@ -0,0 +1,180 @@
...
+     * @var array<string|array{url: string, headers: string[]}>
+     */
+    private $purgeByUrlsData = [];
+
+    public function __construct(
+        Dispatcher $httpDispatcher,
+        array $options = [],
+        RequestFactory $messageFactory = null
+    ) {
+        if (!function_exists('json_encode')) {
```
dbu, line 57: i added a constructor to check about ext-json, because it is a "soft-dependency" and not explicitly noted in the composer.json (and most clients don't need json, but this one and fastly do)

doc/proxy-clients.rst
```diff
@@ -296,6 +302,66 @@ And adapt your bootstrapping code to use the cache kernel::
...
+
+    Cloudflare supports different cache purge methods depending on your account.
+    All Cloudflare accounts support purging the cache by URL and clearing all
+    cache items. You need a `Cloudflare Enterprise`_ account to purge by cache
+    tags.
+
+    When purging the cache by URL see the `Cloudflare Purge by URL`_ docs for
+    information about how Cloudflare purges by URL and what headers you can
+    pass to a :doc:`invalidatePath() <cache-invalidator>` request to clear the
+    cache correctly.
```
dbu, line 352: @simonrjones do you have a code example how purging would look? maybe one with the simple case, and one with special cloudflare headers?

src/ProxyClient/Cloudflare.php
```diff
@@ -169,4 +167,14 @@ protected function configureOptions()
 
         return $resolver;
     }
+
+    private function json_encode(array $data): string
```
simonrjones, line 171: @dbu a minor point but should this method be named `jsonEncode()` to fit in with naming format used elsewhere?

dbu: my thinking was that it then looks like the native function, to be clear it is just a wrapper around that. in our applications, we use https://github.com/thecodingmachine/safe to wrap around all the old school php functions to have them throw exceptions. but i don't want to introduce that here as a dependency.

ok to keep it with underscore, or do you find it confusing that way?

doc/proxy-clients.rst
simonrjones: I've only tested this with plain URLs (no headers), but an example:

```
An example of clearing the cache by URL::

    $cacheInvalidator->invalidatePath('https://example.com/path')->flush();
```


I think you'd also need to add on line 329:

```
    $cacheInvalidator = new CacheInvalidator($cloudflare);
```

src/ProxyClient/Cloudflare.php
simonrjones: That makes sense, please do leave as is :)

doc/proxy-clients.rst
dbu: thanks. yeah adding $cacheInvalidator would make it complete.

i think i will change the doc where i talk about the baseUri to explicitly say that baseUri does not work with the cloudflare client. (other clients send single requests to the url that should be purged, so the dispatcher can add the domain to the url and invalidation could be just '/path'. in symfony you could then generate just the path of the url, which can be handy e.g. in console context where the domain name might not be known. if we would want the same with the cloudflare client, we would need to pass the baseUri to the cloudflare client and have that logic there. its rather involved though: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/843a10292ba5496b5adb2a143eac5b37ead13680/src/ProxyClient/HttpDispatcher.php#L194-L218 - if we want to share it, we could extract this into a trait that we use in both HttpDispatcher and the Cloudflare client. but we can also do that later, if the need even arises)

dbu: i updated the documentation. care to have another look if it makes sense like this, before i merge it?

simonrjones: Great, all looks good to me 👍

simonrjones (approved):

PR #519, opened 2022-01-14
install legacy varnish

dbu:
make legacy build work again, with help from https://www.varnish-software.com/developers/tutorials/installing-varnish-ubuntu/

we however can't use the actual ubuntu version that we run on, as the packagecloud.io repository for legacy varnish do only provide packages for way older ubuntu versions. hardcoding trusty seems to work \o/


dbu:
static analysis will fail until https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/6238 is fixed

PR #520, opened 2022-01-23
Add Symfony 6 compatibility

Korbeil:
Hello :wave: 

I would like to migrate to Symfony 6 and this is the last package I have to update, so here is a quick contribution to help this :)

### DIFF
--- a/src/CacheInvalidator.php
+++ b/src/CacheInvalidator.php
@@ -26,6 +26,7 @@ use FOS\HttpCache\ProxyClient\Symfony;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\LegacyEventDispatcherProxy;
+use Symfony\Component\HttpKernel\Kernel;
 use Toflar\Psr6HttpCacheStore\Psr6Store;
 
 /**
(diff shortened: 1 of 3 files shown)


Korbeil:
About `PHP 7.2 Varnish 6` test, I don't really understand the relation with my changes.
I'll check other ones :)

dbu:
we need to fix the PHP 8 failures. at least some seem related to the argument order switch in the symfony event handler. afaik at least symfony 5 (and maybe 4.4?) is forward compatible. i suggest to switch and remove symfony 3.4 (and 4, if necessary) from the supported versions - the lowest version build should tell if it is fine or not.

please ignore the phpstan complaint, that is a bug in the newest version. i thik the error with php 7.2 should not be happening, lets see if it was a glitch when you push to this branch, and if not i will try to fix it on the master branch.

dbu:
AND: thanks for the contribution!

Korbeil:
@dbu It seems good for most part except the `PHP 7.2 Varnish 6` that still have the same strange issue, could you review when you have some free time ? thanks !

src/CacheInvalidator.php
```diff
@@ -321,8 +322,7 @@ public function flush()
 
     private function dispatch(Event $event, $eventName)
     {
-        // LegacyEventDispatcherProxy exists in Symfony >= 4.3
-        if (class_exists(LegacyEventDispatcherProxy::class)) {
+        if ((Kernel::MAJOR_VERSION === 4 && Kernel::MINOR_VERSION >= 3) || Kernel::MAJOR_VERSION >= 5) {
```
dbu, line 325: the kernel is only in the dev dependencies, so we can't access it here.

but as you raise the minimum version of the event dispatcher to 4.3 anyways, i think we can simply drop the check and always do the new way, no?

dbu (approved):
thanks. i think we can just remove the legacy support from the code entirely.

the weird build failure i will address in a separate MR

src/CacheInvalidator.php
Korbeil: Sure :+1:

dbu:
thanks a lot!

dbu:
https://github.com/FriendsOfSymfony/FOSHttpCache/releases/tag/2.13.0

PR #522, opened 2022-01-25
cleanup build matrix

dbu:


.github/workflows/ci.yml
```diff
@@ -9,28 +9,25 @@ on:
 env:
   VARNISH_VERSION: '6.6'
   VARNISH_MODULES_VERSION: '0.18.0'
-  DEPENDENCIES: 'toflar/psr6-symfony-http-cache-store:^2.2.0|^3.0.0'
+  DEPENDENCIES: 'toflar/psr6-symfony-http-cache-store:^2|^3|^4'
```
dbu, line 12: @Toflar when i look at the php 8 builds, we seem to skip the tests. although we clearly install the psr6 store. i am a bit lost as to why that happens, do you have an idea?

dbu: oh well. thanks for spotting that one! i changed to use the class to be consistent with the rest (does not much matter which, we care whether the package is installed or not)

PR #529, opened 2022-07-02
Update method signature for CacheInvalidation.php

### LINKED ISSUE #528: 2.13.0 is not fully compatible with Symfony 6
Hi,

I have an issue as I trying to upgrade Symfony components, but it seems some FOSHttpCache components not fully compatible yet.

[CacheInvalidation](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/src/SymfonyCache/CacheInvalidation.php)

is not compatible with

[HttpCache](https://github.com/symfony/http-kernel/blob/6.0/HttpCache/HttpCache.php)

```php
// CacheInvalidation.php
public function fetch(Request $request, $catch = false);

// HttpCache.php
protected function fetch(Request $request, bool $catch = false): Response
```

This causing, that when I try to implement `CacheInvalidation` while extending `HttpCache`, to get a compatibility error on the method's signature:

```php
class AppCache extends HttpCache implements CacheInvalidation {
    public function fetch(Request $request, bool $catch = false): Response
    {
    }
}
```

Error:
```
In AppCache.php line 62:
                                                                                                                                                                                  
Declaration of X\AppCache::fetch(Symfony\Component\HttpFoundation\Request $request, bool $catch = false): Symfony\Component\HttpFoundation\  
  Response must be compatible with FOS\HttpCache\SymfonyCache\CacheInvalidation::fetch(Symfony\Component\HttpFoundation\Request $request, $catch = false)                                                                                                                                                                                              
```

Can you please update the signature?

netwarex:
Fixes #528

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -2,6 +2,8 @@ parameters:
     level: 1
     paths:
         - src
+    bootstrapFiles:
+        - .phpstan/classAliases.php
     excludePaths:
         analyseAndScan:
             # contains code to support legacy phpunit versions
--- a/src/SymfonyCache/CacheInvalidation.php
+++ b/src/SymfonyCache/CacheInvalidation.php
@@ -11,18 +11,17 @@
 
 namespace FOS\HttpCache\SymfonyCache;
 
-use Symfony\Component\HttpKernel\Kernel;
-use Symfony\Component\HttpKernel\HttpKernelInterface;
-
 use function class_alias;
 use function class_exists;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\HttpKernel\Kernel;
 
 if (interface_exists(CacheInvalidation::class)) {
     return;
 }
 
 /*
- * Symfony 6 introduced a BC break in the signature of the protected method HttpKernelInterface::fetch. 
+ * Symfony 6 introduced a BC break in the signature of the protected method HttpKernelInterface::fetch.
  * Load the correct interface to match the signature.
  */
 if (class_exists(Kernel::class) && Kernel::MAJOR_VERSION >= 6) {
@@ -39,9 +38,9 @@ if (class_exists(Kernel::class) && Kernel::MAJOR_VERSION >= 6) {
     );
 }
 
-if (! interface_exists(CacheInvalidation::class)) {
+if (!interface_exists(CacheInvalidation::class)) {
     /**
-     * Provide an empty interface for code scanners
+     * Provide an empty interface for code scanners.
      */
     interface SearchHandler extends HttpKernelInterface
     {
--- a/src/SymfonyCache/CacheInvalidation.php
+++ b/src/SymfonyCache/CacheInvalidation.php
@@ -41,5 +41,5 @@ interface CacheInvalidation extends HttpKernelInterface
      *
      * @return StoreInterface $store The store used by the HttpCache
      */
-    public function getStore(): StoreInterface
+    public function getStore(): StoreInterface;
 }
--- a/src/SymfonyCache/CacheInvalidation.php
+++ b/src/SymfonyCache/CacheInvalidation.php
@@ -11,35 +11,32 @@
 
 namespace FOS\HttpCache\SymfonyCache;
 
-use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\HttpCache\StoreInterface;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
 
-/**
- * Interface for a HttpCache that supports active cache invalidation.
- */
-interface CacheInvalidation extends HttpKernelInterface
-{
-    /**
-     * Forwards the Request to the backend and determines whether the response should be stored.
-     *
-     * This methods is triggered when the cache missed or a reload is required.
-     *
-     * This method is present on HttpCache but must be public to allow event listeners to do
-     * refresh operations.
-     *
-     * @param Request $request A Request instance
-     * @param bool    $catch   Whether to process exceptions
-     *
-     * @return Response A Response instance
-     */
-    public function fetch(Request $request, bool $catch = false): Response;
+use function class_alias;
+use function class_exists;
 
+if (! interface_exists(CacheInvalidation::class) && 
+    class_exists('Symfony\Component\HttpKernel\Kernel') &&
+    \Symfony\Component\HttpKernel\Kernel::MAJOR_VERSION >= 6) {
+    // Load class for Symfony >=6.0
+    class_alias(
+        Compatibility\CacheInvalidationS6::class,
+        CacheInvalidation::class
+    );
+} else {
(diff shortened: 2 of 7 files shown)


dbu:
hi @netwarex , thanks for looking into this!

looks like this is quite the BC break that symfony dropped onto us here... i think we can not be compatible with both symfony < 6 and >= 6 at the same time. and changing the signature on our side also is a BC break.

probably we need to do an ugly thing again: wrap the whole class in an if/else checking for Kernel::MAJOR_VERSION, like we do here: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/91f5e625e0b8c5e667a43849a44a7f34042b6aab/src/Security/Http/Logout/ContextInvalidationSessionLogoutHandler.php (but we should also check for class_exists Kernel and if it does not exist go with the old version, to avoid messing up things that scan the code. the symfony kernel is an optional dependency for the FOSHttpCache)

we should then note in the changelog that for symfony 6, you need to adjust to the symfony BC break.

i think for the implementations of the interface that we have in the tests, we can go with the more restrictive thing, as that is allowed for extending classes.

can you try to adjust the PR accordingly?

netwarex:
> can you try to adjust the PR accordingly?

Sure! I have changed to make it backwards compatible. Can you please review my changes, and release if it's passed? :) Thank you!

src/SymfonyCache/CacheInvalidation.php
```diff
@@ -11,35 +11,32 @@
...
-     * @param Request $request A Request instance
-     * @param bool    $catch   Whether to process exceptions
-     *
-     * @return Response A Response instance
-     */
-    public function fetch(Request $request, $catch = false);
+use function class_alias;
+use function class_exists;
+
+if (! interface_exists(CacheInvalidation::class) &&
```
dbu, line 19: i'd prefer to do early return here:

```
if (interface_exists(CacheInvalidation::class)) {
    return;
}
```

src/SymfonyCache/CacheInvalidation.php
```diff
@@ -11,35 +11,32 @@
...
-     * @param bool    $catch   Whether to process exceptions
-     *
-     * @return Response A Response instance
-     */
-    public function fetch(Request $request, $catch = false);
+use function class_alias;
+use function class_exists;
+
+if (! interface_exists(CacheInvalidation::class) && 
+    class_exists('Symfony\Component\HttpKernel\Kernel') &&
```
dbu, line 20: please use a `use` statement and the `::class` pseudoconstant for this as well. like with the interface, it works to use a ::class even for things that don't exist.

src/SymfonyCache/Compatibility/CacheInvalidationLegacy.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace FOS\HttpCache\SymfonyCache\Compatibility;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\HttpCache\StoreInterface;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+/**
+ * Interface for a HttpCache that supports active cache invalidation.
```
dbu, line 20:
```suggestion
 * Interface for a HttpCache that supports active cache invalidation.
 *
 * Method signature of `fetch` compatible with Symfony 5 and older.
```

src/SymfonyCache/Compatibility/CacheInvalidationS6.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace FOS\HttpCache\SymfonyCache\Compatibility;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\HttpCache\StoreInterface;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+
+/**
+ * Interface for a HttpCache that supports active cache invalidation.
```
dbu, line 20:
```suggestion
 * Interface for a HttpCache that supports active cache invalidation.
 *
 * Method signature of `fetch` compatible with Symfony 6 or newer.
```

src/SymfonyCache/CacheInvalidation.php
```diff
@@ -11,35 +11,32 @@
...
-     *
-     * @param Request $request A Request instance
-     * @param bool    $catch   Whether to process exceptions
-     *
-     * @return Response A Response instance
-     */
-    public function fetch(Request $request, $catch = false);
+use function class_alias;
+use function class_exists;
+
```
dbu, line 18:
```suggestion

/*
 * Symfony 6 introduced a BC break in the signature of the protected method HttpKernelInterface::fetch. 
 * Load the correct interface to match the signature.
 */
```

dbu (commented):
thanks! i only have a few nitpicks, but i think this is correct. i will have a look whats up with the build, quite a few jobs fail but it does not look related to your changes.

could you also write the changelog entry to explain that when implementing the CacheInvalidation interface, one will need to be aware of the Symfony BC break?

dbu:
i fixed the builds except for legacy varnish 5 in #530, can you please rebase on master?

dbu:
dang, i messed up the phpstan src job. can you please rebase again?

and can you please check the failures other than " CI / PHP 7.4 Legacy Varnish 5"? varnish 5 has a known problem to install.

there are some cs fixes necessary, and somehow phpstan seems to choke on the class_alias construction - if you find no solution for that, i am ok to add it to phpstan ignore rule or if that is not possible to have the ci job copy the legacy interface over the selection file and remove the new interface for now.

dbu:
thanks. i will quickly check with phpstan if they have an idea how to handle this.

the codestyle check is broken by your changes, can you please adjust these things: 
https://github.com/FriendsOfSymfony/FOSHttpCache/runs/7198550603?check_suite_focus=true

dbu:
https://github.com/phpstan/phpstan/issues/7582

dbu:
for phpstan: we need to configure the alias in a bootstrap file: https://phpstan.org/user-guide/discovering-symbols#class-aliases

i suggest that we alias to legacy if php < 8.1 and the new one otherwise. that way the build will stay ok when the phpstan task updates to php 8.1 and installs symfony 6.

can you do that or do you want me to take over?

phpstan.neon.dist
```diff
@@ -2,6 +2,8 @@ parameters:
     level: 1
     paths:
         - src
+    bootstrapFiles:
+        - classAliases.php
```
dbu, line 6: thanks! please put the alias file into a `.phpstan` subfolder so we don't pollute the root directory.

and then we can the .phpstan folder to the .gitattributes as export-ignore so installations dont get the file.

dbu (commented):
yay, its getting green!

can we please move the config file into a subfolder (also to make the context of the file clear)? i think then we are ready to merge.

dbu:
https://github.com/FriendsOfSymfony/FOSHttpCache/releases/tag/2.14.0

PR #530, opened 2022-07-05
fix builds

dbu:


dbu:
legacy support currently is broken because of https://github.com/varnish/varnish-modules/issues/200

PR #534, opened 2022-07-06
Fix the CacheInvalidation fix for code scanners

leofeyer:


dbu (approved):
i reviewed #529 but somehow missed that the name was wrong...

thanks for the fix.

dbu:
tagged as 2.14.1

PR #538, opened 2022-11-02
Fix authorization header name

jdreesen:
This is a regression from #497.


dbu (approved):
ouch, indeed. the autor of that regression was me, but nobody authorized me to break it :-/

dbu:
released in https://github.com/FriendsOfSymfony/FOSHttpCache/releases/tag/2.14.2

PR #549, opened 2023-11-29
Allow installation with Symfony 7

alexander-schranz:
Requires another update:

 - [ ] php-http/client-common      2.7.0   requires symfony/options-resolver (~4.0.15 || ~4.1.9 || ^4.2.1 || ^5.0 || ^6.0) https://github.com/php-http/client-common/pull/232

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,16 @@ jobs:
         - php: '8.1'
           varnish-version: '7.1'
           varnish-modules-version: '0.20.0'
+        - php: '8.2'
+          symfony-version: '7.*'
+        - php: '8.2'
+          varnish-version: '7.1'
+          varnish-modules-version: '0.20.0'
+        - php: '8.3'
+          symfony-version: '7.*'
+        - php: '8.3'
+          varnish-version: '7.1'
+          varnish-modules-version: '0.20.0'
 
     steps:
       - name: Setup PHP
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Not sure about the tests as they even fail on Symfony 6. @dbu do you have any idea they seems also failed last time on target branch: https://github.com/FriendsOfSymfony/FOSHttpCache/actions/runs/6640786360/job/18041892112

dbu:
i am fixing the regression with the tests in #550 

unfortunately i noticed that the RequestMatcher disappeared from symfony/http-foundation. we need to find a solution for that, did not yet investigate.

alexander-schranz:
Looks like we need a bridge there between the 2 classes based on the Symfony Upgrade:

```
Remove RequestMatcher, use ChainRequestMatcher instead
Remove ExpressionRequestMatcher, use RequestMatcher\ExpressionRequestMatcher instead
```

but not sure how much compatible they are.

dbu:
i think supporting both should be less difficult than i feared: https://github.com/FriendsOfSymfony/FOSHttpCache/pull/551

dbu:
fixed it in the 2.x branch, can you please rebase?

dbu:
the return type declarations would be a BC break if somebody overwrote the store method our trait defines.

the safe thing is to do a new major version of foshttpcache to add return types.

dbu:
continuing in #552

PR #552, opened 2023-12-11
Sf 7

dbu:


composer.json
```diff
@@ -21,9 +21,9 @@
         }
     ],
     "require": {
-        "php": "^7.2 || ^8.0",
-        "symfony/event-dispatcher": "^4.3 || ^5.0 || ^6.0",
-        "symfony/options-resolver": "^4.3 || ^5.0 || ^6.0",
+        "php": "^8.0",
+        "symfony/event-dispatcher": "^5.4 || ^6.0 || ^7.0",
```
Toflar: May I suggest to go for `^6.4 || ^7.0` only in v3? 6.4 is LTS too so I think you would do yourself a favor in terms of maintenance :)

dbu: yeah, actually lets do that. the package has been quite stable, so i don't think we will have many important fixes for old versions.

PR #553, opened 2024-01-19
Allow empty tag list

ro0NL:
Fixes `array_chunk(): Argument #2 ($length) must be greater than 0` and removes the need for redundant boilerplate `if` everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Test with PHP 8.2 and 8.3
 * Drop support for PHP < 8.1
 * Parameter and return type declarations where possible.
+* Ignore empty tag lists
 
 2.x
 ===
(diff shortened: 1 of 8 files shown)


dbu (approved):
thanks, makes sense to do it like that.

lets make it consistent though: TagCapable is implemented by many clients. lets document on the interface that an empty tag list is ignored, and fix all the other clients as well. can you adjust that?

and can you please add a changelog entry for it, so people are aware of it?

ro0NL:
@dbu done

dbu (approved):
thanks!

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 * Test with PHP 8.2 and 8.3
 * Drop support for PHP < 8.1
 * Parameter and return type declarations where possible.
+* Ignore empty tag lists
```
dbu:
```suggestion
* Ignore empty tag lists passed to `TagCapable::invalidateTags` so you don't need to check if there are tags or not.
```

Issue #555, opened 2024-02-01
`php-http/message-factory` is abandoned 


mvhirsch:
Hey :-)

One of my projects is using `friendsofsyfmony/fos-http-cache-bundle`. I recently noticed composers output: 

```
Package php-http/message-factory is abandoned, you should avoid using it. Use psr/http-factory instead.
```

Obviously this got abandoned: https://packagist.org/packages/php-http/message-factory

What should be the next step? I'm new to HTTPlug and PSR17, but I'm eager to learn.

dbu:
indeed we should switch to PSR-17 (and PSR-18 with php-http/discovery for autoconfiguration if desired).

as we have the not yet released 3.x branch, this is the perfect moment for changes that potentially incur BC breaks. if you are motivated, its awesome if you can attempt the switch. 

in a first step, that would be to remove the php-http/message-factory and require psr/http-factory, then change to use the PSR-17 RequestFactoryInterface (and maybe we need UriFactoryInterface too) instead of the current request factory.

for the discovery part, it means replacing MessageFactoryDiscovery with Psr17FactoryDiscovery (in HttpProxyClient and if there are other places, those too).

i suggest doing only that part in a first step, and only start looking into PSR-18 once we have wrapped up above and have been able to merge it.

mvhirsch:
Thank you for your answer. Sounds simple and easy to me.

I'll provide a PR next week (this week are holidays in Austria) :+1:

toxicity1985:
Hello,

So the pr for the cache bundle is on waiting now ?

dbu:
fixed in #566

mvhirsch:
Thank you @dbu !

PR #556, opened 2024-03-04
Remove deprecated library

### LINKED ISSUE #443: Switch from HTTPlug to PSR-17 and PSR-18
Change the requirements from php-http/httplug to psr/http-client and psr/http-factory and adjust the code accordingly. (new major version)

toxicity1985:
fix #443 #555

### DIFF
--- a/src/ProxyClient/Cloudflare.php
+++ b/src/ProxyClient/Cloudflare.php
@@ -14,7 +14,7 @@ namespace FOS\HttpCache\ProxyClient;
 use FOS\HttpCache\ProxyClient\Invalidation\ClearCapable;
 use FOS\HttpCache\ProxyClient\Invalidation\PurgeCapable;
 use FOS\HttpCache\ProxyClient\Invalidation\TagCapable;
-use Http\Message\RequestFactory;
+use Psr\Http\Message\RequestFactoryInterface;
 
 /**
  * Cloudflare HTTP cache invalidator.
@@ -52,7 +52,7 @@ class Cloudflare extends HttpProxyClient implements ClearCapable, PurgeCapable, 
     public function __construct(
         Dispatcher $httpDispatcher,
         array $options = [],
-        RequestFactory $messageFactory = null
+        RequestFactoryInterface $messageFactory = null
     ) {
         if (!function_exists('json_encode')) {
             throw new \Exception('ext-json is required for cloudflare invalidation');
--- a/src/ProxyClient/HttpDispatcher.php
+++ b/src/ProxyClient/HttpDispatcher.php
@@ -23,11 +23,13 @@ use Http\Client\Exception\HttpException;
 use Http\Client\Exception\NetworkException;
 use Http\Client\HttpAsyncClient;
 use Http\Discovery\HttpAsyncClientDiscovery;
-use Http\Discovery\UriFactoryDiscovery;
+use Http\Discovery\Psr17FactoryDiscovery;
 use Http\Message\UriFactory;
 use Http\Promise\Promise;
 use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\UriFactoryInterface;
 use Psr\Http\Message\UriInterface;
+use Http\Discovery\UriFactoryDiscovery;
 
 /**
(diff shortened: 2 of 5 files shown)


dbu:
thanks for the pull request. this will need quite some refactoring. we need to accept the PSR17 factories (in addition to the legacy factories for BC) in places like the (HttpDispatcher)[https://github.com/FriendsOfSymfony/FOSHttpCache/blob/e446b18c2fe80484684c67ebd1848ec726c8a4ad/src/ProxyClient/HttpDispatcher.php#L104] and use the psr17 discovery (but again for BC should fallback to the legacy discovery if not found)

the tests need to use the new discovery as well, it seems.

for the declared dependencies, i am not sure about removing them in version 2. if somebody instantiates something explicitly, it would break for them once they update.

if we allow the new things while still supporting the old, we can tag a last 2.x release and then start 3.x where we actually drop things and completely move to PSR-17.

if you have some time to work on this, it would be awesome!

toxicity1985:
Hello,
Yes, i was looking more deeper in the library but i don't find my way. 
It's something complicated. I think that some other libray has to be updated but not sure about that.

src/ProxyClient/Cloudflare.php
```diff
@@ -52,7 +52,7 @@ class Cloudflare extends HttpProxyClient implements ClearCapable, PurgeCapable,
     public function __construct(
         Dispatcher $httpDispatcher,
         array $options = [],
-        RequestFactory $messageFactory = null
+        RequestFactoryInterface $messageFactory = null
```
dbu, line 55: to avoid BC breaks, we should remove the type declaration and instead have a phpdoc that does `RequestFactory|RequestFactoryInterface`

src/ProxyClient/HttpDispatcher.php
```diff
@@ -85,14 +87,15 @@ class HttpDispatcher implements Dispatcher
...
-     * @param UriFactory|null      $uriFactory Factory for PSR-7 URIs. If not specified, a
+     * @param UriFactoryInterface|null      $uriFactory Factory for PSR-7 URIs. If not specified, a
      *                                         default one is created
      */
     public function __construct(
         array $servers,
         $baseUri = '',
         HttpAsyncClient $httpClient = null,
-        UriFactory $uriFactory = null
+        Psr17FactoryDiscovery $uriFactory = null,
```
dbu, line 97: and this needs a phpdoc `UriFactory|Psr17FactoryDiscovery` and our code needs to handle both cases

src/ProxyClient/HttpDispatcher.php
```diff
@@ -101,14 +104,19 @@ public function __construct(
...
+            $this->uriFactory = $uriFactory ?: $legacyUriFactory::find();
+        }
 
         $this->setServers($servers);
         $this->setBaseUri($baseUri);
     }
 
     public function invalidate(RequestInterface $invalidationRequest, $validateHost = true)
     {
+        var_dump('ttttt');
```
dbu: don't forget to remove this when debugging is done

src/ProxyClient/HttpProxyClient.php
```diff
@@ -50,17 +50,17 @@ abstract class HttpProxyClient implements ProxyClient
...
     public function __construct(
         Dispatcher $httpDispatcher,
         array $options = [],
-        RequestFactory $messageFactory = null
+        RequestFactoryInterface $messageFactory = null
     ) {
         $this->httpDispatcher = $httpDispatcher;
         $this->options = $this->configureOptions()->resolve($options);
-        $this->requestFactory = $messageFactory ?: MessageFactoryDiscovery::find();
+        $this->requestFactory = $messageFactory ?: Psr17FactoryDiscovery::findRequestFactory();
```
dbu, line 64: we should fallback to MessageFactoryDiscovery::find if the Psr17FactoryDiscovery does not find anything.

dbu (commented):
thanks for looking into this!

dbu:
wrapped up in #566

because we do a new major version, i decided to not provide BC with the old factories. anyways most legacy factories  at the same time implement the psr factory as well.

PR #565, opened 2024-03-25
refactor to phpunit 10 event system

### LINKED ISSUE #561: replace WebServerListener with phpunit extensions
see https://docs.phpunit.de/en/10.5/extending-phpunit.html#implementing-an-extension

dbu:
fix #561


src/Test/WebServerSubscriber.php
```diff
@@ -11,49 +11,51 @@
...
-    public function startTestSuite($suite)
+    public function notify(ExecutionStarted $event): void
     {
-        // Only run on PHP >= 5.4 as PHP below that and HHVM don't have a
-        // built-in web server
-        if (defined('HHVM_VERSION')) {
-            return;
-        }
-
-        if (null !== $this->pid || !in_array('webserver', $suite->getGroups())) {
```
dbu, line 38: we used to check the groups of the test suite, but that seems not available with phpunit 10. is there a way to know which groups a test suite includes?

dbu:
if https://github.com/sebastianbergmann/phpunit/issues/5769 gets solved, we can improve the setup.

meanwhile i will merge, the overhead of starting `php -S` seems to be small.

composer.json
```diff
@@ -37,11 +37,11 @@
         "php-http/mock-client": "^1.6.0",
         "symfony/process": "^6.4|| ^7.0",
         "symfony/http-kernel": "^6.4|| ^7.0",
-        "phpunit/phpunit": "^9"
+        "phpunit/phpunit": "^10.5"
     },
     "conflict": {
         "toflar/psr6-symfony-http-cache-store": "<2.2.1",
-        "phpunit/phpunit": "<8"
+        "phpunit/phpunit": "<10"
```
alexander-schranz, line 44: It is strange that I can not use the `3.x` package when my PHPUnit version is `9.6` which is still a widely. used version of PHPUnit.

I would like to avoid this conflict think most of the people will not require a test besed on the `WebServerSubscriber` and so this should only be a softrequirement.

dbu: i think you are right. 

fixed in 24a47d586b81a7a9a21fe8aad88b20948e6856cf

PR #567, opened 2024-03-28
use fixed discovery instead of declaring conflict with old guzzlehttp…

dbu:
…/psr

the prefer-lowest job goes into an endless loop, see https://github.com/php-http/discovery/pull/263

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,8 +40,7 @@
         "phpunit/phpunit": "^10.5"
     },
     "conflict": {
-        "toflar/psr6-symfony-http-cache-store": "<2.2.1",
-        "guzzlehttp/psr7": "<2"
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
         "symfony/event-dispatcher": "^6.4 || ^7.0",
         "symfony/options-resolver": "^6.4 || ^7.0",
         "php-http/client-common": "^1.1.0 || ^2.0",
-        "php-http/discovery": "^1.12",
+        "php-http/discovery": "dev-fix-interop-stickyness as 1.16.0",
```
dbu: do not merge like this. depend on release once discovery

nicolas-grekas:
My reco here would be to disable the plugin for the CI.

PR #575, opened 2024-07-22
Add fallbackToSmaxage flag to CustomTtlListener and check for isCacheable before calling store

alexander-schranz:
We have the `CustomTtlListener` enabled but we have controllers which set only a `s-maxage` the `s-maxage` is in our case always only for browsers and should never used by Symfony or Varnished HttpCache / reverse proxy.

The current `CustomTtlListener` still fallbacks always back to s-maxage / maxage this should in my opinion not the case if `CustomTtlListener` the HttpCache should only listen on the configured header and ignore the max age headers.

/cc @Toflar @dbu Alternativily we should also add a option to the `CustomTtlListener` `disableFallback` / `enableFallback`.

**Update**

As discussed below we will add additional `fallbackToSmaxage = true` parameter to make it possible to disable that fallback mechanism.

Additional this pull request adds a recheck for `$response->isCachable` after the listeners were run. That not only improve performance for response setting a ttl to 0 but its now possible to set a response to private in a preHandle listener event to also disable the caching.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 ------
 
 * Add flag to `CustomTtlListener` for disabling fallback to s-maxage if header not defined.
-* Fix not calling store if Response object is not longer cachable after event listeners.
+* Fix not calling store if Response object is not longer cacheable after event listeners.
 
 2.15.3
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 ------
 
 * Add flag `fallbackToSmaxage` to `CustomTtlListener` to allow controlling fallback to s-maxage if custom TTL header is not defined on the response.
-* Fix: Do not call store if Response object is not longer cacheable after event listeners.
+* Fix for Symfony HttpCache: Do not call store if Response object is not longer cacheable after event listeners. If you use the custom TTL system, this is only a performance improvement, because the TTL of the response would still be 0. With a custom listener that changes the response explicitly to not be cached but does not change s-maxage, this bug might have led to caching responses that should not have been cached
 
 2.15.3
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 2.15.4
 ------
 
-* Fix always use defined `CustomTtlListener` header even if not set.
+* Add flag to `CustomTtlListener` for disabling fallback to s-maxage if header not defined.
 * Fix not calling store if Response object is not longer cachable after event listeners.
 
 2.15.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@ See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpC
 2.15.4
 ------
 
-* Add flag to `CustomTtlListener` for disabling fallback to s-maxage if header not defined.
-* Fix not calling store if Response object is not longer cacheable after event listeners.
+* Add flag `fallbackToSmaxage` to `CustomTtlListener` to allow controlling fallback to s-maxage if custom TTL header is not defined on the response.
+* Fix: Do not call store if Response object is not longer cacheable after event listeners.
 
 2.15.3
 ------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+2.15.4
+------
+
+* Fix always use defined `CustomTtlListener` header even if not set.
+* Fix not calling store if Response object is not longer cachable after event listeners.
+
 2.15.3
 ------
 
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -360,7 +360,7 @@ You can enable keeping the custom header with the `keepTtlHeader` parameter::
 
     new CustomTtlListener('My-TTL-Header', keepTtlHeader: true);
 
-By default if the custom ttl header is not found it will fallback to s-maxage.
+By default if the custom ttl header is not present, the listener falls back to the s-maxage cache-control directive.
 To disable this behavior you can set the `fallbackToSmaxage` parameter to false::
 
     new CustomTtlListener('My-TTL-Header', keepTtlHeader: true, fallbackToSmaxage: false);
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -356,6 +356,14 @@ but you can customize that in the listener constructor::
     new CustomTtlListener('My-TTL-Header');
 
 The custom header is removed before sending the response to the client.
+You can enable keeping the custom header with the `keepTtlHeader` parameter::
+
+    new CustomTtlListener('My-TTL-Header', keepTtlHeader: true);
+
+By default if the custom ttl header is not found it will fallback to s-maxage.
+To disable this behavior you can set the `fallbackToSmaxage` parameter to false::
+
+    new CustomTtlListener('My-TTL-Header', keepTtlHeader: true, fallbackToSmaxage: false);
 
 .. _symfony-cache x-debugging:
 
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -361,7 +361,7 @@ You can enable keeping the custom header with the `keepTtlHeader` parameter::
     new CustomTtlListener('My-TTL-Header', keepTtlHeader: true);
 
 By default if the custom ttl header is not present, the listener falls back to the s-maxage cache-control directive.
-To disable this behavior you can set the `fallbackToSmaxage` parameter to false::
+To disable this behavior, you can set the `fallbackToSmaxage` parameter to false::
 
     new CustomTtlListener('My-TTL-Header', keepTtlHeader: true, fallbackToSmaxage: false);
 
--- a/doc/symfony-cache-configuration.rst
+++ b/doc/symfony-cache-configuration.rst
@@ -363,7 +363,7 @@ You can enable keeping the custom header with the `keepTtlHeader` parameter::
 By default if the custom ttl header is not present, the listener falls back to the s-maxage cache-control directive.
 To disable this behavior, you can set the `fallbackToSmaxage` parameter to false::
 
-    new CustomTtlListener('My-TTL-Header', keepTtlHeader: true, fallbackToSmaxage: false);
+    new CustomTtlListener('My-TTL-Header', fallbackToSmaxage: false);
 
 .. _symfony-cache x-debugging:
 
--- a/src/SymfonyCache/CustomTtlListener.php
+++ b/src/SymfonyCache/CustomTtlListener.php
@@ -34,6 +34,11 @@ class CustomTtlListener implements EventSubscriberInterface
     private $keepTtlHeader;
 
     /**
+     * @var bool
+     */
+    private $disableFallback;
+
+    /**
      * Header used for backing up the s-maxage.
      *
      * @var string
@@ -41,13 +46,15 @@ class CustomTtlListener implements EventSubscriberInterface
     public const SMAXAGE_BACKUP = 'FOS-Smaxage-Backup';
 
     /**
-     * @param string $ttlHeader     The header name that is used to specify the time to live
-     * @param bool   $keepTtlHeader Keep the custom TTL header on the response for later usage (e.g. debugging)
+     * @param string $ttlHeader       The header name that is used to specify the time to live
+     * @param bool   $keepTtlHeader   Keep the custom TTL header on the response for later usage (e.g. debugging)
+     * @param bool   $disableFallback If the custom TTL header is not set, do not fall back to s-maxage
      */
-    public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader = false)
+    public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader = false, $disableFallback = false)
     {
         $this->ttlHeader = $ttlHeader;
         $this->keepTtlHeader = $keepTtlHeader;
+        $this->disableFallback = $disableFallback;
     }
 
     /**
@@ -59,6 +66,13 @@ class CustomTtlListener implements EventSubscriberInterface
     public function useCustomTtl(CacheEvent $e)
     {
         $response = $e->getResponse();
+
+        if (!$response->headers->has($this->ttlHeader)
+            && false === $this->disableFallback
+        ) {
+            return;
+        }
+
         $backup = $response->headers->hasCacheControlDirective('s-maxage')
             ? $response->headers->getCacheControlDirective('s-maxage')
             : 'false'
--- a/src/SymfonyCache/CustomTtlListener.php
+++ b/src/SymfonyCache/CustomTtlListener.php
@@ -46,8 +46,8 @@ class CustomTtlListener implements EventSubscriberInterface
     public const SMAXAGE_BACKUP = 'FOS-Smaxage-Backup';
 
     /**
-     * @param string $ttlHeader       The header name that is used to specify the time to live
-     * @param bool   $keepTtlHeader   Keep the custom TTL header on the response for later usage (e.g. debugging)
+     * @param string $ttlHeader         The header name that is used to specify the time to live
+     * @param bool   $keepTtlHeader     Keep the custom TTL header on the response for later usage (e.g. debugging)
      * @param bool   $fallbackToSmaxage If the custom TTL header is not set, should s-maxage be used?
      */
     public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader = false, $fallbackToSmaxage = true)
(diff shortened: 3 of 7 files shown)


src/SymfonyCache/EventDispatchingHttpCache.php
```diff
@@ -109,7 +109,11 @@ protected function store(Request $request, Response $response)
     {
         $response = $this->dispatch(Events::PRE_STORE, $request, $response);
 
-        parent::store($request, $response);
+        // CustomTtlListener or other Listener or Subscribers might have set a non-cacheable response so we need revalidate this
+        // So store is only called when cacheable like Symfony core would do: https://github.com/symfony/symfony/blob/v7.1.2/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L409
+        if ($response->isCacheable()) {
+            parent::store($request, $response);
+        }
```
alexander-schranz: Current store was still called when the CustomTtlListener did set a cachelifetime of 0.

This specially can be seen if you have a empty @toflar PSR6Store.php used the `write_operations` counter still increase here: https://github.com/Toflar/psr6-symfony-http-cache-store/blob/9b3eb5b5f449a886b3ce202ddf3705e3d2ac79c9/src/Psr6Store.php#L459.

The store should only be called when the response still is cacheable after the events where dispatched. Symfony also checks for `isCacheable` here: https://github.com/symfony/symfony/blob/v7.1.2/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L409

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -59,15 +59,16 @@ public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader =
     public function useCustomTtl(CacheEvent $e)
     {
         $response = $e->getResponse();
-        if (!$response->headers->has($this->ttlHeader)) {
-            return;
-        }
```
alexander-schranz: This currently behaves that `s-maxage` wins when a CustomTtlListener is enabled but header not defined. This I think is not the expected behaviour of the CustomttlListener and leads so to unexpected responses being cached (in our case big media files).

src/SymfonyCache/EventDispatchingHttpCache.php
dbu: i agree with this change, the event handlers should be able to change the decision.

src/SymfonyCache/CustomTtlListener.php
dbu: that is by design. if s-maxage is set, any random proxies between the client and your server might cache the response. why would you specifically not want to cache in your reverse proxy, but in uncontrolled proxies somewhere else?

you mention large media files you don't want to cache: for that use case, i would either write your own listener that checks the size and disables caching for the response, or have the application send a custom ttl of 0. i would recommend the custom listener, as it would logically be okay to cache the response, but from a practical point of view not suitable. the drawback would be that you have no control of telling in the cache layer whether the response was expensive to generate or not. (btw, varnish cache allows to have 2 separate cache stores for small and large responses to avoid large responses bombing out bunches of small ones.). and another btw: if you deliver large media files from the filesystem (e.g. for access control) you could look into x-accel / x-sendfile to make your webserver do the heavy lifting and avoid piping binary data through php. that would also have the side effect of being able to cache the response again, as the symfony cache happens before the web server resolves x-accel.

alexander-schranz: > Why would you specifically not want to cache in your reverse proxy, but in uncontrolled proxies somewhere else?

Because the s-maxage is in our case for third party proxies but not for our own caches. Think https://github.com/symfony/symfony/issues/47288 would still be best solution for such cases to control what layer caches what things.

The media are not files in this case and so we can not work with the BinaryFileResponse of Symfony in this case / x-sendfile.

---

What do you think about an additional flag for the `CustomTtlListener`? See my latest changes.

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+2.15.4
+------
+
+* Add flag to `CustomTtlListener` for disabling fallback to s-maxage if header not defined.
```
dbu:
```suggestion
* Add flag `fallbackToSmaxage` to `CustomTtlListener` to allow controlling fallback to s-maxage if custom TTL header is not defined on the response.
```

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ Changelog
...
+* Fix not calling store if Response object is not longer cacheable after event listeners.
```
dbu: what exactly was the behaviour before? did it actually store, or is this only a performance optimization?

```suggestion
* Fix: Do not call store if Response object is not longer cacheable after event listeners.
```

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -33,6 +33,11 @@ class CustomTtlListener implements EventSubscriberInterface
      */
     private $keepTtlHeader;
 
+    /**
+     * @var bool
+     */
+    private $disableFallback;
```
dbu: negative formulations can be confusing, i prefer a positive formulation

```suggestion
    private $fallbackToSmaxage;
```

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -41,13 +46,15 @@ class CustomTtlListener implements EventSubscriberInterface
     public const SMAXAGE_BACKUP = 'FOS-Smaxage-Backup';
 
     /**
-     * @param string $ttlHeader     The header name that is used to specify the time to live
-     * @param bool   $keepTtlHeader Keep the custom TTL header on the response for later usage (e.g. debugging)
+     * @param string $ttlHeader       The header name that is used to specify the time to live
+     * @param bool   $keepTtlHeader   Keep the custom TTL header on the response for later usage (e.g. debugging)
+     * @param bool   $disableFallback If the custom TTL header is not set, do not fall back to s-maxage
```
dbu:
```suggestion
     * @param bool   $fallbackToSmaxage If the custom TTL header is not set, should s-maxage be used?
```

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -41,13 +46,15 @@ class CustomTtlListener implements EventSubscriberInterface
...
      */
-    public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader = false)
+    public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader = false, $disableFallback = false)
```
dbu: we should have used an `$options` array. this becomes a bit cumbersome with several options. then again, from the bundle it works just fine.

```suggestion
    public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader = false, $fallbackToSmaxage = true)
```

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -41,13 +46,15 @@ class CustomTtlListener implements EventSubscriberInterface
...
     {
         $this->ttlHeader = $ttlHeader;
         $this->keepTtlHeader = $keepTtlHeader;
+        $this->disableFallback = $disableFallback;
```
dbu:
```suggestion
        $this->fallbackToSmaxage = $fallbackToSmaxage;
```

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -59,15 +66,23 @@ public function __construct($ttlHeader = 'X-Reverse-Proxy-TTL', $keepTtlHeader =
     public function useCustomTtl(CacheEvent $e)
     {
         $response = $e->getResponse();
-        if (!$response->headers->has($this->ttlHeader)) {
+
+        if (!$response->headers->has($this->ttlHeader)
+            && false === $this->disableFallback
```
dbu:
```suggestion
            && true === $this->fallbackToSmaxage
```

src/SymfonyCache/EventDispatchingHttpCache.php
```diff
@@ -109,7 +109,11 @@ protected function store(Request $request, Response $response)
     {
         $response = $this->dispatch(Events::PRE_STORE, $request, $response);
 
-        parent::store($request, $response);
+        // CustomTtlListener or other Listener or Subscribers might have set a non-cacheable response so we need revalidate this
```
dbu:
```suggestion
        // CustomTtlListener or other Listener or Subscribers might have changed the response to become non-cacheable, revalidate.
```

src/SymfonyCache/EventDispatchingHttpCache.php
```diff
@@ -109,7 +109,11 @@ protected function store(Request $request, Response $response)
...
+        // So store is only called when cacheable like Symfony core would do: https://github.com/symfony/symfony/blob/v7.1.2/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L409
```
dbu:
```suggestion
        // Only call store for a cacheable response like Symfony core does: https://github.com/symfony/symfony/blob/v7.1.2/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L409
```

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -89,12 +99,15 @@ public function cleanResponse(CacheEvent $e)
             } else {
                 $response->headers->addCacheControlDirective('s-maxage', $smaxage);
             }
+
+            $response->headers->remove(static::SMAXAGE_BACKUP);
         }
 
-        if (!$this->keepTtlHeader) {
+        if ($response->headers->has($this->ttlHeader)
```
dbu: are you sure about checking for existence? removing a header that does not exist is not wrong - and looking at the code i am not sure which is more efficient.

tests/Unit/SymfonyCache/CustomTtlListenerTest.php
```diff
@@ -87,6 +87,23 @@ public function testNoCustomTtl()
         $this->assertFalse($response->headers->has(CustomTtlListener::SMAXAGE_BACKUP));
     }
 
+    public function testNoCustomTtlNoFallback()
+    {
+        $ttlListener = new CustomTtlListener('X-Reverse-Proxy-TTL', false, true);
```
dbu:
```suggestion
        $ttlListener = new CustomTtlListener('X-Reverse-Proxy-TTL', false, false);
```
(adjust to my inversion of the flag)

dbu (commented):
thanks. we can do the extra flag - though setting custom ttl to 0 should have the same effect.

can you explain the flag in the documentation as well, please?
https://github.com/FriendsOfSymfony/FOSHttpCache/blob/3.x/doc/symfony-cache-configuration.rst#custom-ttl

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -89,12 +99,15 @@ public function cleanResponse(CacheEvent $e)
...
+            && !$this->keepTtlHeader
+        ) {
```
alexander-schranz:
```suggestion
        if (!$this->keepTtlHeader) {
```

src/SymfonyCache/CustomTtlListener.php
alexander-schranz: I just moved the if down as the early return was confusing for me as I first did read it as `||` instead of `&&` so I removed the early return and moved the has check here. But you are right removing it think would not hurt and I'm not sure what is faster.

doc/symfony-cache-configuration.rst
```diff
@@ -356,6 +356,14 @@ but you can customize that in the listener constructor::
     new CustomTtlListener('My-TTL-Header');
 
 The custom header is removed before sending the response to the client.
+You can enable keeping the custom header with the `keepTtlHeader` parameter::
+
+    new CustomTtlListener('My-TTL-Header', keepTtlHeader: true);
+
+By default if the custom ttl header is not found it will fallback to s-maxage.
```
dbu:
```suggestion
By default if the custom ttl header is not present, the listener falls back to the s-maxage cache-control directive.
```

doc/symfony-cache-configuration.rst
```diff
@@ -356,6 +356,14 @@ but you can customize that in the listener constructor::
...
+To disable this behavior you can set the `fallbackToSmaxage` parameter to false::
```
dbu:
```suggestion
To disable this behavior, you can set the `fallbackToSmaxage` parameter to false::
```

doc/symfony-cache-configuration.rst
```diff
@@ -356,6 +356,14 @@ but you can customize that in the listener constructor::
...
+
+    new CustomTtlListener('My-TTL-Header', keepTtlHeader: true, fallbackToSmaxage: false);
```
dbu: when using named parameter, we don't need the one we are not concerned with, i think?
```suggestion
    new CustomTtlListener('My-TTL-Header', fallbackToSmaxage: false);
```

doc/symfony-cache-configuration.rst
```diff
@@ -356,6 +356,14 @@ but you can customize that in the listener constructor::
     new CustomTtlListener('My-TTL-Header');
 
 The custom header is removed before sending the response to the client.
+You can enable keeping the custom header with the `keepTtlHeader` parameter::
+
+    new CustomTtlListener('My-TTL-Header', keepTtlHeader: true);
```
dbu, line 361: thanks for this one :+1:

src/SymfonyCache/CustomTtlListener.php
```diff
@@ -89,12 +99,13 @@ public function cleanResponse(CacheEvent $e)
             } else {
                 $response->headers->addCacheControlDirective('s-maxage', $smaxage);
             }
+
+            $response->headers->remove(static::SMAXAGE_BACKUP);
```
dbu, line 103: ah, i see. yeah agree to move the removal in here, because clearly we don't need to remove if it is not present :joy: 

but good to keep removing the ttl header without checking for its existence.

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+2.15.4
+------
+
+* Add flag `fallbackToSmaxage` to `CustomTtlListener` to allow controlling fallback to s-maxage if custom TTL header is not defined on the response.
+* Fix: Do not call store if Response object is not longer cacheable after event listeners.
```
dbu: how severe was this bug? did we actually end up storing uncacheable things? did we also deliver them?

that is: did this bug lead to serving data that should not have been cached? or is it "just" a performance fix?

alexander-schranz: > how severe was this bug? did we actually end up storing uncacheable things? did we also deliver them?

It didn't have any effects or bugs to the end user or the cached items as there was no cachelifetime for this items and so they weren't stored. I did not see that any files where created.

It only did trigger a unnecessary write operations in the PSR6Store, as there is a counter for a autoPrune behaviour here: https://github.com/Toflar/psr6-symfony-http-cache-store/blob/9b3eb5b5f449a886b3ce202ddf3705e3d2ac79c9/src/Psr6Store.php#L459 which is normally only called when public responses are cached. Now it was in our case also for "private" endpoints called and so autpruning happened a lot of more times and this unnecessary write counter increases.

So at the end this specific case is only performance improvement.

dbu: thanks for the clarification. what if a listener changes other cache-control directives like adding a no-store?

i suggest we are a bit more detailed to not unnecessairy scare people looking at the changelog

alexander-schranz: @dbu That is a good point. Calling `$response->setPrivate();` or `$response->headers->addCacheControlDirective('no-store')` inside a PreHandle Listener behaviour is now changed:

Before: `$response->setPrivate()` did not have any effects and the response was still cached if it was cacheable before.

After (with this change): `$response->setPrivate();` will make it uncacheable also for the http cache.

Same for no-store part. They were only validate before the listeners but never after the listeners were called: https://github.com/symfony/symfony/blob/04f9e6dc07cc228fb3023579c0a63a6ae06a4674/src/Symfony/Component/HttpFoundation/Response.php#L544.

So making a cacheable response uncachable didn't work before. But I did not find any builtin listener which is modifying that even the UserContextListener does not, they handle all via `Vary` header correctly and so keep the cache-control public as they should.

alexander-schranz: I would expect if I set inside a preHandle listener a response to private I don't want to cache it. Else if I still want to cache it I would call the setPrivate on postHandle listener instead.

In this case it's not only a performance fix but also a fix or change to support making responses private / nocacheable inside preHandle listener.

dbu: you are right. lets say it like this?

```suggestion
* Fix for Symfony HttpCache: Do not call store if Response object is not longer cacheable after event listeners. If you use the custom TTL system, this is only a performance improvement, because the TTL of the response would still be 0. With a custom listener that changes the response explicitly to not be cached but does not change s-maxage, this bug might have led to caching responses that should not have been cached
```

alexander-schranz: @dbu sounds good 👍

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
+2.15.4
+------
+
+* Add flag `fallbackToSmaxage` to `CustomTtlListener` to allow controlling fallback to s-maxage if custom TTL header is not defined on the response.
```
alexander-schranz, line 9:
```suggestion
2.16.0
------

* Add flag `fallbackToSmaxage` to `CustomTtlListener` to allow controlling fallback to s-maxage if custom TTL header is not defined on the response.
```

Do you want to tag this as a new minor with the new flag? @dbu

dbu: yes. i am editing the changelog locally and will squash the commits.

alexander-schranz: 👍

dbu:
squashed and merged as 6957deed6a2190159f35466ccb2c1702b0a80803

PR #577, opened 2024-07-23
fix custom ttl with symfony HttpCache to work regardless of s-maxage

### LINKED ISSUE #576: custom ttl should work even without s-maxage
As discussed in https://github.com/sulu/sulu/issues/7441, the CustomTtlListener only gets active if s-maxage is > 0. 

https://github.com/symfony/symfony/blob/04f9e6dc07cc228fb3023579c0a63a6ae06a4674/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L408-L410 decides whether to call `store` at all.

We need the listener to rewrite s-maxage before the store decision happens, so PRE_STORE is too late. the best option i can see is to have events around `HttpCache::forward`. changing the caching rules in POST_FORWARD should be the right spot to get the response cached. while we are at it, we should also do a PRE_FORWARD to allow users to modify the request before giving it to symfony if they want to.

to validate: does the control flow look the same in older symfony versions?

dbu:
fix #576

### DIFF
--- a/tests/Functional/Symfony/EventDispatchingHttpCacheTest.php
+++ b/tests/Functional/Symfony/EventDispatchingHttpCacheTest.php
@@ -43,13 +43,14 @@ class EventDispatchingHttpCacheTest extends TestCase
 
         $httpKernel = \Mockery::mock(HttpKernelInterface::class)
             ->shouldReceive('handle')
-            ->withArgs([$request, HttpKernelInterface::MASTER_REQUEST, true])
             ->andReturn($expectedResponse)
             ->getMock();
         $store = \Mockery::mock(StoreInterface::class)
+            ->shouldReceive('lookup')->andReturn(null)->times(1)
+            ->shouldReceive('write')->times(1)
+            ->shouldReceive('unlock')->times(1)
             // need to declare the cleanup function explicitly to avoid issue between register_shutdown_function and mockery
-            ->shouldReceive('cleanup')
-            ->atMost(1)
+            ->shouldReceive('cleanup')->atMost(1)
             ->getMock();
         $kernel = new AppCache($httpKernel, $store);
         $kernel->addSubscriber(new CustomTtlListener());
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,24 @@ Changelog
 
 See also the [GitHub releases page](https://github.com/FriendsOfSymfony/FOSHttpCache/releases).
 
-2.16 (unreleased)
+2.16
 ----
 
-* Add flag `fallbackToSmaxage` to `CustomTtlListener` to allow controlling fallback to s-maxage if custom TTL header is not defined on the response.
-* Fix for Symfony HttpCache: Do not call store if Response object is not longer cacheable after event listeners. If you use the custom TTL system, this is only a performance improvement, because the TTL of the response would still be 0. With a custom listener that changes the response explicitly to not be cached but does not change s-maxage, this bug might have led to caching responses that should not have been cached
+### Symfony HttpCache
+
+* Add events `PRE_FORWARD` and `POST_FORWARD` to allow event listeners to alter
+  the request before and after it is sent to the backend.
+* Changed CustomTtlListener to use the `POST_FORWARD` event instead of
+  `PRE_STORE`. Using PRE_STORE, requests that are not considered cacheable by
+  Symfony were never cached, even when they had a custom TTL header.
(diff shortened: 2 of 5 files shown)


tests/Functional/Symfony/EventDispatchingHttpCacheTest.php
```diff
@@ -43,13 +43,14 @@ public function testEventListeners()
 
         $httpKernel = \Mockery::mock(HttpKernelInterface::class)
             ->shouldReceive('handle')
-            ->withArgs([$request, HttpKernelInterface::MASTER_REQUEST, true])
             ->andReturn($expectedResponse)
             ->getMock();
         $store = \Mockery::mock(StoreInterface::class)
+            ->shouldReceive('lookup')->andReturn(null)->times(1)
+            ->shouldReceive('write')->times(1)
```
dbu, line 50: when i only add this, the test starts failing because write is never called.

alexander-schranz:
For reference here I added a comment to: https://github.com/sulu/sulu/issues/7441#issuecomment-2245685753 which also is little bit related to new flag `CustomTtlListener(fallbackToSmaxage: false)` from https://github.com/FriendsOfSymfony/FOSHttpCache/pull/575

**Copy of the comment:** so changes here are understandable in future:

The only part which we may need to recheck is the `CustomTtlListener` makes now `private` requests always public, specially in case of `$fallbackToSmaxAge = false`.

If we want that `private` responses are never cached also when `CustomTtlListener` is used we would require:

```diff
diff --git a/src/SymfonyCache/CustomTtlListener.php b/src/SymfonyCache/CustomTtlListener.php
index 4be17d9..9ff303e 100644
--- a/src/SymfonyCache/CustomTtlListener.php
+++ b/src/SymfonyCache/CustomTtlListener.php
@@ -77,12 +77,18 @@ class CustomTtlListener implements EventSubscriberInterface
             ? $response->headers->getCacheControlDirective('s-maxage')
             : 'false'
         ;
+
         $response->headers->set(static::SMAXAGE_BACKUP, $backup);
+        $isPrivate = $response->headers->getCacheControlDirective('private');
+        // calling setTtl will make a response public, so we need to set it back to private if it was private before
         $response->setTtl(
             $response->headers->has($this->ttlHeader)
                 ? $response->headers->get($this->ttlHeader)
                 : 0
         );
+        if ($isPrivate) {
+            $response->setPrivate();
+        }
     }

     /**
```

Or alternative us `addCacheControlDirective` instead of `setTtl`:

```diff
diff --git a/src/SymfonyCache/CustomTtlListener.php b/src/SymfonyCache/CustomTtlListener.php
index 4be17d9..31126f3 100644
--- a/src/SymfonyCache/CustomTtlListener.php
+++ b/src/SymfonyCache/CustomTtlListener.php
@@ -77,8 +77,11 @@ class CustomTtlListener implements EventSubscriberInterface
             ? $response->headers->getCacheControlDirective('s-maxage')
             : 'false'
         ;
+
         $response->headers->set(static::SMAXAGE_BACKUP, $backup);
-        $response->setTtl(
+
+        $response->headers->addCacheControlDirective(
+            's-maxage',
             $response->headers->has($this->ttlHeader)
                 ? $response->headers->get($this->ttlHeader)
                 : 0
```

What do you think would be expected behaviour for:

```php
$response = new Response();
$response->setPrivate(0);
$response->header->set('X-Reverse-Proxy-TTL', 1200);
```

Current version (2.15.3) would not cache it. With https://github.com/FriendsOfSymfony/FOSHttpCache/pull/577 and `CustomTtlListener(fallbackToSmaxage: false)` it will now be cached. But I think explicit set private responses should not be cached even when a custom ttl is set? As example private response with a s-maxage defined is also not cached.

dbu:
> For reference here I added a comment to: [sulu/sulu#7441 (comment)](https://github.com/sulu/sulu/issues/7441#issuecomment-2245685753) which also is little bit related to new flag `CustomTtlListener(fallbackToSmaxage: false)` from #575

uh, very good catch! i change the code to use setCacheControlDirective. i guess until now this mistake was hidden by the listener not being called at all for a private response. i think we should keep that behaviour. 

indeed targeted cache control would be very useful to avoid this mess: https://github.com/symfony/symfony/issues/47288 (we could also provide that functionality with FOSHttpCache rules, but first should try to do the pull request to core symfony)

alexander-schranz (approved):

alexander-schranz:
@dbu retested the change and looks good for me.

PR #582, opened 2024-08-02
Mark hash request as sub request

alexander-schranz:
Reference: https://github.com/sulu/sulu/pull/7557#issuecomment-2265054566


alexander-schranz:
okay after push rethinking about this it doesn't make sense as in case of `Varnish` it will be a main request so we are not allowed to handle this differently.

dbu:
yeah, agree we should not have different behaviour between varnish and symfony cache.

PR #588, opened 2024-12-30
fix(doc): typo in the header name for Fastly

94noni:
doc https://www.fastly.com/documentation/reference/http/http-headers/Surrogate-Key/


dbu (approved):
oh damn, that was a sneaky one. thanks a lot for fixing it, i am sure this will save some time for people.

PR #592, opened 2025-06-16
Escape special characters in `Varnish::banPath()` host patterns

jdreesen:
Hostnames contain characters such as `.`, which should be escaped in a regex in order to match literally.

### DIFF
--- a/src/ProxyClient/Varnish.php
+++ b/src/ProxyClient/Varnish.php
@@ -110,7 +110,7 @@ class Varnish extends HttpProxyClient implements BanCapable, PurgeCapable, Refre
             if (!count($hosts)) {
                 throw new InvalidArgumentException('Either supply a list of hosts or null, but not an empty array.');
             }
-            $hosts = '^('.implode('|', $hosts).')$';
+            $hosts = '^('.implode('|', array_map('preg_quote', $hosts)).')$';
         }
 
         $headers = [


dbu (commented):
the documentation on the interface tells that if hosts is a string, it is expected to be a regex, but if its an array, its "exact host names". so this fix seems correct to me.

thanks for looking into it.

i would rather only change this in the 3.x branch, as 2.x is legacy at this point and 3.x the current version. this is not a security issue - it prevents unnecessary cache banning in an extreme edge cases where domain names only differ by a tiny bit. okay for you if i merge this to 3.x only?

jdreesen:
Yes, we're still working on the `2.x` branch, but we're not currently using this function. We have just discovered it during a review.

I'll rebase onto `3.x`.

dbu:
i fixed the build in the 3.x branch and added a changelog in #592

src/ProxyClient/Varnish.php
```diff
@@ -110,7 +110,7 @@ public function banPath(string $path, ?string $contentType = null, array|string|
             if (!count($hosts)) {
                 throw new InvalidArgumentException('Either supply a list of hosts or null, but not an empty array.');
             }
-            $hosts = '^('.implode('|', $hosts).')$';
+            $hosts = '^('.implode('|', array_map('preg_quote', $hosts)).')$';
```
jdreesen, line 113: Since `3.x` is PHP 8.1+, should we use a first class callable here?

```suggestion
            $hosts = '^('.implode('|', array_map(preg_quote(...), $hosts)).')$';
```

dbu: ah, good point. i added that in #594 and also adjust the test expectation to make the test not fail

PR #596, opened 2025-06-20
Remove outdated phpstan ignore

jdreesen:
Seems like this is fixed now: https://github.com/phpstan/phpstan/issues/8464#issuecomment-2850256467


jdreesen:
Ooh, the pipeline still uses PHPStan `2.1.11`, and I have no idea how to update it. 

On my local PHPStan `2.1.17` the error is gone.

dbu:
thanks!

we use a phpstan github action. that has not been tagged with the latest phpstan version - i asked for it in https://github.com/OskarStark/phpstan-ga/issues/85 and when this happens, i can re-run the test and it should get green.

dbu:
the githuab action for the phpstan github action has been fixed, so that we get latest phpstan again.

thanks for the cleanup!

PR #600, opened 2026-03-04
add support for prefix invalidation

deniak:
Cloudflare supports the [invalidation by prefix](https://developers.cloudflare.com/api/resources/cache/methods/purge/#(params)%200%20%3E%20(param)%20body%20%3E%20(schema)%20%3E%20(variant)%202%20%3E%20(property)%20prefixes).
That PR adds support for that method.

### DIFF
--- a/doc/proxy-clients.rst
+++ b/doc/proxy-clients.rst
@@ -19,17 +19,17 @@ Supported invalidation methods
 Not all clients support all :ref:`invalidation methods <invalidation methods>`.
 This table provides of methods supported by each proxy client:
 
-============= ======= ======= ======= ======= =======
-Client        Purge   Refresh Ban     Tagging Clear
-============= ======= ======= ======= ======= =======
+============= ======= ======= ======= ======= ====== =======
+Client        Purge   Refresh Ban     Tagging Prefix Clear
+============= ======= ======= ======= ======= ====== =======
 Varnish       ✓       ✓       ✓       ✓
-Fastly        ✓       ✓               ✓       ✓
+Fastly        ✓       ✓               ✓              ✓
 NGINX         ✓       ✓
-Symfony Cache ✓       ✓               ✓ (1)   ✓ (1)
-Cloudflare    ✓                       ✓ (2)   ✓
-Noop          ✓       ✓       ✓       ✓       ✓
-Multiplexer   ✓       ✓       ✓       ✓       ✓
-============= ======= ======= ======= ======= =======
+Symfony Cache ✓       ✓               ✓ (1)          ✓ (1)
+Cloudflare    ✓                       ✓ (2)   ✓ (2)  ✓
+Noop          ✓       ✓       ✓       ✓              ✓
+Multiplexer   ✓       ✓       ✓       ✓              ✓
+============= ======= ======= ======= ======= ====== =======
 
 | (1): Only when using `Toflar Psr6Store`_.
 | (2): Only available with `Cloudflare Enterprise`_.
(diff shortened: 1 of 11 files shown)


src/ProxyClient/Invalidation/PrefixCapable.php
```diff
@@ -0,0 +1,28 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\HttpCache\ProxyClient\Invalidation;
+
+use FOS\HttpCache\ProxyClient\ProxyClient;
+
+/**
+ * An HTTP cache that supports invalidation by a prefix, that is, removing
+ * or expiring objects from the cache with a given prefix or set of prefixes.
```
dbu: i would prefer to use a different wording instead of repeating "prefix", in case a reader is not sure what prefix means:

```suggestion
 * or expiring objects from the cache starting with the given string or strings.
```

src/ProxyClient/Cloudflare.php
```diff
@@ -87,6 +88,30 @@ public function invalidateTags(array $tags): static
         return $this;
     }
 
+    /**
+     * {@inheritdoc}
+     *
+     * URL prefix only available with Cloudflare enterprise plans
```
dbu:
```suggestion
     * URL prefix only available with Cloudflare enterprise plans.
     * 
     * The prefixes need to include the domain name, but not the protocol, e.g. "www.example.com/path"
```

dbu (commented):
thanks for this pull request. prefix is basicaly a special case of ban that does not allow wildcards anywhere else than at the end, right?

Multiplexer should multiplex prefix and noop implement prefix without doing anything.
we could make all clients that support ban support prefix, by converting the prefix into a banPath statement. do you think you can add that to the varnish client?

dbu:
please rebase on the 3.x branch to get the cleanup of phpstan that i did. after that, if phpstan is not green it would be because of changes in this branch.

deniak:
> thanks for this pull request. prefix is basicaly a special case of ban that does not allow wildcards anywhere else than at the end, right?

Correct!

> Multiplexer should multiplex prefix and noop implement prefix without doing anything. we could make all clients that support ban support prefix, by converting the prefix into a banPath statement. do you think you can add that to the varnish client?

I believe my last commits should address your suggestions. Thanks for the review.

doc/proxy-clients.rst
```diff
@@ -19,17 +19,17 @@ Supported invalidation methods
...
-Noop          ✓       ✓       ✓       ✓       ✓
-Multiplexer   ✓       ✓       ✓       ✓       ✓
-============= ======= ======= ======= ======= =======
+Symfony Cache ✓       ✓               ✓ (1)          ✓ (1)
+Cloudflare    ✓                       ✓ (2)   ✓ (2)  ✓
+Noop          ✓       ✓       ✓       ✓       ✓      ✓
+Multiplexer   ✓       ✓       ✓       ✓       ✓      ✓
+============= ======= ======= ======= ======= ====== =======
 
 | (1): Only when using `Toflar Psr6Store`_.
```
dbu, line 35: can we add a `*` to Prefix and add this note?

```suggestion
| (*): A limited version of Ban, that allows to invalidate by the beginning of a path
| (1): Only when using `Toflar Psr6Store`_.
```

dbu (commented):
thanks a lot! one last thing: can you please add tests for the invalidatePrefix that you added to more places, to keep the test coverage high?

doc/proxy-clients.rst
deniak: addressed in https://github.com/FriendsOfSymfony/FOSHttpCache/pull/600/commits/1490d8e2d5afda258282979e04c3a09a47f91cdf

deniak:
> thanks a lot! one last thing: can you please add tests for the invalidatePrefix that you added to more places, to keep the test coverage high?

Sorry I missed that. I added a few more tests.

dbu:
thanks a lot! i added a changelog entry and merged on the cli.

i released this as 3.2.0 now.

deniak:
Thanks @dbu

